PR #416, opened 2014-01-15
Added initial support for parsing form variables in email body

rsaccani:
I have added support for parsing dynamic form variables in the email body.
Here is how it works:
- The variables can be assigned in the email body in the form [variable name]=[value]. Example: topicId=12
- If the variable topicId is not defined no parsing is done (no topic, no dynamic form)
- Except for topicId, only variables defined by the user in the dynamic form associated to the topic can be assigned in email body

I know that this is implementation is far from being elegant but this is my first approach to osTicket and I wanted to contain the patch to class.mailfetch.php in order to be able to start using the feature as soon as poosible.

Please provide suggestions on how to improve this feature.

The main issue I faced is that FormField::getSource allows such form variables to be defined only via HTTP POST.
To work-around such limitation I pre-populate field->_clean values before saving the form.


Cm-8:
It can works also for other fields like:
Owner Name, Owner email, Owner Phone Number?

I use CF7 on WP website and if i want to create a ticket from a online form now, i can't use customer email on sender email, in sender email i've to use a fake mail in the same domain of my website.

PR #471, opened 2014-01-25
Allow staff with numeric usernames to be authenticated

### LINKED ISSUE #464: AD auth won't work for usernames with only number
Active Directory authentication is not working if staff's usename does not contain any alphabets. We have some staff who's username is just 3 digit numbers. They are not being authenticated.

dbinoj:
If a staff's username contains only numbers, it may be misunderstood as internal ID. While this patch fixes #464 and allows users with numeric usernames to login, there may be a case where internal ID of one staff is username of other staff. TODO: Staff::load should accept another variable stating if var is internal ID or username or email.

### DIFF
--- a/include/class.staff.php
+++ b/include/class.staff.php
@@ -54,12 +54,15 @@ class Staff {
             $sql .= 'email='.db_input($var);
         elseif (is_string($var))
             $sql .= 'username='.db_input($var);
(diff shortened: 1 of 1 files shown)


greezybacon:
Maybe y'all could try #2393

PR #643, opened 2014-03-03
Support REDIRECT_PATH_INFO suggested by @felixjingga

greezybacon:
This expands the idea to the rest of the system (API and AJAX)


felixjingga:
Hi,

from what I reviewed, this is good enough. I haven't seen "REDIRECT_REDIRECT_PATH_INFO" case though..

PR #748, opened 2014-04-02
Add help topic selection to email and filter management

greezybacon:
This patch allows the email channel and ticket filters to apply the help topic to new tickets. Previously, all emails created tickets without a help topic, and the filter system could neither set nor change the help topic.
### Outstanding
- [ ] What should be the preference of `priority` and `department` settings? Should the help topic settings trump the email settings?


dadjul:
Dear @greezybacon,

Is this still working on 1.10 ? We currently have email setup but not matter which topic we default the ticket is well assigned but custom forms are not added automatically.
Is this by design ?

Thank you.

PR #786, opened 2014-04-17
Change DOMDocument->loadHTML encoding fix

beholdr:
Current fix for wrong encoding in DOMDocument->loadHTML (http://us3.php.net/manual/en/domdocument.loadhtml.php#95251) did not work with cyrillic letters in e-mail letters. There is better way: http://stackoverflow.com/a/11310258 which works fine and require less code.

### DIFF
--- a/include/html2text.php
+++ b/include/html2text.php
@@ -28,19 +28,11 @@ function convert_html_to_text($html, $width=74) {
 
     $html = fix_newlines($html);
     $doc = new DOMDocument('1.0', 'utf-8');
-    if (strpos($html, '<?xml ') === false)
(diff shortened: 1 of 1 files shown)


greezybacon:
@beholdr bump. Could you spare some example HTML code that I could use to trigger the issue?

greezybacon:
@beholdr could you confirm that removing the `<?xml` processing instruction and simply replacing it with your `<meta>` tag also fixes the issue?

beholdr:
Sorry, it was a long time, I have no code at hand to confirm fix. Plus it is depends on server settings. So you could close issue.

greezybacon:
@beholdr I believe it's a real issue; however, I'm not sure what is necessary to reproduce the issue. I'd like to see it fixed

PR #866, opened 2014-05-07
Allow for duplicate tickets

protich:
osTicket uses message-id to detect previously "seen" emails in order to prevent duplicate tickets from being created from a single email. This is necessary to avoid duplicate tickets when fetched emails are left in the INBOX.

The proposal will allows for less strict matching which can allow for duplicate tickets when an email is sent to multiple email queues.

__Use Case Scenario__:
 * We have multiple departments with strict access control on tickets
  e.g members of Billing department do not have access to tickets in Legal department
 * Joe sends an email to billing@company.com and legal@company.com (CC)
 * If a ticket gets created in Legal first (unpredictable - depends on which inbox gets processed first)
  - Billing department will be left in the dark
  - No way for legal to know that the same email was also addressed to Billing and vise versa

====
- [ ] Add setting to enable/disable the strict mode.
- [x] Log internal notes on duplicate tickets (?)

### DIFF
--- a/include/class.dept.php
+++ b/include/class.dept.php
@@ -303,6 +303,10 @@ class Dept {
         return $num;
     }
 
+    function __toString() {
+        return $this->getName();
+    }
+
     /*----Static functions-------*/
 	function getIdByName($name) {
         $id=0;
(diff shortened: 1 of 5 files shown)


singularitytrav:
I just discovered we are having this issue. Any update on if/when this might be implemented? Thanks.

singularitytrav:
(I found an easy work around though so no pressure)

PR #951, opened 2014-05-26
forms: Add concept of inline form fields

greezybacon:
In-line forms are forms that are rendered as one field. The data of the inline form is also saved in the data for one field. Currently, the data is rendered to JSON and stashed in the database.

The data in the field is also accessible via the variable replacement system, so something line `%{ticket.field.subfield}` can be handled by the inline form field.
### Outstanding
- [ ] In-Code documentation of new classes
- [ ] PDF printing for inline fields
- [ ] Ensure behavior of internal fields for clients
- [ ] Add filtering capability for inline form fields
#### Form Edit

screen shot 2014-05-26 at 09 07 48
#### Field Configuration

screen shot 2014-05-26 at 09 08 00
#### Display for Agents

screen shot 2014-05-26 at 09 06 37
#### Edit for Agents

screen shot 2014-05-26 at 09 07 32
#### Display for Clients

screen shot 2014-05-26 at 09 12 04
#### Edit for Clients

screen shot 2014-05-26 at 09 13 07

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -140,6 +140,7 @@ class FormField {
             'bool' => array('Checkbox', 'BooleanField'),
             'choices' => array('Choices', 'ChoiceField'),
             'break' => array('Section Break', 'SectionBreakField'),
+            'inline' => array('Inline Form', 'InlineFormField'),
         ),
     );
     static $more_types = array();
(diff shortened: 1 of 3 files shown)


luiizfenix:
Hi @greezybacon is there any way this is compatible with the new  osTicket v1.10.1 build since i see all the lines added to the class.forms.php file but 
 

> `'inline' => array('Inline Form', 'InlineFormField'),`

 even the files
include/client/templates/inline-form.tmpl.php
and 
include/staff/templates/inline-form.tmpl.php

are present on the base installation no need of adding them. 


If i add the line 
> 'inline' => array(	/* @trans */ 'Inline Form', 'InlineFormField'),

to class.forms.php, the field type is present but i can't select any form in configuration( see screenshot)

image

, also when i select a help topic with that type of field the forms simply doesnt show(see screenshot) 

image

PR #954, opened 2014-05-27
Use priority index instead of priorityId for incoming email as the dynam...

torohill:
...ic field expects the input to named priority. This enables filtering of incoming emails by priority. osTicket/osTicket-1.8#499

This fix enables filtering of incoming emails by priority. Note that this only works if the 'Emailed Tickets Priority' option under Settings > Emails is turned on. If it's not turned on then tickets created from emails are assigned the default priority but that happens after the filters are processed, so no matches will happen.

Also, note that the original feature request (from the first comment) is already achievable with the 'Emailed Tickets Priority' setting, as it's just for setting priority, however you need filters to apply other actions (eg. assigning to a user).

I'm pretty sure I didn't break anything by renaming priorityId to priority in all these places, and it works in my particular case, but it may need testing for other situations.


torohill:
Sorry, looks like I stuffed up this pull request by deleting my priority-filter branch before it was merged, not sure why I did that. Reopening and I'll rebase/update to work with the develop branch now.

torohill:
This pull request has now been rebased on top of develop and tested to make sure it is still working. Sorry about the mess I created previous to that.

include/class.ticket.php
```diff
@@ -2572,10 +2572,6 @@ static function create($vars, &$errors, $origin, $autorespond=true,
         if (isset($vars['autorespond']))
             $autorespond = $vars['autorespond'];
 
-        # Apply filter-specific priority
-        if ($vars['priorityId'])
-            $form->setAnswer('priority', null, $vars['priorityId']);
-
```
greezybacon: Are you sure this block can be removed?

torohill: Yeah, @protich actually asked the same question on the previous pull request (the one I stuffed up). Here's a copy of my response:

> Because the input is now assigned to $vars['priority'] instead of $vars['priorityId'] it gets automatically loaded into the form (the dynamic field expects the input to be named 'priority'). Hence there is no need to explicitly use $form->setAnswer() to set the priority to $vars['priorityId'], which is what this block of code was doing.  

And here's where you can see my original response in case that is helpful:
https://github.com/torohill/osTicket-1.8/commit/1689ae90302ecfd132cd39a6ba39907ad42c48b1#comments

greezybacon: Fair enough. I agree with that logic

PR #1092, opened 2014-07-07
Fixing ticket search when dates are in the future

okasurya:
force ticket search result to null when input dates (startTime & endTime) are in the future or startTime > endTime


t-oster:
I think we only need https://github.com/osTicket/osTicket/pull/3470. After merging this, the range works for me also with future dates.

PR #1153, opened 2014-08-14
Fix attachments not working on ticket open

jchenevey:
When opening a ticket, and attaching a file, currently the code does not look for or process attachments.

My proposed code will find any attachments, process them, and add them upon opening the ticket.

### DIFF
--- a/scp/tickets.php
+++ b/scp/tickets.php
@@ -491,7 +491,8 @@ if($_POST && !$errors):
                 } else {
                     $vars = $_POST;
                     $vars['uid'] = $user? $user->getId() : 0;
-
+                    if($_FILES['attachments'])
+                	 	$vars['files'] = AttachmentFile::format($_FILES['attachments']);
                     if(($ticket=Ticket::open($vars, $errors))) {
                         $msg='Ticket created successfully';
                         $_REQUEST['a']=null;


mckaygerhard:
the #1171 breaks support for "esr" browsers in enterprise environments, such firefox v 17 that its used in my etwork with 3000 users aprox.. forces me to update firefox to v 31 in the sun terminals stations and this means licensing upgrades,... and some problems of migration...

greezybacon:
@mckaygerhard #1171 still allows for single file uploads in older browsers

mckaygerhard:
single file uploads works only in opera, in firefox suppost to work but "upload" question ("Drop files here or choose them ") anre not working.. 

i corroborates that works due i forced the "choose them" link in the form and works for single file upload in firefox 7, and chromiun 13, in firefox 17 works but sometihing hang the upload

ntozier:
Firefox 7? (release date Sept 27 2011)
Chromium 13? (release date Apr 26, 2011)
FireFox 17?  (release date Nov 20, 2012)
Current Firefox version is 35.0.
Current Chromium version is 41.0.
Maybe you should seriously consider upgrading?

mckaygerhard:
errr well there's a say: "if everithing works, dont touch anything", 

also in the enterprice environment, OBI oracle system only works with those versions.. and here, we are a lot of users.. je je a lot lot!

PR #1171, opened 2014-08-21
forms: Drag and drop uploads

greezybacon:
This replaces the multi-file upload widget for a new-fangled drag and drop widget. The new widget supports a few super-powered features
- Multi-file drag and drop
- Multi-file select (for real this time)
- Progress bar
- Trash can for removal
- Link for preview
- Attachments can be disabled for the thread entry
- Separate attachment fields can be added to any form
- Attachment settings can be configured for each upload field
- Attachments of a certain type can now be required

The widget is not connected to the Redactor editor, so that the editor can be disabled and the upload will still function properly.
### Outstanding
- [ ] Internet Explorer 9 support (FileReader javascript class is required). More help [here](https://developer.mozilla.org/en-US/docs/Web/API/FileReader)
- [x]  Integrate thumbnail and preview support for uploads
- [x] Upload cancel button
- [x] Fallback support for unsupported browsers (upload on submit)
- [x] Remove outdated settings in Admin Panel
- [x] Allow filtering of attachments by MIME type
- [ ] Add nice file type icons
##### Upload Progress

screen shot 2014-08-21 at 12 56 15 pm
##### File list view with removal option

screen shot 2014-08-21 at 12 56 41 pm
##### Thread entry settings

screen shot 2014-08-21 at 1 05 55 pm
##### New file attachment field

screen shot 2014-08-21 at 1 05 27 pm
##### Custom configuration for attachment field

screen shot 2014-08-21 at 1 05 42 pm

### DIFF
--- a/include/i18n/en_US/help/tips/settings.ticket.yaml
+++ b/include/i18n/en_US/help/tips/settings.ticket.yaml
@@ -128,40 +128,23 @@ enable_html_ticket_thread:
         If enabled, this will permit the use of rich text formatting between
         Clients and Agents.
 
-attachments:
-    title: Attachments
+ticket_attachment_settings:
+    title: Ticket Thread Attachments
     content: >
+        Configure settings for files attached to the <span
+        class="doc-desc-title">issue details</span> field. These settings
+        are used for all new tickets and new messages regardless of the
+        source channel (web portal, email, api, etc.).
 
-allow_attachments:
-    title: Allow Attachments
-    content: >
-
-emailed_api_attachments:
-    title: Emailed/API Attachments
-    content: >
-
-online_web_attachments:
-    title: Online/Web Attachments
-    content: >
-
-max_user_file_uploads:
-    title: Max. User File Uploads
-    content: >
-
-max_staff_file_uploads:
-    title: Max. Staff File Uploads
-    content: >
-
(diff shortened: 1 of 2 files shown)


rabiajavaid:
Hi,
  This not working in older version of firefox. But the file filedrop.field.js says. 
Works with Firefox 3.6+

ntozier:
Firefox current version is like 35.0.1.  Are you seriously referencing something for 3.6?  3.6 came out in Jan 2010.  I'd say that your a few years late for upgrading.

rabiajavaid:
yeah we cant upgrade it. is there is any way?

rabiajavaid:
We cannot upgrade browsers here due to other applications. It should work on Firefox at least

Chefkeks:
Sorry, but in my opinion you have no other option than upgrading
or
installing a second instance of Firefox
or
installing another (up-2-date) browser (e.g. Chrome)

mckaygerhard:
now if user lef the page and press response or publice note, file does not load due user not wait .. 

also break support for official supported browsers in oracle OBI 10 in huge network enterprices organizations..

mckaygerhard:
@rabiajavaid  have right! not working i'm using firefox 28, max supported in my provider!

PR #1340, opened 2014-09-29
Upgrader done.inc.php: scp/settings.php not working

Chefkeks:
Hey guys,

just found out that the link here is not working: scp/settings.php
The full link looks like: http://my-server.com/osticket/scp/scp/settings.php

First I thought that the leading slash is missing, but this resulted also in a not working link: /scp/settings.php
The full link misses the "osticket" directory: http://my-server.com/scp/settings.php

Finally I decided to just use "settings.php" which worked and showed a correct link:
http://my-server.com/osticket/scp/settings.php

Greetings
Michael

PS: Don't know if it should be just "settings.php" - but that is working ;)


Chefkeks:
@greezybacon
Did you missed that PR or are there any changes coming to 1.9.6 that will include that fix?

Chefkeks:
@greezybacon 
BUMP :D

greezybacon:
@Chefkeks sorry I missed this. It looks good.

PR #1349, opened 2014-10-01
filter: Filter 2.0 — greater extensibility

greezybacon:
This patch rebases filters into a row-based layout and redesigns the filter apply method to be more extensible. It also redesigns the UI to be more dynamic and to allow for actions to be added without database modification and actions can also have complex configurations.

##### Simpler view — only show configured actions
screen shot 2014-09-30 at 11 08 55 pm

##### Add new action from drop down
screen shot 2014-09-30 at 11 09 12 pm


Chefkeks:
@greezybacon 
With the new filters, will it be possible to send a canned response to the user and afterwards reject a ticket?

Asking, because currently it does not seem to be possible, even when creating 2 filters (#1 for canned response, #2 for ticket rejection) with different execution order (#1 then #2) the ticket is only rejected, but no canned response is send out to the user.

We tried to created such a filter for end users answering tickets which are closed (with reopen disabled) to say "Thanks" or "Great job!" or so and these messages we like to reject with a note / canned response that we are happy to hear that we did a good service, but the ticket was closed and we reject answers to closed ticket, instead they shall open a new ticket when they it's necessary. 

Hope that was not too much info and it will hopefully be possible to execute a filter action before a ticket gets then rejected.

greezybacon:
@Chefkeks this patch adds the ability to send an arbitrary email before the ticket is rejected. The canned response; however, will still not work, as the ticket must be accepted in order to attach a canned response.

include/class.filter_action.php
```diff
@@ -0,0 +1,513 @@
...
+
+class FA_DisableAutoResponse extends TriggerAction {
+    static $type = 'noresp';
+    static $name = /* @trans */ "Disable autoresponse";
+
+    function apply(&$ticket, array $info) {
+        # TODO: Disable alerting
+        # XXX: Does this imply turning it on as well? (via ->sendAlerts())
+        $config = $this->getConfiguration();
+        if ($config['enable']) {
```
greezybacon: Oops

Chefkeks:
@greezybacon 
Great! Thanks for the quick answer.

PR #1358, opened 2014-10-06
Tickets: Total Ticket Time and Thread Time and Type

robintoy:
Hi Guys,
one thing we find extremely useful is the mod by Edge where you can add time and a time type to the ticket allowing you to see what where and how long things were etc.

I have created a branch for this for this and now asking for a merge to the main source as I believe this is now complete and working fully, the code does the following: -
- Ability to add time to a ticket.
- Ability to select a time type which is stored in a Dynamic / Custom List and fully customisable.
- The client view shows the total time spent in minutes / hours in the ticket header.
- The staff view shows the same total in the header but also shows a breakdown in each thread title (time per update with time type of update).
- Patch SQL file also included to update the databases as required but I'm not sure this is done 100% so might need double checking.

So hope you like the mod as it so useful for us and looking at the forums so many other people too.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,7 @@ stage
 
 # Ignore packaged plugins and language packs
 *.phar
+
+# Ignore mPDF temp files
+include/mpdf/ttfontdata
+include/mpdf/tmp
--- a/css/filedrop.css
+++ b/css/filedrop.css
@@ -26,15 +26,19 @@
     margin: 0;
     border-radius: 5px;
 }
+.rtl .filedrop .files .file {
+    padding-left: 10px;
+    padding-right: 20px;
+}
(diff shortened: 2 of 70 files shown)


rp-tech:
Guys I just wanted to add I think this is an awesome potential feature and would open the door for MSP's to start using OsTicket as their main ticket tracking system (since they usually function around billable time against some sort of agreement). While I don't work for an MSP currently I think I would also find this feature to be useful to show the execs at our mid-sized company how IT actually manages their time.

Cheers and thanks for all that you do!

robintoy:
Hi @rp-tech 
have you looked at the latest version of this has this particular thread / mod has been stopped and I have re-written it much better with better features.

We are a MSP and it is a must for us as you describe.

PR #1370, opened 2014-10-07
forms: Add unique constraint to short-answer field

greezybacon:
This will allow folks using a unique identifier such as an account or customer number to have confidence that a value is not reused anywhere else for data in the same field.


protich:
Hey @greezybacon  - did you get to retest this PR?

greezybacon:
The last time I demoed it, it didn't work. I'll try and warm it back up. Should I also move it to the `develop` branch?

greezybacon:
Ok, there was a mixup in the code that, if the field did not have a validator set, then the unique check would not be performed. I moved the code ahead of the validator check, and now it seems to work better.

PR #1402, opened 2014-10-16
os ticket-1.8

cpigat:


greezybacon:
@cpigat please give some more feedback of the currently available language packs if you have time in the future.

PR #1450, opened 2014-10-30
UI: Fix datepicker vs. redactor toolbar z-index issue

Chefkeks:
Tested with 1.9.4 - found no other / related / new issues due to this change of the z-index.

Fixes: https://github.com/osTicket/osTicket-1.8/issues/980
Fixes: https://github.com/osTicket/osTicket-1.8/issues/1411

### DIFF
--- a/css/redactor.css
+++ b/css/redactor.css
@@ -327,7 +327,7 @@ body .redactor_box_fullscreen {
   background: #fff;
   border: none;
   box-shadow: 0 1px 2px rgba(0, 0, 0, 0.2);
-  z-index: 3;
+  z-index: 1;
 }
 .redactor_toolbar:after {
   content: "";


Chefkeks:
@greezybacon 
Sorry for may bothering, but any chance that this issue gets finally fixed, since I always need to make sure to apply it after an upgrade, before one of our users complains about it.

greezybacon:
@Chefkeks the patch actually looks good. I was confused as github wrote something about `.redactor_box_fullscreen`. I wasn't sure has a patch to the fullscreen feature would fix the main toolbar

Chefkeks:
:D Thank you. Let me also mentioned that we use this fix since 1.9.1 or 1.9.2 and never had any issues with it, so it should not cause any new problems ;)

greezybacon:
It's in v1.9.5.1, which came out today. Thanks, again!

Chefkeks:
@greezybacon
Great! Did you notice the SECOND XSS vulnerabilities REPORT which was added by erickroco over at the forum? http://osticket.com/forum/discussion/83738/osticket-1-9-5-xss-vulneravilities-reported-by-owasp-zap#latest

greezybacon:
@Chefkeks I just read it. It looks like the same vulnerability

PR #1535, opened 2014-12-08
Add print support for clients

greezybacon:
This adds printing support to the client portal. Clients can now print tickets.  This also changes the printing system to use an HTML template, which fixes several other issues with respect to pdf printing.


simonretter:
Hi!
printing for Clients sounds great. We have a similar demand from out Clients.
When can we expect this Feature?
/Simon

greezybacon:
@simonretter definitely this year. That's about the most narrow window I can suggest

PR #1546, opened 2014-12-15
email: Message-Id header with user and thread ID

greezybacon:
This patch suggests a change to the message-id creation process that includes stamping the receiving user-id (staff or client) along with the thread-id of the originating notice. This allows detection of threading if the clinically brain-dead mail client drops all the other header detection mechanisms, including the tag placed in the email body, on response.

This patch works for both client and agent communication.

*Hopefully this will make the next level of threading detection.*


include/class.mailer.php
```diff
@@ -90,6 +90,142 @@ function addAttachments($attachments) {
...
+     *        HMAC(sha1, tag+rand, SECRET_SALT)
+     *   -or- Original From email address
+     */
+    function getMessageId($recipient, $options=array(), $version='A') {
+        $tag = '';
+        $rand = str_replace('-','_', Misc::randCode(9));
+        $sig = $this->getEmail()?$this->getEmail()->getEmail():'@osTicketMailer';
+        if ($recipient instanceof EmailContact) {
+            // Create a tag for the outbound email
+            $tag = pack('VVa',
```
greezybacon: Should we use a character rather than a string?

include/class.mailer.php
```diff
@@ -90,6 +90,142 @@ function addAttachments($attachments) {
...
+                // Drop the leading version code
+                $rv['code'] = substr($rv['code'], 1);
+                // Verify tag signature
+                $chksig = substr(hash_hmac('sha1', $tag.$rv['id'], SECRET_SALT), -10);
+                if ($tag && $sig == $chksig && ($tag = base64_decode($tag))) {
+                    // Find user and ticket id
+                    $rv += unpack('Vuid/VthreadId/auserClass', $tag);
+                    // Attempt to make the user-id more specific
+                    $classes = array(
+                        'S' => 'staffId', 'U' => 'userId'
```
greezybacon: Perhaps we should define what collaborator classes mean to the returned email

greezybacon:
This patch triggers a [php bug](https://bugs.php.net/bug.php?id=43200) on PHP between 5.3.0 and 5.3.8. It is [fixed in 5.3.9](http://stackoverflow.com/a/22521203)

include/class.client.php
```diff
@@ -50,6 +51,9 @@ function __call($name, $args) {
 
     }
 
+    function getId() { return $this->user->getId(); }
+    function getEmail() { return $this->user->getEmail(); }
```
greezybacon: This should be one of these
```php
function getEmail() { return $this->__call('getEmail'); }
```
```php
function getEmail() { return ($this->user) ? $this->user->getEmail() : null; }
```

PR #1672, opened 2015-01-13
Fix very predictable random data on some platforms

greezybacon:
`Misc::randCode()` does not generate significantly random data for Windows platforms with a local database. This stems from the random seed using the milliseconds from the current time of day and the database connection time, in microseconds. Because Windows™ has especially poor sub-second time resolution via the `microtime()` function, the seed does not have many variations.

This patch addresses the issue by using the included `Crypto::random()` function as a source of random data rather than the `mt_rand()` function, as it uses native cryptographic random data generators if possible to generate the data, and uses `microtime()` as a fallback if no other source of random data is available on the platform.

Also `.` characters are not allowed on the left hand side of the `@` char in email message IDs, where this function is sometimes used to generate random chars.

The caveat here is that the system may generate duplicate message-id codes for thread entry items which may cause emails to be matched with incorrect ticket threads.

### DIFF
--- a/include/class.misc.php
+++ b/include/class.misc.php
@@ -16,14 +16,31 @@
 
 class Misc {
 
-	function randCode($count=8, $chars=false) {
-        $chars = $chars ? $chars
-            : 'abcdefghijklmnopqrstuvwzyzABCDEFGHIJKLMNOPQRSTUVWXYZ0123456789-.';
-        $data = '';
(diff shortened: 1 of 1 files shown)


mckaygerhard:
well, most better use right platform, how knows what king of errors have also the guindows platform appart of that!  U_U

PR #1676, opened 2015-01-14
upgrader: Support upgrading from 1.7 to v1.9.6

protich:
Make the upgrader play nice when loading latest yaml files while upgrading from versions prior to v1.8.


protich:
Pending upgrader issues to discuss.
 -  [ ]  `select_related` in `class.staff.php` on nonexistent `group` table
 -  [ ]   ORM fatal crash/exception on failed object `lookup`

PR #1677, opened 2015-01-14
Session never expires

greezybacon:
This patch sends updated session cookies to the browser when the session is refreshed on the server. This allows the session cookie to expire on the browser at the same time the session timeout occurs at the server. In the event the session timeout is configured in osTicket not to expire, the cookie will expire after seven days on the client browser, and will expire in PHP when it is garbage collected sometime after 86400 seconds after the time last refresh time.

Using this method, the session will never expire if the session timeout in osTicket is configured to 0, and the session is refreshed at least daily.

This patch also forces the client to delete the cookie at logout.

### DIFF
--- a/include/class.usersession.php
+++ b/include/class.usersession.php
@@ -133,6 +133,8 @@ class ClientSession extends EndUser {
     }
 
     function refreshSession($force=false){
+        global $cfg;
+
         $time = $this->session->getLastUpdate($this->token);
         // Deadband session token updates to once / 30-seconds
         if (!$force && time() - $time < 30)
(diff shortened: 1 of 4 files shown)


include/class.usersession.php
```diff
@@ -140,6 +142,13 @@ function refreshSession($force=false){
 
         $this->token = $this->getSessionToken();
         //TODO: separate expire time from hash??
+
+        setcookie(session_name(), session_id(),
+            ($time ?: time()) + ($cfg->getClientTimeout() ?: 604800),
```
greezybacon: Maybe we should just use the SESSION_TTL here since PHP will likely drop the session before this number of seconds, and then the client would keep sending two cookies until the first expired.

PR #1692, opened 2015-01-20
oops: ticket filters

protich:
This pull request addresses several regression issues introduced by pull request #1349 

Refactor filter.yaml and related code base for installer to load banlist filter
Move isBanned routine to banlist class and make it banlist specific
Refactor how canned responses get list of attached filters

### DIFF
--- a/include/class.banlist.php
+++ b/include/class.banlist.php
@@ -25,8 +25,31 @@ class Banlist {
         return self::getSystemBanList()->removeRule('email','equal',$email);
     }
 
-    function isbanned($email) {
-        return TicketFilter::isBanned($email);
+    /**
+     * Quick function to determine if the received email-address is in the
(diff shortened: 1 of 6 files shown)


include/class.banlist.php
```diff
@@ -25,8 +25,31 @@ function remove($email) {
...
+
+        if (!($filter=self::getFilter()))
+            return false;
+
+        $sql='SELECT filter.id '
+            .' FROM '.FILTER_TABLE.' filter'
+            .' INNER JOIN '.FILTER_RULE_TABLE.' rule'
+            .'  ON (filter.id=rule.filter_id)'
+            .' WHERE filter.id='.db_input($filter->getId())
+            .'   AND filter.isactive'
```
greezybacon: Should we add a check here to only find rules attached to the `SYSTEM BAN LIST` filter?

PR #1698, opened 2015-01-23
charset: Normalize charsets

protich:
This pull request adds a cleanup util for bogus and invalid charsets, mostly added by a nameless company out of Redmond, WA.

### DIFF
--- a/include/class.format.php
+++ b/include/class.format.php
@@ -41,43 +41,13 @@ class Format {
         return $size;
     }
 
-	/* encode text into desired encoding - taking into accout charset when available. */
-    function encode($text, $charset=null, $encoding='utf-8') {
-
-        //Try auto-detecting charset/encoding
-        if (!$charset && function_exists('mb_detect_encoding'))
-            $charset = mb_detect_encoding($text);
-
-        // Normalize bogus or ambiguous charsets
-        $charset = Charset::normalize(trim($charset));
-        $original = $text;
-        if (function_exists('iconv'))
-            $text = iconv($charset, $encoding.'//IGNORE', $text);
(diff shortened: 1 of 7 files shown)


include/class.format.php
```diff
@@ -47,12 +48,8 @@ function encode($text, $charset=null, $encoding='utf-8') {
         if (!$charset && function_exists('mb_detect_encoding'))
```
greezybacon: I kinda feel like `Format::encode` should move to `Charset::transcode`:
```php
partial class Charset {
    function transcode($text, $from, $to='UTF-8') {
        $from = self::normalize(trim($from));
        $to = self::normalize(trim($to));
        // ...
    }
}
```

protich: Good idea....I have do the deed.

Ziak:
I am having some issues getting this to work on v1.9.5.1 . When I apply the updated files to our install it fixed our issue with emails that are not UTF-8 coming in truncated but it seems to cause some other issues. One issue is embedded images in the emails are missing and when you mouse over the place holder image there is an option to download the image file but it just goes to a 404 error page. The other issue is once I apply the changes to our v1.9.5.1 install it no longer allows us to change the user on a ticket. It will let me get to the point of selecting the user but when you hit continue the screen just flashes, the popup goes away but it never brings up the confirmation popup telling you that the user will no longer have access and no change is made to the user of the ticket. Is there any way to get this encoding fix to work on 1.9.5.1? It seems to work if I sync this branch but not when applied to 1.9.5.1. I have tried it by using the raw files as wells as hand editing the 1.9.5.1 files with the code changes. Either way I get the same result. My server is running IIS6 with PHP 5.4.3.

PR #1708, opened 2015-01-30
Image annotations

greezybacon:
This patch adds the ability to add simple annotations to images in the Redactor editor. It also makes an attempt at upgrading to Redactor 10.0.5 while still supported copy+paste images.

##### Upload a graphic, select the image and click the [Annotate] button
screen shot 2015-01-30 at 1 47 59 pm

##### Add arrows, boxes, circles, and text
screen shot 2015-01-30 at 1 50 44 pm

##### Click the [Save] button and the image is replaced with the annotated version
screen shot 2015-01-30 at 2 38 16 pm


Chefkeks:
Looks like ... Feature Friday ;) :D 
Really awesome stuff! So the next version will be really nice :+1:

PR #1709, opened 2015-01-30
thread: Remove collaborators when removing the thread

greezybacon:
##### References
http://www.osticket.com/forum/discussion/84718/bug-report-orphans-in-ost-ticket-collaborator#latest

### DIFF
--- a/include/class.thread.php
+++ b/include/class.thread.php
@@ -200,8 +200,16 @@ class Thread {
             return false;
 
         $this->deleteAttachments();
+        $this->removeCollaborators();
 
         return true;
+    }
+
+    function removeCollaborators() {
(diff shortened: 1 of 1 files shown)


greezybacon:
Perhaps we should add to an upgrade task to remove all collaborators without an associated ticket thread

Chefkeks:
@greezybacon 
In my opinion you should add such a upgrade task ;) 

Beside I remember that some versions ago, when a form was deleted and had a custom list as field, some data was kept in the database preventing that the list itself could be deleted. Just mentioning since I guess that I stumbled across that issue again this week - unfortunately been really busy, but I will check again next week to be sure that's the reason or something else is it.

PR #1735, opened 2015-02-16
[FIX] Missing icon before "Add New Filter"

Chefkeks:
Fixes the missing icon before "Add new filter" in scp at the filters.php page.

Guess it was some copy and paste thing ;)


Chefkeks:
BEFORE:
image


AFTER:
image

PR #1738, opened 2015-02-16
Allow characters in random ticket number.

A-Lawrence:
Allow random numbers to include characters when specified using the @ symbol.

I.e. US-####-@@

c9c1170b63b7fee9f53ad6113cab6dc1

### DIFF
--- a/include/class.sequence.php
+++ b/include/class.sequence.php
@@ -80,22 +80,25 @@ class Sequence extends VerySimpleModel {
      *
      * Formats a number to the given format. The number will be placed into
      * the format string according to the locations of hash characters (#)
-     * in the string. If more hash characters are encountered than digits
-     * the digits are left-padded accoring to the sequence padding
+     * or @ symbols in the string. If more hash characters are encountered
+     * than digits the digits are left-padded accoring to the sequence padding
      * character. If fewer are found, the last group will receive all the
      * remaining digits.
      *
      * Hash characters can be escaped with a backslash (\#) and will emit a
      * single hash character to the output.
      *
+     * At symbols can be escaped with a backslash (\@) and will emit a single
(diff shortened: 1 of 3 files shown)


Chefkeks:
Great! :+1:

A-Lawrence:
@Chefkeks thanks! I figured that I make so many modifications to osTicket that I'd start sharing some of those I feel are more useful!

Chefkeks:
Sharing is great! Hope that @protich and @greezybacon also like it and will implement some of your changes.

include/class.sequence.php
```diff
@@ -104,15 +107,24 @@ function format($format, $number) {
         $number = str_pad($number, $total, $this->padding, STR_PAD_LEFT);
         $output = '';
         $start = $noff = 0;
+        $alphabet = "ABCDEFGHIJKLMNOPQRSTUVWXYZ";
```
greezybacon: How should we handle varying alphabets / international characters here?

Chefkeks:
@greezybacon 
I see 3 possibilities here:
- Maybe by adding a new field in the "Manage" popup for the ticket numbering options
- Using the default alphabet from A to Z (maybe not so good for Japanese, Arabic, etc.)
- Leave it as it is right now without characters

greezybacon:
@Chefkeks I too though the sequence manager popup would be a nice place

greezybacon:
I'm tagging this `db-migration` because there is no place in the `sequence` table for an alphabet currently

A-Lawrence:
The database sounds the best place for it.  Our implementation didn't require none ASCII characters, so I never went any further. If I can find the time in the coming weeks, I'll put in some time to add this to the sequence manager too.

PR #1739, opened 2015-02-16
Clear overdue on close

### LINKED ISSUE #1736: Closing overdue tickets does not remove the overdue flag
Prior to upgrading to version 1.9.4, when an overdue ticket was closed, if it was flagged as overdue, the flag was removed. I have noticed since the latest upgrade that they still appear overdue even in the closed ticket list. We have since upgraded to 1.9.5.1 and have noticed this is still the case.

This doesn't affect functionality, but wondered if I should report it as a bug?

Thanks for all the work everyone does on this product.

A-Lawrence:
Clear the overdue status of a ticket when it's closed.  Fixes #1736

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,3 +17,4 @@ stage
 # Ignore mPDF temp files
 include/mpdf/ttfontdata
 include/mpdf/tmp
+nbproject/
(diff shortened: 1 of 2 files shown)


Chefkeks:
Like this. Would be great to also have a database migration patch to reset the "is_overdue" flag to "0" for all tickets with the closed state ;)

greezybacon:
Thanks for the pull request

PR #1753, opened 2015-02-20
Extended api + Switch to disable INTL_BREAK_ITERATOR

jaakkom:
Hi,
Here is the final pullrequest for extended api. Features:
-Search tickets
-Get ticket data
-Get ThreadEntry data
-Change ticket status
-php client example on api directory.

screenshot from 2015-02-19 14 41 09


include/api.tickets.php
```diff
@@ -161,6 +161,91 @@ function processEmail($data=false) {
         return $this->createTicket($data);
     }
 
+    function search($format) {
+      $this->initalizeUser($format);
+      $params = $this->getRequest($format);
+
```
yadimon: did not read all the code, but anyway here is something wrong, what must be $search?

PR #1754, opened 2015-02-20
Always show department on agents ticket listing

jaakkom:
Hi,
UI patch, that always shows department on ticket listing.

It more easy to use with multiple departments if you know which department owns ticket.

### DIFF
--- a/include/staff/tickets.inc.php
+++ b/include/staff/tickets.inc.php
@@ -387,7 +387,7 @@ if ($results) {
             <?php
             }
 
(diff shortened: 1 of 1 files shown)


greezybacon:
@jaakkom Thanks for the idea. Are you seriously suggesting a pull request to comment out part of another feature's code? Shouldn't your pull request suggest completely removing the other feature to implement this? Or shouldn't you add an option to the admin panel and implement it here by always showing the department?

PR #1755, opened 2015-02-20
Dev switch to disable autocron.

jaakkom:
This is necessary if using copy of live data on development server.

This prevents unwanted email fetching on developement server, when having live config on database.

### DIFF
--- a/include/ost-sampleconfig.php
+++ b/include/ost-sampleconfig.php
@@ -47,6 +47,9 @@ define('DBPASS','%CONFIG-DBPASS');
 # Table prefix
 define('TABLE_PREFIX','%CONFIG-PREFIX');
 
+#Dev switch for disabling autocron
+define('DEV', CFG::DEV);
+
 #
 # SSL Options
 # ---------------------------------------------------
--- a/scp/autocron.php
+++ b/scp/autocron.php
@@ -49,7 +49,7 @@ Cron::TicketMonitor(); //Age tickets: We're going to age tickets regardless of c
 if (mt_rand(1, 9) == 4)
     Cron::CleanOrphanedFiles();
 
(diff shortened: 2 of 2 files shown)


include/ost-sampleconfig.php
```diff
@@ -47,6 +47,9 @@
 # Table prefix
 define('TABLE_PREFIX','%CONFIG-PREFIX');
 
+#Dev switch for disabling autocron
+define('DEV', CFG::DEV);
```
greezybacon: Where is `CFG::DEV` defined?

greezybacon:
There is a one-click setting in Settings / Emails to disable fetching completely for this reason

include/ost-sampleconfig.php
jaakkom: Whoops. Should be false on sample config. I`ll fix that tomorrow
25.2.2015 17.03 kirjoitti "Jared Hancock" <notifications@github.com>:

jaakkom:
Yeah, but if you get copy of live database to dev, the autocron will run
before you can disable the autocron.

2015-02-25 17:04 GMT+02:00 Jared Hancock <notifications@github.com>:

ntozier:
I guess that I don't understand why this is necessary.
Since you can simply disable it before you export the database [to import it into the new server].

jaakkom:
Ok. I close this
 25.2.2015 20.17 kirjoitti "ntozier" <notifications@github.com>:

greezybacon:
I agree with @ntozier, couldn't the export/import procedure also include a line to turn off email fetching?

```sql
update ost_config set value=0 where (`namespace`, `key`) = ('core', 'enable_mail_polling');
```

jaakkom:
You are right.
25.2.2015 20.33 kirjoitti "Jared Hancock" <notifications@github.com>:

jaakkom:
Seems to be unnecessary

PR #1759, opened 2015-02-23
Add option to unassign ticket on department transfer

worthers:
Add option to automatically unassign tick on department transfer.

### DIFF
--- a/include/class.dept.php
+++ b/include/class.dept.php
@@ -317,7 +317,7 @@ class Dept {
         // Misc. config settings
         $this->config->set('assign_members_only', $vars['assign_members_only']);
 
-        $this->config->set('unassign_tickets_dept', $vars['unassign_tickets_dept']);
+        $this->config->set('unassign_tickets_dept', isset($vars['unassign_tickets_dept']) ? 1 : 0)
 
         return true;
     }
--- a/include/class.dept.php
+++ b/include/class.dept.php
@@ -317,7 +317,7 @@ class Dept {
         // Misc. config settings
         $this->config->set('assign_members_only', $vars['assign_members_only']);
 
-        $this->config->set('unassign_tickets_dept', isset($vars['unassign_tickets_dept']) ? 1 : 0)
+        $this->config->set('unassign_tickets_dept', isset($vars['unassign_tickets_dept']) ? 1 : 0);
 
         return true;
     }
--- a/include/class.dept.php
+++ b/include/class.dept.php
@@ -294,6 +294,10 @@ class Dept {
         return $this->groups;
     }
 
+    function unassignTicketsDept() {
+        return ($this->config->get('unassign_tickets_dept', 0));
+    }
+
     function updateSettings($vars) {
 
         // Groups allowes to access department
@@ -312,6 +316,8 @@ class Dept {
 
         // Misc. config settings
         $this->config->set('assign_members_only', $vars['assign_members_only']);
+
+        $this->config->set('unassign_tickets_dept', $vars['unassign_tickets_dept']);
 
         return true;
     }
--- a/include/staff/department.inc.php
+++ b/include/staff/department.inc.php
@@ -115,6 +115,19 @@ $info=Format::htmlchars(($errors && $_POST)?$_POST:$info);
                 </span>
             </td>
         </tr>
+
+        <tr>
+            <td><?php echo __('Ticket Unassignment'); ?>:</td>
+            <td>
+                <span>
+                <input type="checkbox" name="unassign_tickets_dept" <?php echo
+                $info['unassign_tickets_dept']?'checked="checked"':''; ?>>
+                <?php echo __('Unassign ticket assignment on department transfer'); ?>
+                <i class="help-tip icon-question-sign" href="#sandboxing"></i>
+                </span>
+            </td>
+        </tr>
+
         <tr>
             <th colspan="2">
                 <em><strong><?php echo __('Outgoing Email Settings'); ?></strong>:</em>
--- a/include/staff/department.inc.php
+++ b/include/staff/department.inc.php
@@ -117,12 +117,12 @@ $info=Format::htmlchars(($errors && $_POST)?$_POST:$info);
         </tr>
 
         <tr>
-            <td><?php echo __('Ticket Unassignment'); ?>:</td>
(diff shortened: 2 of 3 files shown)


Chefkeks:
Looks good, but for me it seems that some info is stored in the database - so out of curiosity is any database hack required too?

PS: Some screenshots would be nice ;)

worthers:
No database hack required.
There's a config table (in my case ost_config) which I noticed was storing data for assign_members_only
In the include/class.dept.php (line 310ish) I noticed that the 
$this->config->set('assign_members_only', $vars['assign_members_only']); was populating the database, without altering the database schema, so I added

  $this->config->set('unassign_tickets_dept', $vars['unassign_tickets_dept']);
This successfully adds the entry into the database.

worthers:
Here is a screenshot.

dept_new_button

Chefkeks:
Thanks for the follow up! :)

jonnott:
Ready to merge... ?

scp/tickets.php
```diff
@@ -129,6 +129,12 @@
                 //If no errors - them attempt the transfer.
                 if(!$errors && $ticket->transfer($_POST['deptId'], $_POST['transfer_comments'])) {
                     $msg = sprintf(__('Ticket transferred successfully to %s'),$ticket->getDeptName());
+
+                    //unassign ticket
+                    if($dept->unassignTicketsDept()=="on") {
```
greezybacon: This should be a boolean function:
```php
if ($dept->unassignOnTransfer()) {
    $ticket->unassign();
}
```
Can we get rid of "on" and just use 1 or 0?

include/staff/department.inc.php
```diff
@@ -115,6 +115,19 @@
...
             </td>
         </tr>
+
+        <tr>
+            <td><?php echo __('Ticket Unassignment'); ?>:</td>
+            <td>
+                <span>
+                <input type="checkbox" name="unassign_tickets_dept" <?php echo
+                $info['unassign_tickets_dept']?'checked="checked"':''; ?>>
+                <?php echo __('Unassign ticket assignment on department transfer'); ?>
```
greezybacon: I think we should make this clear that this setting is applied when a ticket is transfer *into* this department. (Rather than when a ticket is transferred *away* from the department).

include/class.dept.php
```diff
@@ -313,6 +317,8 @@ function updateSettings($vars) {
         // Misc. config settings
         $this->config->set('assign_members_only', $vars['assign_members_only']);
 
+        $this->config->set('unassign_tickets_dept', $vars['unassign_tickets_dept']);
```
greezybacon:
```php
$this->config->set('unassign_tickets_dept', isset($vars['unassign_tickets_dept']) ? 1 : 0);
```
should get rid of storing "on"

include/staff/department.inc.php
Chefkeks: Maybe the term should be something like "Unassign ticket when transferring to another department" or just "Unassign ticket on,department transfer".

scp/tickets.php
Chefkeks: /Totally agree with boolean !

Chefkeks:
Needs some tweaking, but looks really good so far - at least for my system admin eyes :D

worthers:
I've done a bit of tweaking based upon suggestions.
1) Changed to boolean from 'on' based on greezybacon's suggestion
2) Changed text based on Chefkek's suggestion.
I've tested this, and it works, but if data in the config table is currently using on, this will have to be manually changed to 1

greezybacon:
I don't mean to be a thorn in anyone's side. I do a bit of customer support for osTicket, and I can see this being a point of confusion...

So, as the code is written (the way I read it anyway), the assignee is cleared when a ticket is transferred into a department having this option set. But the documentation seems to say the opposite. Let me demonstrate:

Department A -> Standard department
Department B -> Unassign option set

If a ticket is created and routed to department A and assigned to Jared, what happens when the ticket is transferred to department B. Furthermore, if a ticket is currently routed to department B and assigned to Bob, what happens when it is transferred to department A.

In my opinion, the logic as written is correct, that if the unassign option is set for department B, then that would mean that someone in department B needs to work on the ticket. So when a ticket is transferred *into* department B, then the assignee should be cleared.

However, the documentation isn't written this way. It seems to say that "when transferring to another department", i.e., when transferring *away* from department B, then the assignee would be cleared. I think this is a less useful case. And it would also be far more difficult to discover why no one is assigned to the ticket after it was transferred away from department B. (*I would start by reviewing the settings for the ticket's current associations — not its previous ones).

Chefkeks:
@greezybacon 
Maybe I am not so deep into the problem right now or my English is too bad at that point, but for me everything seems ok. So how would you make it clear what's meant with that option and how to avoid a possible confusion? Just asking since I feel like I missed something here or it's really my English that's too bad.

greezybacon:
> 'Unassign ticket when transferring to another department'

Perhaps I would suggest, "Clear current assignee when tickets are transferred into this department". The phrase *To another department*, when seen inside the configuration page for a department says to me that the setting does not affect **this** department—it would affect the ticket when *transferring to another department*. Am I misreading it?

Chefkeks:
Maybe not, maybe I am misreading it...

From my point there are basically 2 possibilities (following below) and I am currently a bit unsure how this PR is going to work. 

Possibility A) 
Source department triggers this setting -> "Unassign ticket when transferring to another department"

Possibility B)
Target department triggers this setting -> "Clear current assignee when tickets are transferred into this department"

So basically what I do not currently clearly understand is, whether the source or the target department triggers that setting? Even after looking at the code again I am still unsure, but based on that it should be pretty easy to define if A) or B) is then right/correct.

greezybacon:
Could we perhaps use the "Restrict ticket assignment ..." option for this as well? For instance, when transferring a ticket into a department with the "Restrict ..." option set, and the current assignee is not a member of the department, then clear the assignee. I think that would be a simpler approach and would not require the extra checkbox nor the confusing documentation.

Chefkeks:
Really good idea :+1:

I think it would be great to implement it that way and it would perfectly makes sense (at least for me) to combine these two options and avoid confusions too.

protich:
Suggestion put forth by Jared implemented in pull request #1923.

protich:
Merged and released as part of v1.9.7

PR #1760, opened 2015-02-24
Upgrader bug fixes

protich:


include/class.group.php
```diff
@@ -250,6 +250,7 @@ static function getIdByName($name){
 
     static function create($vars=false) {
         $group = parent::create($vars);
+        $group->flags = 0;
```
greezybacon: Should we instead do
```php
if (!isset($vars['flags']))
    $vars['flags'] = 0;
```
before we call the base `create` method?

PR #1761, opened 2015-02-24
Add %{message} as an available variable for assignment emails.

torohill:
This change allows the %{message} variable to be used in the Ticket Assignment Alert email template. It just uses the first message from the ticket thread.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -1242,10 +1242,15 @@ class Ticket {
         if ($recipients
                 && ($msg=$tpl->getAssignedAlertMsgTemplate())) {
 
+            //Add %{message} as an available template variable.
+            //Shows the first message from the thread.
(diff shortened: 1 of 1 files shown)


protich:
Thank you for the PR, but I think it is unnecessary and potentially confusing (historically we've used ``%{message}`` in context for **new** message). 

osTicket already supports getting original and the latest message by using  ``%{ticket.thread.original}`` and ``%{ticket.thread.lastmessage}`` respectively.

PS: We don't have good documentation on supported variable. Hopefully we'll address it soon.

torohill:
Yes, I agree it would be confusing. I didn't realise that `%{ticket.thread.original}` and `%{ticket.thread.lastmessage}` were available, thanks for letting me know.

greezybacon:
I've opened #1781 to address the documentation issue

PR #1765, opened 2015-02-25
forms: Add Textbox Selection Widget

protich:
Add a text box widget to selection field. It functions like typeahead and drop down widgets with the exception that the user doesn't get the hint. This is useful when an internal list needs to be used to validate user's input e.g serial numbers.

screen shot 2015-02-25 at 1 56 23 am

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -2274,6 +2274,30 @@ class TextboxWidget extends Widget {
     }
 }
 
+
+class TextboxSelectionWidget extends TextboxWidget {
+    //TODO: Support multi-input e.g comma separated inputs
+    function render($options=array()) {
+
+        if ($this->value && is_array($this->value))
+            $this->value = current($this->value);
+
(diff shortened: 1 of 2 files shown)


protich:
@greezybacon - I think this baby is ready to go.

include/class.forms.php
```diff
@@ -2274,6 +2274,30 @@ function render($options=array()) {
...
+        if ($this->value && is_array($this->value))
+            $this->value = current($this->value);
+
+        parent::render($options);
+    }
+
+    function getValue() {
+
+        $value = parent::getValue();
+        if (($i=$this->field->getList()->getItem((string) $value)))
```
greezybacon: Just a thought here. Should we also consider content in the `extra` column so abbreviations would be welcome in the textbox too?

protich: Good idea :+1:

PR #1771, opened 2015-02-26
forms: Show ••• placeholder for saved pw entries

greezybacon:
Also, fix roundtrip encryption if the form name changes (which is based on a user's session), and avoid saving a NULL value if the password is not changed when a form is updated.

* [ ]  Needs to verify that it still works when the field is required.


protich:
Rebase for develop-next

PR #1776, opened 2015-02-27
filedrop: Support Firefox down to v6

### LINKED ISSUE #1746: Odd failure attaching file in old browser
This is rather minor, but since there is an error code for unsupported browsers:

Attaching a file to a ticket thread entry in a browser that doesn't have a "Blob" constructor function fails in a very unintuitive way. The upload process crashes with a javascript error but the file is still displayed on screen below the comment box.

When the comment actually posts, ThreadEntry::create() ends up with $vars['cannedattachments'] = array(0 => 0). Because this is an array, it calls ThreadEntry::saveAttachments. Because zero is numeric, Thread::entrySaveAttachment accepts this as a file ID and creates a record in the ticket attachment table (with file_id=0). This makes little attachment icons show up in various places even though the file was never successfully uploaded.

In brief testing, doing this in filedrop.field.js presents an appropriate error dialog. Not being familiar with the code base I'm not sure if this is the best solution though.
```javascript
    // Respond to an upload
    function upload() {
      stop_loop = false;

      if (!files) {
        opts.error(errors[0]);
        return false;
      } else if (typeof Blob != "function") {
        opts.error(errors[0]);
        return false;
      }
```

greezybacon:
Also, give a "Browser not supported" error for browsers which do not support the Blob constructor.

Maybe fixes #1746

References:
http://caniuse.com/#feat=blobbuilder

### DIFF
--- a/js/filedrop.field.js
+++ b/js/filedrop.field.js
@@ -330,7 +330,8 @@
       globalProgressUpdated: empty,
       speedUpdated: empty
       },
-      errors = ["BrowserNotSupported", "TooManyFiles", "FileTooLarge", "FileTypeNotAllowed", "NotFound", "NotReadable", "AbortError", "ReadError", "FileExtensionNotAllowed"];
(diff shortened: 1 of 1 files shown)


protich:
Reviewed-By: Peter Rotich <peter@osticket.com>

PR #1778, opened 2015-02-27
access: Allow empty staff banner. Show update errors

### LINKED ISSUE #1775: Agent Login Banner can't be saved if summary field are empty
by default Agent Login Banner have empty summary field. If we try change subject field and save we will see Loading... banner without any result.

If we write any text to summary field Agent Login Banner saved successfully

OT 1.9.5.1

greezybacon:
Allow the staff banner to be set to empty (which is the default).

Also display the update errors back on the dialog for failed updated.

Maybe fixes #1775

### DIFF
--- a/include/ajax.content.php
+++ b/include/ajax.content.php
@@ -136,6 +136,7 @@ class ContentAjaxAPI extends AjaxController {
             Http::response(403, 'Login Required');
 
         $content = Page::lookup($id, $lang);
+        $info = $content->getHashtable();
         include STAFFINC_DIR . 'templates/content-manage.tmpl.php';
     }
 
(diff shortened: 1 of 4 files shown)


include/ajax.content.php
```diff
@@ -154,19 +156,25 @@ function updateContent($id) {
...
-                Http::response(500, 'Unable to update content: '.print_r($errors, true));
+        // Allow empty content for the staff banner
+        if ($content->save($id, $vars, $errors,
+            $content->getType() == 'banner-staff')
+        ) {
+            Http::response(201, 'Have a great day!');
         }
+        if (!$errors['err'])
+            $errors['err'] = __('Correct the error(s) below and try again!');
+        $info = $_POST;
```
protich: Should we htmlchar ``$_POST``?

greezybacon: I thought the same; however, it looks like usages in the template are all `htmlchars()ed`

PR #1782, opened 2015-02-27
fixes #1777

### LINKED ISSUE #1777: following the /scp link causes a redirect to plain HTTP when using HTTPS
When visiting an osTicket login page `https://myosticket.example.com/login.php` secured with TLS and clicking on

> I'm an agent — sign in here

the user is redirected to a plain HTTP version of the site.

Apparently, the logic that adds a trailing slash to the url is causing the redirect.

output from `curl -i https://myosticket.example.net/scp`
```
HTTP/1.1 301 Moved Permanently
Server: nginx
Date: Fri, 27 Feb 2015 14:09:50 GMT
Content-Type: text/html; charset=iso-8859-1
Content-Length: 238
Connection: keep-alive
Location: http://myosticket.example.net/scp/

<!DOCTYPE HTML PUBLIC "-//IETF//DTD HTML 2.0//EN">
<html><head>
<title>301 Moved Permanently</title>
</head><body>
<h1>Moved Permanently</h1>
<p>The document has moved <a href="http://myosticket.example.net/scp/">here</a>.</p>
</body></html>
```

I had a look through the `.htaccess` files, but my apache foo is pretty limited, and it doesn't look like there is one for the base directory.

If this is a misconfiguration of the server, I'm happy to liase with my shared hosts, but it seems that the exposed URL should have the slash appended anyway.

I'm currently on 87f3028ba3ea1a779d1989a57dab8dee1615817a

ldrumm:
See the discussion at https://github.com/osTicket/osTicket-1.8/issues/1777 for details

### DIFF
--- a/include/client/login.inc.php
+++ b/include/client/login.inc.php
@@ -56,7 +56,7 @@ if ($cfg && $cfg->isClientRegistrationEnabled()) {
 <?php } ?>
     <div>
     <b><?php echo __("I'm an agent"); ?></b> —
-    <a href="<?php echo ROOT_PATH; ?>scp"><?php echo __('sign in here'); ?></a>
+    <a href="<?php echo ROOT_PATH; ?>scp/"><?php echo __('sign in here'); ?></a>
     </div>
     </div>
 </div>


protich:
Thanks for the pull request :+1:

ldrumm:
:)

PR #1793, opened 2015-03-02
forms: Add ability to delete properties form

protich:
This adds ability to delete form properties on list deletion.

### DIFF
--- a/include/class.dynamic_forms.php
+++ b/include/class.dynamic_forms.php
@@ -189,9 +189,11 @@ class DynamicForm extends VerySimpleModel {
     }
 
     function delete() {
+
         if (!$this->isDeletable())
             return false;
 
+        // Soft Delete: Mark the form as deleted.
         $this->setFlag(self::FLAG_DELETED);
         return $this->save();
     }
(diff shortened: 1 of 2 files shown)


protich:
TODO: We cannot assume a dynamic entry/answer has  an associated form.

Chefkeks:
I am not so deep into the code, so stupidly asking the following:
Does it fix https://github.com/osTicket/osTicket-1.8/issues/1018 ?

If so, you should may consider also a database cleanup / upgrade patch to delete orphaned data which is still in the database. If not, sorry as said not so deep into the code, sometimes I can read what a PR will do sometimes I am struggling with reading the code ;)

greezybacon:
@Chefkeks the dynamic forms code is the most abstract code in the system. This code specifically disconnects fields from their form when the form is deleted. Except for custom list deletions, which will cause the fields of the properties to actually be deleted. Unfortunately, the inability to drop a custom list is still outstanding.

Chefkeks:
@greezybacon 
Thanks for the explanation :)

greezybacon:
@protich rebase required

protich:
Rebased on top of develop-next

PR #1794, opened 2015-03-02
oops: Fix corruption of thread body posted by collabs

greezybacon:
If the system receives an email by a collaborator which has not yet been added to the ticket (a friend of a friend — that is, a collaborator forwards an email to a third-party), a header is added to the thread body something like:
```
Received From: afriendofafriend@anothercompany.tld
```
However, if the thread body is text and the HTML ticket thread is enabled, then the text formatting hint will be lost and the body will be assumed as HTML deeper inside the thread entry creation process. Therefore, the whitespace inside the resulting thread entry will be collapsed.

This patch addresses the issue by maintaining the original format hint with the thread body.

### DIFF
--- a/include/class.thread.php
+++ b/include/class.thread.php
@@ -742,9 +742,10 @@ Class ThreadEntry {
         else {
             //XXX: Are we potentially leaking the email address to
             // collaborators?
-            $header = sprintf("Received From: %s\n\n", $mailinfo['email']);
+            $header = sprintf("Received From: %s <%s>\n\n", $mailinfo['name'],
+                $mailinfo['email']);
             if ($body instanceof HtmlThreadBody)
-                $header = nl2br($header);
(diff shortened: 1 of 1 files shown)


include/class.thread.php
```diff
@@ -742,8 +742,14 @@ function postEmail($mailinfo) {
         else {
             //XXX: Are we potentially leaking the email address to
             // collaborators?
-            $vars['message'] = sprintf("Received From: %s\n\n%s",
-                $mailinfo['email'], $body);
+            $header = sprintf("Received From: %s\n\n", $mailinfo['email']);
+            if ($body instanceof HtmlThreadBody)
+                $header = nl2br($header);
```
greezybacon:
```php
$header = nl2br(Format::htmlchars($header));
```

PR #1808, opened 2015-03-05
cli: user: Add list, activate, lock features

greezybacon:
This adds the ability to find, activate, lock, and optionally send a password-reset email to one ore more users in the help desk.

### DIFF
--- a/include/class.user.php
+++ b/include/class.user.php
@@ -457,14 +457,22 @@ class User extends UserModel {
             $users[] = $data;
         }
 
+        db_autocommit(false);
+        $error = false;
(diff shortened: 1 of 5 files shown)


include/class.user.php
```diff
@@ -932,6 +940,10 @@ function sendConfirmEmail() {
         return static::sendUnlockEmail('registration-client') === true;
     }
 
+    function setPassword($new) {
+        $this->set('passwd', Passwd::hash($vars['passwd1']));
```
greezybacon: Oops. `$new` should be used for the updated password

PR #1820, opened 2015-03-11
wording: must be agents, because its agents table

yadimon:
Its a Agents table. For good translation the column of agents must have the name "Agents"


greezybacon:
Thanks for the pull request

PR #1841, opened 2015-03-16
Fix bad plain text initial ticket's internal note

joelpelaez:
When we created a new ticket with a internal note this is processed
as HTML by default without check if HTML Thread is enabled or not.

The issue is than not convert newlines (\n) to line break (<br>)
and show the internal post with its properly line breaks.


greezybacon:
@joelpelaez thanks for reporting the bug. Could you see if the patch in #2001 addresses the issue?

greezybacon:
@joelpelaez we merged a fix for this in #2001 towards v1.9.8. Thanks for the report

PR #1845, opened 2015-03-16
files: Clear attachment box after posting a response

### LINKED ISSUE #1833: Client Ticketview, Fileupload bug
v1.9.6
After the client made a fileupload, the files are still in the editor as attachment. So if the client sends then a second message, it will upload the files a second time. but it musstn't

greezybacon:
Maybe fixes #1833


infectormp:
I confirm, this patch work.

PR #1846, opened 2015-03-16
thread: Log remote IP for staff replies and notes

### LINKED ISSUE #724: Log IP address for each thread reply.
I would like to log ipaddress from which the reply was made for each thread reply. I see that the ip_address field is empty for replies in ost_ticket_thread. How do I populate these fields?

greezybacon:
Fixes #724


Chefkeks:
Hope to find the time to try that out later.

PR #1851, opened 2015-03-17
kb: Add setting to enable only for authenticated clients

greezybacon:
screen shot 2015-03-17 at 11 39 01 am

### DIFF
--- a/include/class.config.php
+++ b/include/class.config.php
@@ -210,6 +210,13 @@ class OsticketConfig extends Config {
     }
 
     function isKnowledgebaseEnabled() {
+        global $thisclient;
+
+        if ($this->get('restrict_kb', false)
+            && (!$thisclient || $thisclient->isGuest())
+        ) {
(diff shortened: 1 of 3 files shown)


Chefkeks:
I guess this will make a lot of osTicket admins happy ;)
:+1:

robintoy:
What would be great is if you could enable that but then turn this on for certain topics and not for others.

For instance we have a spam filering system called "SpamSnake" which would be nice to keep KBs in-house to registered users / clients and not public where base Outlook issues we do not mind anyone accessing.

greezybacon:
@robintoy I'm trying to avoid a more sophisticated ACL tool for the moment. I think long term we need the ACL option though

robintoy:
Hi @greezybacon , its good to know it is on the cards but yes you have to start somewhere. Been looking at some of the other pulls and looks like an ACL is needed for a lot of them.

PR #1856, opened 2015-03-17
dashboard: Fix initial page load and stats pagination

### LINKED ISSUE #1828: Dashboard graph comes up blank, needs one or more refreshes to display properly.
This occurs pretty much every time I go to the dashboard - I'll have to refresh, leave the page, come back, refresh again, or some dance like that to get the dashboard to display properly.  At some point it will draw parts of the graph like the node labels, but have to do it several times to get the whole thing to show up.

osticketblankdashboard

This has been pretty consistent since I installed the software, through multiple updates.

Currently running on Debian wheezy, nginx 1.62, MySQL 5.5.41, php 5.4.37-1~dotdeb.1.  All patches up to date and currently on osTicket v1.9.6 (e3e77001)

### LINKED ISSUE #1853: Uncaught TypeError: Cannot call method 'attr' of undefined in reports topics
when i ty to see another page of reports by topics, i get on console:

```
Uncaught TypeError: Cannot call method 'attr' of undefined 
```

greezybacon:
This patch addresses the issue where PJAX permanently breaks the loading order of scripts. Therefore the Raphael library currently used for the graph visualization is not loaded properly with PJAX. Therefore, PJAX is simply disabled on the dashboard page.

This also properly implements pagination for the statistics detail tabs, which seems to have been broken along the development trail somewhere.

Fixes #1853
Fixes #1828

### DIFF
--- a/include/staff/templates/navigation.tmpl.php
+++ b/include/staff/templates/navigation.tmpl.php
@@ -1,7 +1,10 @@
 <?php
 if(($tabs=$nav->getTabs()) && is_array($tabs)){
     foreach($tabs as $name =>$tab) {
-        echo sprintf('<li class="%s"><a href="%s">%s</a>',$tab['active']?'active':'inactive',$tab['href'],$tab['desc']);
+        echo sprintf('<li class="%s %s"><a href="%s">%s</a>',
+            $tab['active'] ? 'active':'inactive',
+            @$tab['class'] ?: '',
+            $tab['href'],$tab['desc']);
         if(!$tab['active'] && ($subnav=$nav->getSubMenu($name))){
             echo "<ul>\n";
             foreach($subnav as $k => $item) {
--- a/scp/js/dashboard.inc.js
+++ b/scp/js/dashboard.inc.js
@@ -136,6 +136,8 @@
         }
 
         var group = current_tab.attr('table-group');
+        var pagesize = 25;
+        getConfig().then(function(c) { if (c.page_size) pagesize = c.page_size; });
         $.ajax({
             method:     'GET',
             dataType:   'json',
@@ -144,7 +146,6 @@
             success:    function(json) {
                 var q = $('<table>').attr({'class':'table table-condensed table-striped'}),
                     h = $('<tr>').appendTo($('<thead>').appendTo(q)),
-                    pagesize = 25,
                     max = [];
                 for (var c in json.columns) {
                     h.append($('<th>').append(json.columns[c]));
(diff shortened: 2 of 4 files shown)


include/staff/templates/navigation.tmpl.php
```diff
@@ -1,7 +1,10 @@
 <?php
 if(($tabs=$nav->getTabs()) && is_array($tabs)){
     foreach($tabs as $name =>$tab) {
-        echo sprintf('<li class="%s"><a href="%s">%s</a>',$tab['active']?'active':'inactive',$tab['href'],$tab['desc']);
+        echo sprintf('<li class="%s %s"><a href="%s">%s</a>',
+            $tab['active'] ? 'active':'inactive',
+            @$tab['class'] ?: '',
```
mckaygerhard: hey @greezybacon the "@" are valid? i currently commiting this to my testing env to test, but found that "@" in tab class!

scp/js/dashboard.inc.js
```diff
@@ -194,30 +198,31 @@
...
                     page = Math.max(1, parseInt(current) - 1);
-                    goabs($('tbody[page='+page+']', q));
+                    return goabs($('tbody[page='+page+']', q));
                 }
                 function gonext() {
                     current = $('tbody:not(.hide)', q).attr('page');
                     page = Math.min(Math.floor(json.data.length / pagesize) + 1,
                         parseInt(current) + 1);
-                    goabs($('tbody[page='+page+']', q));
+                    return goabs($('tbody[page='+page+']', q));
```
mckaygerhard: the gonext will not should return any ? @greezybacon due this recent commit?

mckaygerhard:
fixeds the refresh error, but stil error in pagination

include/staff/templates/navigation.tmpl.php
greezybacon: `@` means silence errors in PHP

scp/js/dashboard.inc.js
greezybacon: `gonext` is used in a javascript event routine, which returning `false` indicates that the default event handler should not be invoked.

greezybacon:
@mckaygerhard I can't confirm the remaining error in pagination

mckaygerhard:
ok, lest enumerate why:

1) currently i see a commit that firefox v6 to 12 are supported..
2) so i tested currently in firefox 7 (icecat gnu), iceweasel 17 and firefox 11, error still persist
3) so then i tested in newer browsers, chromium 28 and chrome 32, error still persist.
4) so then i tested in more newer laste up to " paris a la moda" and works with problems (desing desorganization)

paste pic with error on chrome 40 (chrome 40 and lasted have error in cahce for guindows os's) so using newer browsers are not a choice for my company:

error-reporte

ntozier:
my guess would be language pack.
Does it do the same thing in English?

greezybacon:
It looks like a bad merge to me. It doesn't look anything like that on my view

mckaygerhard:
Works great on "chrome" chromium "," opera "," konqueror "but fails with" firefox ". no matter what firefox version i use, only fails if i use firefox

and.. Certainly, in Midory or chromium, resource intensive ..

PR #1865, opened 2015-03-18
Agent internal activity alert

greezybacon:
This adds the ability of the system to notify agents when another agent replies to a ticket. This is implemented by extending the "New Internal Note Alert" to also include replies by agents. The same alert settings will be applied to both alerts — new internal notes and new replies.

### DIFF
--- a/include/class.config.php
+++ b/include/class.config.php
@@ -673,19 +673,20 @@ class OsticketConfig extends Config {
         return ($this->get('message_alert_acct_manager'));
     }
 
-    function alertONNewNote() {
+    //TODO: change note_alert to activity_alert
+    function alertONNewActivity() {
         return ($this->get('note_alert_active'));
     }
 
-    function alertLastRespondentONNewNote() {
+    function alertLastRespondentONNewActivity() {
         return ($this->get('note_alert_laststaff'));
     }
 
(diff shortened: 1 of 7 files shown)


include/class.config.php
```diff
@@ -673,19 +673,20 @@ function alertAcctManagerONNewMessage() {
         return ($this->get('message_alert_acct_manager'));
     }
 
-    function alertONNewNote() {
+    //TODO: change note_alert to activity_alert
```
protich: I think we should change `note_alert_*` to `activity_alert_*` to ward-off confusions going forward.

greezybacon: I added the `db-migration` label to signify renaming the config settings on some upgrade task

Chefkeks:
Another great enhancement coming down the road ;) :+1:

robintoy:
Looking forward to this as it keep all people in the loop without having to open each ticket.

protich:
I think we should consider adding a db patch to change `note_alert_*` to `activity_alert_*`

greezybacon:
@protich I was thinking the patch should go on one of the existing patches toward `develop-next` so that the db patches don't need to be rewritten

PR #1868, opened 2015-03-18
auth: Add concept of bk passwd update and policy

greezybacon:
This adds the concept of a PasswordPolicy registration system, which provides an extensible way of administering and configuring password complexity policies.

It also adds a setPassword() method for the authentication backends which will allow for the respective backend up update the password according to whatever method is suitable for the respective backend (such as remote updates for LDAP).

I'll open a concept plugin to take advantage of this feature (osTicket/core-plugins#60)

### DIFF
--- a/include/class.auth.php
+++ b/include/class.auth.php
@@ -347,6 +347,38 @@ abstract class AuthenticationBackend {
         return false;
     }
 
+    /**
+     * Request the backend to update the password for a user. This method is
+     * the main entry for password updates so that password policies can be
+     * applied to the new password before passing the new password to the
+     * backend for updating.
+     *
+     * Throws:
+     * BadPassword — if password does not meet policy requirement
+     * PasswordUpdateFailed — if backend failed to update the password
+     */
+    function setPassword($user, $password, $current=false) {
+        PasswordPolicy::checkPassword($password, $current);
+        $rv = $this->syncPassword($user, $password);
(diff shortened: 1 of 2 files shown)


include/class.auth.php
```diff
@@ -347,6 +347,38 @@ function supportsPasswordChange() {
...
+    /**
+     * Request the backend to update the password for a user. This method is
+     * the main entry for password updates so that password policies can be
+     * applied to the new password before passing the new password to the
+     * backend for updating.
+     *
+     * Throws:
+     * BadPassword — if password does not meet policy requirement
+     * PasswordUpdateFailed — if backend failed to update the password
+     */
```
greezybacon: No docs for the return value

include/class.staff.php
```diff
@@ -539,15 +582,6 @@ function updateProfile($vars, &$errors) {
...
 
-        if ($vars['passwd1']) {
-            $this->change_passwd = 0;
-            $this->passwdreset = SqlFunction::NOW();
-            $this->passwd = Passwd::hash($vars['passwd1']);
-            $info = array('password' => $vars['passwd1']);
-            Signal::send('auth.pwchange', $this, $info);
-            $this->cancelResetTokens();
-        }
-
```
greezybacon: What about the password update section in `::update()`

PR #1882, opened 2015-03-20
Add concept of thread editing

greezybacon:
Threads can be edited by marking the original as hidden and setting it's PID to the id of the new entry. The new entry has cloned data from the original and sets the `updated` timestamp to reflect the time of last edit. An edited flag is added to the new entry to reflect its origin.

This patch suggests that agents can edit their own posts, department managers can edit posts while the ticket is in their department, and that help desk admins can edit anything. Perhaps this should be yet another role privilege.

If a post is edited more than once, only the most recent edit is kept. This is not strictly required by the data model, but avoids the possibility of significant recursion to find a thread's complete history.

### Outstanding
  * [x] Allow resend only if not allowed to edit
  * [x] Add selection original sender's signature
  * [x] Thread entry counts are not correct
  * [x] New thread edit role should float to bottom of ticket roles
  * [ ] Add convert message to internal note
  * [x] Edit **agents** into *Ability to edit thread items of other users*

##### Oops, I forgot to type my reply message
screen shot 2015-03-19 at 11 40 27 pm
##### No big deal
screen shot 2015-03-19 at 11 40 36 pm
##### Add one in the new popup dialog
screen shot 2015-03-19 at 11 41 49 pm

##### Type and save away
screen shot 2015-03-19 at 11 42 12 pm

##### The original is still available
screen shot 2015-03-20 at 6 56 10 pm
screen shot 2015-03-20 at 6 52 53 pm

##### Works with plain text entries as well
screen shot 2015-03-20 at 12 08 09 am

##### Power users can be granted access to edit any post
screen shot 2015-03-20 at 10 22 24 am

##### Responses can be resent
screen shot 2015-03-20 at 6 49 43 pm

### DIFF
--- a/include/class.thread.php
+++ b/include/class.thread.php
@@ -82,6 +82,7 @@ class Thread extends VerySimpleModel {
             'has_attachments' => SqlAggregate::COUNT('attachments', false,
                 new Q(array('attachments__inline'=>0)))
         ));
+        $base->exclude(array('flags__hasbit'=>ThreadEntry::FLAG_HIDDEN));
         if ($criteria)
             $base->filter($criteria);
         return $base;
@@ -398,7 +399,8 @@ class ThreadEntry extends VerySimpleModel {
     static $meta = array(
         'table' => THREAD_ENTRY_TABLE,
         'pk' => array('id'),
-        'select_related' => array('staff', 'user'),
+        'select_related' => array('staff', 'user', 'email_info'),
+        'ordering' => array('created'),
         'joins' => array(
             'thread' => array(
                 'constraint' => array('thread_id' => 'Thread.id'),
@@ -430,6 +432,8 @@ class ThreadEntry extends VerySimpleModel {
     );
 
     const FLAG_ORIGINAL_MESSAGE         = 0x0001;
+    const FLAG_EDITED                   = 0x0002;
+    const FLAG_HIDDEN                   = 0x0004;
 
     var $_headers;
     var $_thread;
@@ -1734,7 +1738,7 @@ abstract class ThreadEntryAction {
     static $id;                 // Unique identifier used for plumbing
     static $icon = 'cog';
 
-    var $thread;
+    var $entry;
 
     function getName() {
         $class = get_class($this);
@@ -1751,13 +1755,13 @@ abstract class ThreadEntryAction {
     }
 
     function __construct(ThreadEntry $thread) {
-        $this->thread = $thread;
+        $this->entry = $thread;
     }
 
     abstract function trigger();
 
     function getTicket() {
-        return $this->thread->getTicket();
+        return $this->entry->getObject();
     }
 
     function isEnabled() {
@@ -1791,8 +1795,8 @@ abstract class ThreadEntryAction {
     function getAjaxUrl($dialog=false) {
         return sprintf('%stickets/%d/thread/%d/%s',
             $dialog ? '#' : 'ajax.php/',
-            $this->thread->getThread()->getObjectId(),
-            $this->thread->getId(),
+            $this->entry->getThread()->getObjectId(),
+            $this->entry->getId(),
             static::getId()
         );
     }
--- a/include/class.thread.php
+++ b/include/class.thread.php
@@ -16,6 +16,7 @@
 **********************************************************************/
 include_once(INCLUDE_DIR.'class.ticket.php');
 include_once(INCLUDE_DIR.'class.draft.php');
+include_once(INCLUDE_DIR.'class.role.php');
 
 //Ticket thread.
 class Thread extends VerySimpleModel {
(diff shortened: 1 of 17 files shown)


Chefkeks:
@greezybacon 
Don't know how often I tell you what great work you (meant in plural - so Peter, Karen and you) do, but need to do it here one more time ;)

That feature looks really good and I totally agree that "Perhaps this should be yet another role privilege.". I can imagine that a lot of folks out there like to define who can edit which types of entries their-selves.

protich:
Awesome! I would second that it should be a role based permission with no special considerations. Help desk admin should just mean that -- access to Admin Panel -- I think starting with 1.7+something we got away from giving special permissions to admins in the Agent Panel.

Minor enhancements requests:

- Ability to resend the entry on edit (checkbox). This will be super useful if you want to resend a response to the enduser after the edit (one step vs. edit then resend).

  - Resent responses should be clear to end user -- append "RESENT" to the subject line?
  - Should we be saving the original signature as part of thread entry? I think yes.

- I wonder if we should support "minor edit" mode -- not for it personally, but should adding a period or space count as an edit?

Chefkeks:
Minor Edit mode would definitely be useful, when it's invisible... So based on a diff what was changed, save the edited message and add the label "edited" or replace the original message.

Just my two cents when it comes to that.

ntozier:
"Resent responses should be clear to end user -- append "RESENT" to the subject line?"
maybe add [RESENT] to the end?   I'd like this type of feature with out editing a post.  There are lots of times where I do not receive a response from the user and literally copy and paste the previous response.  Or simply add "Request #2" to the beginging and then paste the previous message.

"Should we be saving the original signature as part of thread entry? I think yes."
I second that.  We need to make sure that there is no way that a person could send a message that is not logged by the system so that there is accountability.

greezybacon:
@protich I was trying to avoid a database edit to slip the feature in, so we'll just have to see if we're able to slip in a column for the `signature` or and `extra` column. Is it really necessary, though? Should we just use the normal preference when opening the dialog to edit and resend? The agent could change it if they wanted. However, if the system was resending another agent's response, perhaps the signature would be of value. However, in that case, we would need a separate *signature* table so we could reference an ID value.

I'm thinking that this feature should be combined with the *resend* feature, because I think that an agent would almost always want to edit the response when resending it (rather than just click [Resend] and not being able to edit anything).

I'll add the privilege. I thought of it after I posted this PR. I'm thinking it should be considered per department, just like the ticket edit and close privileges. Right?

greezybacon:
@Chefkeks we say *y'all* in the southern US to represent you plural. You'll be an official southerner if you post with that word

Chefkeks:
@greezybacon
Thanks for teaching me, next time I use "y'all" :D

greezybacon:
> I second that. We need to make sure that there is no way that a person could send a message that is not logged by the system so that there is accountability.

I agree with this. Currently, only the most recent edit is preserved. I think if the entry is resent then a new entry *must* be added to the history chain in order to preserve accountability, and will also be required for proper email threading

greezybacon:
Ok, I think I have everything implemented except the "RESENT" addition to the subject line.

robintoy:
This feature is excellent!!!

The amount of times my staff post multiple times to correct what was said so a history and correct / updated version like this is brilliant.

I also like the idea of having an option on each thread to not just edit but re-send the update / thread entry as we often have people ask for the information again and we end up either re-posting it or copying to an email to send on.

include/class.thread.php
```diff
@@ -406,12 +409,24 @@ class ThreadEntry extends VerySimpleModel {
...
 
     var $_headers;
     var $_thread;
     var $_actions;
     var $_attachments;
 
+    static protected $perms = array(
+        self::PERM_EDIT => array(
+            'title' => /* @trans */ 'Edit Thread',
+            'desc'  => /* @trans */ 'Ability to edit thread items of other users',
```
Chefkeks: @greezybacon 
Shouldn't this be "agents" here instead of "users"?

PeopleInside:
Wonderful !!!, i will wait this can be available, fully implemented.

harrisyn:
Nice feature. Can't wait for this to be fully merged. However my help desk have always had a specific need to forward mails to other users other than the original requester  (sometimes with the requester in CC, other times not). This feature will help with resending a specific thread,  it would be great if we could change who the recipient should be and also indicate if the requestor should also receive a copy. 
It would also be great to be able to retrieve a trail of the threads in resending.

Life changing work so far.

PeopleInside:
It will take a long time before it is made available?

greezybacon:
@PeopleInside I hope not

PeopleInside:
When will be possible to edit tickets? From next update?

ntozier:
This feature will be included in version 1.10 when it is released.
Although there were a couple little changes that were discussed yesterday that might be added before its released.

PR #1912, opened 2015-03-27
forms: Add choice lookup support

protich:
Support using value or abbreviation for choice/selection input fields. API calls can now use list value or it's abbreviation instead of internal numeric IDs.

For example API calls can now us "LA" or "Louisiana" as input for a state list field.

### DIFF
--- a/include/class.list.php
+++ b/include/class.list.php
@@ -231,13 +231,15 @@ class DynamicList extends VerySimpleModel implements CustomList {
 
     function getItem($val) {
 
-        $criteria = array('list_id' => $this->getId());
-        if (is_int($val))
-            $criteria['id'] = $val;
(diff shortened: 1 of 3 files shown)


include/class.list.php
```diff
@@ -231,13 +231,22 @@ function getItems($limit=false, $offset=false) {
...
         else
-            $criteria['value'] = $val;
+            $items->filter(Q::any(array('value'=>$val, 'extra' => $val)));
+
+        try {
+
+            return $items->first();
 
-         return DynamicListItem::lookup($criteria);
+        } catch (DoesNotExist $e) {
```
greezybacon: `first` will return NULL if the recordset is empty

PR #1913, opened 2015-03-27
bug: Consider list when looking up list item

protich:
List items are only unique within the list.


greezybacon:
Should this concern the `develop` branch?

PR #1914, opened 2015-03-27
email: Some bounce notices actually have rich content

greezybacon:
Here's an example email structure:
```
multipart/mixed
 - multipart/report; delivery-status
   - multipart/alternative
     - text/plain
     - text/html
   - message/delivery-status
   - message/rfc822
```

The previous code would only find the body if the email main headers had: `Content-Type: multipart/report; report-type="delivery-status"`. In such a case it would scan for a plain/text body.

This patch will scan for the usual body if the scan for the body as usual if the report scan did not find anything.

Also, output errors to stderr when running API from the command line

### DIFF
--- a/include/class.api.php
+++ b/include/class.api.php
@@ -297,7 +297,12 @@ class ApiController {
             $msg.="\n*[".$_SERVER['HTTP_X_API_KEY']."]*\n";
         $ost->logWarning(__('API Error')." ($code)", $msg, false);
 
-        $this->response($code, $error); //Responder should exit...
+        if (PHP_SAPI == 'cli') {
+            fwrite(STDERR, "({$code}) $error\n");
+        }
+        else {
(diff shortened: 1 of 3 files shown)


greezybacon:
This could also be fixed by setting the recurse value in ::getDeliveryStatusMessage() to 2

include/class.api.php
```diff
@@ -297,7 +297,12 @@ function exerr($code, $error='') {
             $msg.="\n*[".$_SERVER['HTTP_X_API_KEY']."]*\n";
         $ost->logWarning(__('API Error')." ($code)", $msg, false);
 
-        $this->response($code, $error); //Responder should exit...
+        if (PHP_SAPI == 'cli') {
+            fwrite(STDERR, "({$code}) $error\n");
```
greezybacon: Maybe this should go into ::response()

PR #1921, opened 2015-03-31
email: Add warning for archive folder for POP

greezybacon:
This patch adds an error to the page if unable to configure the archive folder and POP was selected as the mailbox protocol. Generally, POP servers do not support anything but the INBOX

### DIFF
--- a/include/class.email.php
+++ b/include/class.email.php
@@ -296,8 +296,12 @@ class Email {
 
             if(!isset($vars['postfetch']))
                 $errors['postfetch']=__('Indicate what to do with fetched emails');
-            elseif(!strcasecmp($vars['postfetch'],'archive') && !$vars['mail_archivefolder'] )
-                $errors['postfetch']=__('Valid folder required');
+            elseif(!strcasecmp($vars['postfetch'],'archive')) {
+                if ($vars['mail_protocol'] == 'POP')
+                    $errors['postfetch'] =  __('POP mail servers do not support folders');
+                elseif (!$vars['mail_archivefolder'])
+                    $errors['postfetch'] = __('Valid folder required');
+            }
         }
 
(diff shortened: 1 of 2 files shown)


protich:
@greezybacon -- something about error redirection on post fetch archiving doesn't sit well with me. I think it will be a little confusing. Below is my proposal -- error out early and put the error on the post fetch action.

```git
diff --git a/include/class.email.php b/include/class.email.php
index d38806f..26cd686 100644
--- a/include/class.email.php
+++ b/include/class.email.php
@@ -296,8 +296,12 @@ class Email {
 
             if(!isset($vars['postfetch']))
                 $errors['postfetch']=__('Indicate what to do with fetched emails');
-            elseif(!strcasecmp($vars['postfetch'],'archive') && !$vars['mail_archivefolder'] )
-                $errors['postfetch']=__('Valid folder required');
+            elseif(!strcasecmp($vars['postfetch'],'archive')) {
+                if ($vars['mail_protocol'] == 'POP')
+                    $errors['postfetch'] =  __('POP mail servers do not support folders');
+                elseif (!$vars['mail_archivefolder'])
+                    $errors['postfetch'] = __('Valid folder required');
+            }
         }
 
         if($vars['smtp_active']) {
```

greezybacon:
I don't disagree. However, are we sure that POP does in fact NOT support folder archiving?

greezybacon:
[Looks like it doesn't](http://stackoverflow.com/a/4333434/1025836)

greezybacon:
Updated

infectormp:
Maybe better disable or hide archive folder edit box if POP selected?

greezybacon:
@infectormp I think that would make a nice enhancement proposal

PR #1922, opened 2015-03-31
Unassign on transfer

protich:
Unassign tickets on transfer when the target department has assignment restriction and the assigned staff is not a member.

Disable claim (quick self-assignment) when above restriction is in effect.


akash65:
can you tell how to transfer a ticket to department?

PR #1923, opened 2015-03-31
Unassign on transfer

protich:
Unassign tickets on transfer when the target department has assignment restriction and the assigned staff is not a member. Also disable claim (quick self-assignment) when the restriction is in effect.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -1559,10 +1559,20 @@ class Ticket {
         if($this->isClosed()) $this->reopen();
 
         $this->reload();
+        $dept = $this->getDept();
 
         // Set SLA of the new department
         if(!$this->getSLAId() || $this->getSLA()->isTransient())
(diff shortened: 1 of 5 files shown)


include/class.ticket.php
```diff
@@ -1616,6 +1626,21 @@ function transfer($deptId, $comments, $alert = true) {
...
+        global $thisstaff;
+
+        if (!$thisstaff || !$this->isOpen() || $this->isAssigned())
+            return false;
+
+        $dept = $this->getDept();
+        if ($dept->assignMembersOnly() && !$dept->isMember($thisstaff))
+            return false;
+
+        $comments = sprintf(__('Ticket claimed by %s'), $thisstaff->getName());
```
greezybacon: I have been using `_S` for these statements where the system primary language should be used instead of the current user's preference

PR #1931, opened 2015-04-06
List items import/export

protich:
Add ability to import/export list items


include/class.list.php
```diff
@@ -229,19 +229,27 @@ function getItems($limit=false, $offset=false) {
...
+            $items->filter(array('id' => $val));
+        elseif ($extra)
+            $items->filter(array('extra' => $val));
         else
-            $criteria['value'] = $val;
+            $items->filter(array('value' => $val));
+
 
-         return DynamicListItem::lookup($criteria);
+        return $items->one();
```
greezybacon: This will throw an exception in `develop-next`. Seems like we should catch it here

include/class.list.php
```diff
@@ -438,6 +446,134 @@ static function getSelections() {
         return $selections;
     }
 
+    function importCsv($stream, $defaults=array()) {
```
greezybacon: We **really** need to genericize and reuse this code

include/class.list.php
protich: The "temp" commit is temporary. I simply rigged it to work for current import needs.

include/class.list.php
protich: Agreed -- this was from your list management commit on develop-next.

protich:
Ready for review.

greezybacon:
Since this is a new feature, should we rebase against `develop-next`

jdelhome3578:
@protich @greezybacon Was this re-based?

PR #1938, opened 2015-04-08
Feature/ip patterns

yadimon:
ability to use regex paterrn for ip

pattern checkbox

i dont know all the source code that good, and not sure about right english, so pls review.


Remo:
I was about to build something similar using the approach shown in the link below when I luckily saw that you've done something similar, thanks for this PR!

http://stackoverflow.com/questions/10421613/match-ipv4-address-given-ip-range-mask

PR #1948, opened 2015-04-10
email: Match collaborator by email address too

greezybacon:
If a collaborator on a ticket responds to an email NOT originating from osTicket, the system did not consider the replying user's email address. Previously, the system would attempt to identify a collaborator from the tagged Message-ID header (or References header in versions previous to v1.9.6).

This patch will also attempt to match the collaborator by the From email address header, and will post a message to the ticket rather than perhaps falling back to an internal note.


protich:
Rebase required.

greezybacon:
Merged in #1983

PR #1983, opened 2015-04-20
email: Support agents emailing as collaborators

greezybacon:


include/class.thread.php
```diff
@@ -745,10 +745,20 @@ function postEmail($mailinfo) {
...
+            && ($C = Collaborator::lookup(array(
+                'ticketId' => $ticket->getId(), 'userId' => $E->user_id
+            )))
+        ) {
+            $vars['userId'] = $C->getUserId();
+            $vars['message'] = $body;
             return $ticket->postMessage($vars, 'Email');
         }
-        // XXX: Consider collaborator role
         elseif ($mailinfo['staffId']
```
greezybacon: I think this should move up so that we check the $vars data before running a database query

PR #1993, opened 2015-04-23
users: Update information on CSV import

greezybacon:
This patch adds a feature to the import process, so that if a user already exists (identified by its email address), then the name and other custom data are updated with the information in the CSV record being imported.

### Outstanding
* [x] Touch DynamicFormEntry::updated when custom data is updated

### DIFF
--- a/include/class.user.php
+++ b/include/class.user.php
@@ -197,18 +197,8 @@ class User extends UserModel {
             }
         }
         elseif ($update) {
-            foreach ($user->getDynamicData() as $entry) {
-                // FIXME: Convert this to ->form for `develop-next`
-                if ($entry->getForm()->type == 'U') {
-                    $entry->setSource($vars);
-                    $entry->save();
-                }
-            }
-            // Update name
-            if (isset($vars['name'])) {
-                $user->name = $vars['name'];
-                $user->save();
-            }
+            $errors = array();
+            $user->updateInfo($vars, $errors, true);
         }
 
         return $user;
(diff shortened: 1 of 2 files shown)


include/class.user.php
```diff
@@ -196,6 +196,20 @@ static function fromVars($vars) {
                 return null;
             }
         }
+        elseif ($update) {
+            foreach ($user->getDynamicData() as $entry) {
```
greezybacon: Use `$this->updateInfo()`

Remo:
Shouldn't this code result in a user with the name `bbb`?

```php
$data = [
    'name' => 'aaa',
    'email' => 'info@aaa.ch',
    'phone' => 123,
];
$user = User::fromVars($data, true);

$data = [
    'name' => 'bbb',
    'email' => 'info@aaa.ch',
    'phone' => 456,
];
$user = User::fromVars($data, true);
```

ntozier:
email is a unique field, so no.

Remo:
Yes, it is, but this line added an implicit update, didn't it?
https://github.com/osTicket/osTicket-1.8/pull/1993/files#diff-051ae6fc9b5c9f4276374b24cb7bd5b6R166

I would expect that the user gets updated after merging this change..

greezybacon:
@Remo we've had significant troubles with the custom data system including strange issues like this. We made some very significant changes in the `develop-next` branch to remove most of the "magic" employed to make it work in this version. Could you try your code against the code in the `develop-next` branch to see if we've already fixed this and should just consider back-porting the changes?

Remo:
Looks good, will run a more sophisticated test on Thurday

Remo:
@greezybacon the implicit update didn't work with `develop-next`, but writing multiple fields at once worked fine

PR #2031, opened 2015-05-08
session: Add support for Memcache as a session backend

greezybacon:
This adds support for Memcache server(s) as a session storage backend. This is useful for higher traffic sites as it removes a database write per request. Therefore, read requests can be truly read; without database writes. Multiple Memcache servers are supported and used redundantly so that as long as at least one Memcache server remains reachable, session data will not be lost before it expires.

### DIFF
--- a/include/class.ostsession.php
+++ b/include/class.ostsession.php
@@ -66,6 +66,7 @@ class osTicketSession {
         }
         catch (Exception $x) {
             // Use the database for sessions
+            trigger_error($x->getMessage(), E_USER_WARNING);
             $this->backend = new self::$backends['db']($this->ttl);
         }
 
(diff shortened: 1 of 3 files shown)


include/class.ostsession.php
```diff
@@ -49,24 +55,40 @@ function osTicketSession($ttl=0){
...
+
+        try {
+            $bk = SESSION_BACKEND;
+            if (!class_exists(self::$backends[$bk]))
+                $bk = 'db';
+            $this->backend = new self::$backends[$bk]($this->ttl);
+        }
+        catch (Exception $x) {
+            // Use the database for sessions
+            $this->backend = new self::$backends['db']($this->ttl);
```
greezybacon: Emit a PHP warning here

PR #2033, opened 2015-05-08
fast fix for deployment on windows #2032

### LINKED ISSUE #2032: cli deploy script on windows
If you use windows, and want to deploy your git repository (php.exe -f manage.php deploy some\install\folder), the unwanted directories (like .git or setup) are not excluded, because the pattern/path comparision in the deploy.php line 153, uses pure root path for pattern. The comparision fails because of backslash "\" in the pattern.
e.g. fnmatch("C:\wwwroot","C:\wwwroot") returns false.

my hotfix is: 
add one $root var more, $rootPattern. just for exclude patterns, and use it to build $exclusions array

possible need command line argument, to exclude more directories (e.g. ide artifacts like .idea directory etc.)

yadimon:
Fixes #2032

### DIFF
--- a/setup/cli/modules/deploy.php
+++ b/setup/cli/modules/deploy.php
@@ -151,11 +151,12 @@ class Deployment extends Unpacker {
 
         # Locate the upload folder
         $root = $this->find_root_folder();
+        $rootPattern = str_replace("\\","\\\\", $this->find_root_folder()); //need for windows case
 
-        $exclusions = array("$root/include", "$root/.git*",
+        $exclusions = array("$rootPattern/include", "$rootPattern/.git*",
             "*.sw[a-z]","*.md", "*.txt");
         if (!$options['setup'])
-            $exclusions[] = "$root/setup";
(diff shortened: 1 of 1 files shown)


setup/cli/modules/deploy.php
```diff
@@ -151,11 +151,12 @@ function run($args, $options) {
 
         # Locate the upload folder
         $root = $this->find_root_folder();
+        $rootPattern = str_replace("\\","\\\\", $this->find_root_folder()); //need for windows case
```
yakatz: How about `$rootPattern = str_replace("\\","\\\\", $root);` so you don't need to call the find function again?

greezybacon:
Can anyone confirm this fixes deployment issues on Windows?

mfelber:
@greezybacon Yes, I can confirm that this fix changes the behavior as expected. The .git and setup-folder will not be deployed anymore when you use the deploy script without the --setup parameter.

PR #2035, opened 2015-05-08
1.8.x

danrhodes:
Email address validation via DNS MX Record lookup.

/includes/class.validator.php email validation changed to the following.

function is_email($email) {
$exp = "^[a-z\'0-9]+([._-][a-z\'0-9]+)*@([a-z0-9]+([._-][a-z0-9]+))+$";
if(eregi($exp,$email)){
	if(checkdnsrr(array_pop(explode("@",$email)),"MX")){
			return true;
		}else{
			return false;
		}
		}else{
			return false;
	}
}

### DIFF
--- a/WHATSNEW.md
+++ b/WHATSNEW.md
@@ -1,3 +1,15 @@
+osTicket v1.8.5
+===============
+### Improvements
+  * Fix missing `%{recipient}` variable used in canned reply (filters) (#1047)
+  * Fix `%{ticket.close_date}` variable in email message templates (#1090)
+  * Fix premature session expiration (#1111)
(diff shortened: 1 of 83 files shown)


greezybacon:
Why is this a pull request?

ntozier:
He's trying to do this:
http://www.osticket.com/forum/discussion/85455/email-domain-validation#latest

greezybacon:
Fair enough. I've created #2042 to handle this. Note that this will not be backported to v1.8 as the `checkdnsrr` function is not reliably available until PHP 5.3, and v1.8 on requires PHP 4.something

PR #2041, opened 2015-05-11
Fix for staff submitting ticket by email

hawkhead99:
I noticed there is a bug when class.thread.php has to fall back to the ticket number in the subject line.  If the user is a staff member and they send a note to a ticket via email without including the existing thread (just the ticket number), it will create a new ticket because the fall back logic does not account for staff...  This fix would correct that

Tested locally

### DIFF
--- a/include/class.thread.php
+++ b/include/class.thread.php
@@ -993,14 +993,19 @@ class ThreadEntry {
                 //Lookup by ticket number
                 && ($ticket = Ticket::lookupByNumber($match[1]))
                 //Lookup the user using the email address
-                && ($user = User::lookup(array('emails__address' => $mailinfo['email'])))) {
+                && ($user = User::lookup(array('emails__address' => $mailinfo['email'])))
+                && ($staff = Staff::getIdByEmail($mailinfo['email']))) {
             //We have a valid ticket and user
-            if ($ticket->getUserId() == $user->getId() //owner
+            if (!empty($staff) || $ticket->getUserId() == $user->getId() //owner
                     ||  ($c = Collaborator::lookup( // check if collaborator
                             array('userId' => $user->getId(),
                                   'ticketId' => $ticket->getId())))) {
 
-                $mailinfo['userId'] = $user->getId();
+                if (empty($staff))
+	                $mailinfo['userId'] = $user->getId();
+                else
+	                $mailinfo['staffId'] = $staff;
+	                
                 return $ticket->getLastMessage();
             }
         }


include/class.thread.php
```diff
@@ -993,14 +993,19 @@ function lookupByEmailHeaders(&$mailinfo, &$seen=false) {
                 //Lookup by ticket number
                 && ($ticket = Ticket::lookupByNumber($match[1]))
                 //Lookup the user using the email address
-                && ($user = User::lookup(array('emails__address' => $mailinfo['email'])))) {
+                && ($user = User::lookup(array('emails__address' => $mailinfo['email'])))
+                && ($staff = Staff::getIdByEmail($mailinfo['email']))) {
```
greezybacon: Shouldn't this be something like
```php
if ( ... 
    && (
           ($user = User::lookup(array('emails__address' => $mailinfo['email'])))
        || ($staff = Staff::getIdByEmail($mailinfo['email']))
    )
) {
```
Since it's unlikely that the user email will have a hit for both a User and a Staff?

hawkhead99: It appears $user always tends to return on object even for staff members...  But you are right... there would be a bug if it was a user and not a staff member....
$staff = Staff::getIdByEmail($mailinfo['email']  should just be in the if statement, ill fix it

hawkhead99:
New commit that accounts for greezybacon's comment -

greezybacon - I can account for adding the staff variable in the "or" as you suggested as well if you like, but in testing the fix and briefly looking through the lookup code, I could not find a time where $user ever comes back false when a staff member submits the email - it always seems to pull back an id in the object as well (just won't match the ticket owner ID or a collaborator ID - or what the staff ID would be)

include/class.thread.php
greezybacon: Perhaps I should have slept at a Holiday Inn Express last night; but the `User::lookup` should definitely fail if no end user currently exists with that email address. Perhaps your staff members already all have corresponding user accounts?

hawkhead99: I missed this comment...  Hrm...  that could definitely be the case..  We have been using osticket for over 3 years now, more then likely all staff members have a user id in there by now...

I'll account for the or as well

hawkhead99: I just re-commited.  The new commit accounts for both scenarios (when the staff might have a user ID in the system and when they do not)

hawkhead99:
New commit accounts for staff having a user ID and also in the case the staff member does not have a user ID

hawkhead99:
@greezybacon this one should be good to go with the change you recommended (except put staff first to consider a staff can be a user - so want to check that condition first) - unless you already have it resolved for the v 1.10 release- then I can just close it :)

greezybacon:
@hawkhead99 I really appreciate your code. I hope to have time for a formal review this week

PR #2042, opened 2015-05-11
validator: Verify MX record of email addresses

greezybacon:
This allows detection of some incorrectly-typed email addresses before tickets, users, and agents are created.

It also adds a warning to the ticket view page to alert agents of invalid, in-use email addresses. This may happen for emails processed into tickets when the `Reply-To` header is used, as well as for old tickets created before this patch.

### DIFF
--- a/scp/banlist.php
+++ b/scp/banlist.php
@@ -33,7 +33,7 @@ if($_POST && !$errors && $filter){
         case 'update':
             if(!$rule){
                 $errors['err']=sprintf(__('%s: Unknown or invalid'), __('ban rule'));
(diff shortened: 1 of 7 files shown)


scp/banlist.php
```diff
@@ -52,7 +52,7 @@
         case 'add':
             if(!$filter) {
                 $errors['err']=sprintf(__('%s: Unknown or invalid'), __('ban list'));
-            }elseif(!$_POST['val'] || !Validator::is_email($_POST['val'])) {
+            }elseif(!$_POST['val'] || !Validator::is_valid_email($_POST['val'])) {
```
greezybacon: This should be allowed (`test@example.com`)

PR #2043, opened 2015-05-11
ui: Add floating bar to make ticket options more accessible

greezybacon:
This adds a floating bar to the screen when scrolling further down in the ticket queue or the ticket view pages.

### Screenshots
screen shot 2015-05-11 at 12 55 39 pm

screen shot 2015-05-11 at 12 59 06 pm


Chefkeks:
Nice :D :+1:

hawkhead99:
That is a really neat idea!

tha023:
Can it be implemented like i tried to explain here
http://osticket.com/forum/discussion/85650/make-top-always-visible-when-scrolling-down-a-ticket?new=1
So not a floating bar, but a solid top bar? or is it by chance possible to make us choose between both.
Anyway! thanks for your hard work!

PR #1419, opened 2014-10-22
Feaure / Ticket: Time tracking

robintoy:
# Time Tracking Mod / Change
**Updated: 22/06/2016**

Re-written and complete Time Tracking enhancement.
**Requires Version v1.10-rc3**

This enhancement allows you to enable functions for adding just time to a ticket or enable functions to allow adding time to the thread of the ticket along with the type of time spent.

The time types are a custom list that can be modified to your business requirements giving your full billing and reporting abilities.

### To Do List
- [ ] Rebase pull to develop-next (v.1.11.x) as required for feature submission
- [ ] Remove the **current time spend** display from post reply and internal note to clean up display
- [ ] Remove total from hardware table and make this calculated
- [ ] Remove Add time Tab
- [ ] Remove total time from ticket table and make this amount calculated on ticket load
- [ ] Hardware changed from additional page opening to a new tab like Tasks are in V1.10
- [ ] Look at changing it from time to expense
- [ ] Contract time allocations added to organizations
- [ ] Organization billing reports that so decreasing time and turns this to billable once contract time used.

### Pictures
New pictures to come

### Base file install
To be updated

### Database upgrade (SQL commends)
Run the following commands in something like PHPMyAdmin or MySQL CLI
To be updated

### 22/06/2016
The old v1.9.14 code has now been moved to a zip file with full instruction to our website https://www.strobe-it.co.uk/services/opensource/
The code on this page will now be updated to v1.10.x to keep inline with the development process of osTicket

PR #2052, opened 2015-05-13
deploy: Add concept of a MANIFEST to the deployment process

greezybacon:
This allows files which are edited on the fly during deployment to be tracked so that they are not unnecessarily deployed again in the next deployment run. It also allows for more creative deployment strategies using something like `git ls-files -s`

Additionally, the `.MANIFEST` file could be signed and verified with DNS, which, when combined with the unpack script could provide a chain of custody for the source code.


greezybacon:
I've updated this and it works on the first deployment now

digitext:
Hello,
I am a new user of ostickets and trying to see if I can use the time option for tickets .. can you please suggest where to download what, I saw some instruction on 1419 but then there were too many instructions..
Awaiting your

greezybacon:
@digitext I think you'll get more help if you post in #1419

PR #2057, opened 2015-05-13
when opening a ticket, no longer auto-scrolls to the bottom or Reply box

hawkhead99:
This fix resolves the issue of when you open a ticket, it scrolls to the bottom of the page.

It also keeps from breaking all the other pages that use the focus functionality (exs. the main staff tickets screen to auto-focus on the search box, the ticket edit screen).


hawkhead99:
closed/deleted branch  already resolved for V.1.10

PR #2079, opened 2015-05-17
file: Allow different names for duplicated content

greezybacon:
This patch (which may need further development), introduces an extra field to the %attachment table to store a file's name in the event that two files with the same content are stored in the database with differing names. In such a case, the name is stored in the %attachment table. This allows the attachment to specify a different name each time the same file is attached to something in the system.

It also would allow the file's name to be edited after attaching, and it would not affect any other attachment's names with the same, deduplicated content.

### DIFF
--- a/include/class.attachment.php
+++ b/include/class.attachment.php
@@ -56,6 +56,10 @@ class Attachment extends VerySimpleModel {
         return $this->file;
     }
 
+    function getFilename() {
+        return $this->name ?: $this->file->name;
+    }
+
     function getHashtable() {
         return $this->ht;
     }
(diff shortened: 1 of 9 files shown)


include/staff/templates/thread-entries.tmpl.php
```diff
@@ -66,8 +66,8 @@ class="tmeta faded title"><?php
                             Format::file_size($A->file->size));
 ?>
             <a class="Icon file no-pjax" href="<?php echo $A->file->getDownloadUrl();
-                ?>" download="<?php echo Format::htmlchars($A->file->name); ?>"
-                target="_blank"><?php echo Format::htmlchars($A->file->name);
+                ?>" download="<?php echo Format::htmlchars($A->getFilename()); ?>"
+                target="_blank"><?php echo Format::htmlchars($A->getFilename());
```
greezybacon: I think this change is needed on the client interface too

PR #2121, opened 2015-06-01
thread: Multi-thread rendering

protich:
This adds the concept of independent multi-thread rendering.


include/client/templates/thread-entries.tmpl.php
```diff
@@ -6,6 +6,10 @@
 $events->rewind();
 $event = $events->current();
 
+$container = $options['container'] ?: ('thread-'.$this->getId());
```
greezybacon: Seems like `html-id` would be a more true option name

Issue #2124, opened 2015-06-01
send email silently fails


ericLemanissier:
In our production configuration, SMTP setting does not work, so we use the "PHP mail function". Staff members or users having special characters in their name (eg î or ę) never receive email notification from OSTicket. This is very problematic, especially because nothing is logged.

ericLemanissier:
I just submitted a pull request which solves the special characters problem, but not the logs missing: https://github.com/osTicket/osTicket-1.8/pull/2125

PR #2056, opened 2015-05-13
Stabilize v1.10 release

greezybacon:
### Outstanding
* [x] Ticket listing for search by user (followed from ticket-view) incorrect / incomplete
* [x] Teaching Karen to create outstanding issues
* [x] Allow translations of the HTML hints
* [ ] Add a message to the "upload a new logo" widget, like:  `osTicket recommends a PNG logo with a size of about 380px by 64px and a transparent background` 
* [ ] Cascade dashboard statistics (for sub-departments)
* [x] PJAX back navigation doesn't seem to work well
* [x] PJAX navigation from user directory breaks
* [x] Verify layout for single and multi-lingual systems where multi-lingual content is displayed
* [x] Dashboard statistics crash
* [x] Draft delete is broken
* [x] Do something with FAQ view -> manage access link
* [x] Fix image paste+upload in Chrome
* [ ] Fix image annotate with S3 plugin
* [x] When updating canned responses with attachments, saving changes causes white screen (line 53)
* [ ] Incorrect navigation after Ticket Delete
* [ ] Dress up search fields in advanced search (add ticket.user and ticket.user.organization)
* [x] `choose them` doesn't seem to work for the file upload utility
* [x] Remove requirement to provide comments on assign and transfer
* [x] Consider role to view thread edit history
* [x] Auto expand context typeahead if clicking on arrow
* [x] Link dropdown in Redactor [insert link] modal is empty
* [ ] Deleting a form field after clearing the label field fails
* [x] `strpos` warning on agent listing page
* [x] Force password change did not work correctly (forced and forced us over and over again to change password - even we had changed it already several times)
* [x] Remove Transfer and Reassign from bottom of Ticket and relocate to top Quick Buttons
* [ ] When reassigning  or transferring tickets that are currently assigned, add check box to remove initial assignment 
* [x] Inline images are not being sent out as attachments in email
* [ ] When editing/resending thread entry, attachments are not included
* [ ] Add ability to include attachments when editing/resending thread entry
* [ ] Looks like duplicate records in %attachment for ('H', 0) for the same records recorded about the same time for non-zero thread-id numbers.
* [x] Links to "Access Control Templates" are broken
* [ ] Adding new filter with validation errors clears configured actions

### Lists
* [x] Basic search does not work

### Collaborators
 * [ ] Update recipients and participants count on save
 * [ ] Use InstrumentedList broker and UserList to make getting names, emails...etc easier

### UI
* [x] Add tooltip for attachment count and collaborator count in ticket queue
[...]

PR #2125, opened 2015-06-01
Handle special characters in user/staff name

ericLemanissier:
for user/staff name with special characters, email notification are never sent (at least with php mail function) because the name is not encoded. This change adds encoding of the staff/user name in the "to" field of the email to be sent

### DIFF
--- a/include/class.mailer.php
+++ b/include/class.mailer.php
@@ -239,7 +239,7 @@ class Mailer {
             // Add personal name if available
             if (is_callable(array($to, 'getName'))) {
                 $to = sprintf('"%s" <%s>',
-                    $to->getName()->getOriginal(), $to->getEmail()
+                    "=?UTF-8?B?".base64_encode($to->getName()->getOriginal())."?=", $to->getEmail()
                 );
             }
             else {


ericLemanissier:
is also reported the issue here #2124

greezybacon:
This was also fixed in 98fef45cf1dc7ed7208a1c7c40a0b0702f762014 toward `develop-next`. I like its approach better as it relies on the MIME generator to encode the header based on the encoding settings, and should also conform better to RFC standards

ericLemanissier:
Using MIME generator is definitely a better fix.
question not directly linked : is there a roadmap or something showing the progress of next osticket version release ?

greezybacon:
The release page features the major new stuff. #2056 outlines the outstanding known issues

greezybacon:
Thanks again for the suggestion. I backported the other commit into 857dd22

tmp2000:
As I changed the code as the 857dd22, the problem still exist.I am using osTicket-1.9.8.1 .
And the error message is from System Logs:
Mailer Error
Unable to email via SMTP:mailserver***.com:25 [support@mailserver***.com] Validation failed for: ""Sabine Klünder"" .
The error message is from include/pear/Mail/RFC822.php line 509 .
Did anyone have the same problem?

PR #2128, opened 2015-06-02
log php mail error

ericLemanissier:
When the php mail function is used to send email, this change adds error logging if an error is detected by PEAR

### DIFF
--- a/include/class.mailer.php
+++ b/include/class.mailer.php
@@ -514,7 +514,8 @@ class Mailer {
             // Force reconnect on next ->send()
             unset($smtp_connections[$key]);
 
-            $alert=sprintf(_S("Unable to email via SMTP:%1\$s:%2\$d [%3\$s]\n\n%4\$s\n"),
+            $alert=_S("Unable to email via SMTP")
+                    .sprintf(":%1\$s:%2\$d [%3\$s]\n\n%4\$s\n",
                     $smtp['host'], $smtp['port'], $smtp['username'], $result->getMessage());
             $this->logError($alert);
         }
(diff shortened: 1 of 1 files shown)


greezybacon:
I really like this. Thanks for the PR

include/class.mailer.php
```diff
@@ -523,8 +523,14 @@ function($match) use ($domain, $mime, $self) {
         $mail = mail::factory('mail');
         // Ensure the To: header is properly encoded.
         $to = $headers['To'];
-        return PEAR::isError($mail->send($to, $headers, $body))?false:$messageId;
-
+        $result = $mail->send($to, $headers, $body);
+        if(!PEAR::isError($result))
+            return $messageId;
+
+        $alert=sprintf(__("Unable to email via php mail function:%1\$s\n\n%2\$s\n"),
```
greezybacon: A note on the translation functions: `_S` is used to translate to the system primary language, `__` translates to the language of the request. They might be different, and messages logged to the system log should be in the primary language.

Reviewing the code in the `::send()` method currently, it looks like we have this issue in several other places.

include/class.mailer.php
```diff
@@ -514,7 +514,7 @@ function($match) use ($domain, $mime, $self) {
             // Force reconnect on next ->send()
             unset($smtp_connections[$key]);
 
-            $alert=sprintf(__("Unable to email via SMTP:%1\$s:%2\$d [%3\$s]\n\n%4\$s\n"),
+            $alert=sprintf(_S("Unable to email via SMTP:%1\$s:%2\$d [%3\$s]\n\n%4\$s\n"),
```
greezybacon: Could we change this to
```php
$alert=_S("Unable to email via SMTP")
    . sprintf(":%1\$s:%2\$d [%3\$s]\n\n%4\$s\n",
```
So that the translators don't have to deals with the string of random token and newline chars?

PR #2135, opened 2015-06-03
csv: Fix CSV import with whitespace in header

### LINKED ISSUE #2123: csv import of users: some little issues
Issue 1)

I have noticed when  I import in osticket 1.9.8.1 users via a csv-file, the header line in the file will be imported as a user too:

csv-import

Issue 2)

Only the values from the columns „email", „name" and „phone" will be imported if the users don’t exist in the database.

If the users exist in the database also other values from columns like „notes" or custom created fields will be imported/updated.

This is the content of my test csv-file:

`Email, Name, Phone, Notes`
`email1@no-reply.de, Tom Reagan, 0123456789, Testnote1`
`email2@no-reply.de , Johnny Caspar, 0123456790, Testnote2`
`email3@no-reply.de , Eddie Dane, 0123456791, Testnote3`
`email4@no-reply.de , Verna Bernbaum, 0123456792, Testnote4`

You can reproduce it this way:

When you import this list the first time, only the values from the columns „email", „name"and „phone" will be imported.

After importing it the second time also the „notes“ values will be imported.

greezybacon:
Somewhat fixes #2123 

**Outstanding**
* [ ] Fix saving of some other custom fields on create including the notes

### DIFF
--- a/include/class.user.php
+++ b/include/class.user.php
@@ -394,6 +394,7 @@ class User extends UserModel {
         else {
             $headers = array();
             foreach ($data as $h) {
+                $h = trim($h);
                 $found = false;
                 foreach ($all_fields as $f) {
                     if (in_array(mb_strtolower($h), array(


greezybacon:
This was addressed with the new CSV importing system

PR #2145, opened 2015-06-08
html: Balancing is important

greezybacon:
9ae01bf208060528a05164d22817e776fa75c7ef turned off HTML balancing, however, quoted response removal will create unbalanced HTML. That HTML will be stored unbalanced and will require extra processing when displayed on the ticket thread.

This patch adds balancing support using libxml if enabled, which will balance HTML without corrupting the HTML like htmLawed will with balancing enabled.

### DIFF
--- a/include/class.format.php
+++ b/include/class.format.php
@@ -115,11 +115,79 @@ class Format {
         return $len ? wordwrap($text, $len, "\n", true) : $text;
     }
 
-    function html($html, $config=array('balance'=>1)) {
+    function html_balance($html, $remove_empty=true) {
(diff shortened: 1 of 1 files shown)


include/class.format.php
```diff
@@ -235,7 +303,7 @@ function localizeInlineImages($text) {
     function sanitize($text, $striptags=false) {
 
         //balance and neutralize unsafe tags.
-        $text = Format::safe_html($text);
+        $text = Format::safe_html($text, $striptags ? 1 : 0);
```
greezybacon: I think the default here is sufficient. That is, always balance

PR #2152, opened 2015-06-09
Pending Ticket Auto Close 1.9.x

jdelhome3578:
This will automatically close Pending Status tickets based on a database value entered on the ticket settings page.
To do:
- [ ] remove pending ticket queue and stats after #2577 implementation

Need help:
- [ ] set status using status name instead of id

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -3025,6 +3025,7 @@ class Ticket {
 
         }
    }
+
    // Close Pending tickets based on Pending Ticket Auto-Close from config
    function ClosePending() {
 	global $cfg;
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -2361,6 +2361,13 @@ class Ticket {
                 .'WHERE ticket.isanswered = 1 '
                 . $where
 
+                .'UNION SELECT \'pending\', count( ticket.ticket_id ) AS tickets '
+                .'FROM ' . TICKET_TABLE . ' ticket '
+                .'INNER JOIN '.TICKET_STATUS_TABLE. ' status
+                    ON (ticket.status_id=status.id
+                            AND status.name=\'pending\') '
+                . $where
+
                 .'UNION SELECT \'overdue\', count( ticket.ticket_id ) AS tickets '
                 .'FROM ' . TICKET_TABLE . ' ticket '
                 .'INNER JOIN '.TICKET_STATUS_TABLE. ' status
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -3030,18 +3030,20 @@ class Ticket {
    function ClosePending() {
 	global $cfg;
 	
-	$days = $cfg->getAutoCloseGrace();
+	$grace = $cfg->getAutoCloseGrace();
 
-	if($days != 0){ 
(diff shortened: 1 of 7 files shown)


infectormp:
:+1:

infectormp:
You use new status Pending, but why you don't use Resolved status?

jdelhome3578:
We were using Pending way before resolved ever became.

infectormp:
Maybe better make status configurable?

jdelhome3578:
Maybe @greezybacon or @protich have some ideas on that.

ntozier:
I'm biased [since I wrote the Pending/autoclose for 1.6/1.7] but I think that this is a long over due feature.

Pending to me means that its pending customer response (and may or may not be resolved)
Resolved to me means that the issue is resolved.
Or at least that's how we have used them here since 1.6.

jdelhome3578:
We use this Pending status when we are waiting for a response to confirm a fix has worked. If they don't respond this closes the ticket for us.

jdelhome3578:
@greezybacon @protich Would this be something to implement in 1.10 or a plugin?

jdelhome3578:
When custom queues #2577 is merged this code can be reduced.

include/class.ticket.php
```diff
@@ -3019,5 +3026,26 @@ function checkOverdue() {
...
+	if($grace == 1){$dayplural = 'day';} else {$dayplural = 'days';}
+
+	// select all tickets marked as 'pending' where updated is older than ($days) days ago
+	$sql  = 'SELECT ticket_id FROM ' .TICKET_TABLE .' ticket '
+		 .' INNER JOIN '.TICKET_STATUS_TABLE.' status ON (status.id=ticket.status_id AND status.name = "pending") '
+	     .' AND TIME_TO_SEC(TIMEDIFF(NOW(),ticket.updated))>='.$grace.'*86400';
+
+        if(($res=db_query($sql)) && db_num_rows($res)) {
+            while(list($id)=db_fetch_row($res)) {
+                if(($ticket=Ticket::lookup($id)) && $ticket->setStatus('3', 'Ticket Closed by the SYSTEM after '.$grace.' '.$dayplural.' of no activity.', false));
```
jdelhome3578: @greezybacon I am unable to get this to work with `setStatus('closed', ...` it looks like it should work based off of ` function setState($state, $alerts=false) {
        switch(strtolower($state)) {
            case 'open':
                return $this->setStatus('open');
                break;
            case 'closed':
                return $this->setStatus('closed');
                break;`

jdelhome3578:
@ntozier any ideas on the code comments

jdelhome3578:
@greezybacon can you provide any assistance with the `setstatus()` function?

jdelhome3578:
> Maybe better make status configurable?

@infectormp Made an update in #2865 with your suggestion.

jdelhome3578:
For older version 1.9.x 
see PR #2917 for 1.10 version

include/class.ticket.php
mckaygerhard: any feedback bout use the setState ?

mckaygerhard:
any feedback about working or not?

PR #2154, opened 2015-06-10
Advanced search was killing my MySQL 

paulhollands:
Advanced search was killing my MySQL due to what looks like a cartesian join on the _search table.

N.B. Only class.search.php and ajax.tickets.php have been changed.

Couldn't unpick the query for this. After indexing all my tables, I gave up and produced a new MySQL view to query instead.

I then added a new find2() function into class.search.php. The SQL for the view is at the top in comments.

I also fixed a bug for date ranged searches in ajax.tickets.php. Start and end dates were being set as the same.

I'm not suggesting my view approach be merged directly, but might inform how you address this issue.

I have reduced query time from 30+ seconds (and a potential MySQL server take down) to around 0.14 seconds.

### DIFF
--- a/ajax.reports.php
+++ b/ajax.reports.php
@@ -1,0 +1,239 @@
+<?php
+/*********************************************************************
+    ajax.reports.php
+
+    AJAX interface for reports -- both plot and tabular data are retrievable
+    in JSON format from this utility. Please put plumbing in /scp/ajax.php
+    pattern rules.
+
+    Jared Hancock <jared@osticket.com>
+    Copyright (c)  2006-2013 osTicket
+    http://www.osticket.com
+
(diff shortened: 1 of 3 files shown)


ntozier:
Thanks for the submission!
(I can't say for sure that the devs will use it, but we always appreciate people contributing.)

paulhollands:
No worries. I detailed the hack in a blog post here. 

http://176.58.112.129/wordpress/blog/2015/06/09/trying-to-fix-osticket-search/

The date range bug needs to be fixed (line 206 of ajax.tickets.php).

balojs:
+1

greezybacon:
Correct me if I'm wrong, but your `find2` function avoids the `_search` table altogether, right?

PR #2173, opened 2015-06-22
Add attachments support to FreeTextField

protich:
Support ability to include attachments to FreeTextField.


protich:
I'm thinking `DynamicField` deletion should trigger `db_cleanup` on a field that has no data. This is necessary so FreeTextField can cleanup attachments.

PR #2175, opened 2015-06-23
Remove packaged third-party CSS and JS

greezybacon:
This adds the ability to the deployment (and packaging) system to automatically fetch third-party javascript and CSS from CDN sources when deploying and packaging osTicket. This makes adding and upgrading unmodified third-party code easier as the code does not need to be checked into the osTicket repo.

It also compresses necessary javascript and CSS files for the staff and client interfaces into a single file (rather than 12 individual files) and minimizes the source files. This dramatically increases page load times, especially for the staff portal, because only 4 files are sourced from the server rather than about 24.

### Outstanding
* [ ] Client CSS is still broken
* [ ] fabric.min.js is packaged into `/js/`
* [ ] When a javascript or CSS file is modified, the deploy script needs to rebuild the entire compressed file it is packaged in. For instance, if `/scp/js/scp.js` is modified, then the compressed javascript file build for the scp will need to be rebuilt and deployed.


greezybacon:
I rebased this one and cleaned it up. It works pretty nicely now

PR #2189, opened 2015-06-26
Time drop down update

jdelhome3578:


ntozier:
Can you please post a screen shot of what this does where?

jdelhome3578:
osticket time update

ntozier:
Great thanks.  Thats what I thought you meant... but for a split second I thought, what if this is actually for a custom time field or something. :)

jdelhome3578:
Another thought along with this would be a user preference setting to set a 12 hour and 24 hour format per user.

greezybacon:
Could we preselect the value greater than but closest to the current time?

jdelhome3578:
I don't see the need, I usually select times at the start, middle or end of my day so that doesn't make sense to me.

PR #2195, opened 2015-06-27
Pass data to the callabale by reference

nimamo:
When using call_user_func_array arguments should be passed by reference in order to allow callable receive them by reference (only defining callable to receive data by reference is not enough).

### DIFF
--- a/include/class.signal.php
+++ b/include/class.signal.php
@@ -95,7 +95,7 @@ class Signal {
                 continue;
             elseif ($check && !call_user_func_array($check, array($object, $data)))
                 continue;
-            call_user_func_array($callable, array($object, $data));
+            call_user_func_array($callable, array($object, &$data));
         }
     }
 }


greezybacon:
See the NOTE on the [php reference page](http://php.net/manual/en/language.references.pass.php). Arguments are only *received* by reference. Since php 5.3, variables should no longer be passed by reference as you suggested.

What is the problem you are solving?

nimamo:
Thanks for the reply.

My comment is misleading, sorry for that. I should not have said `pass by reference` because that's not what I actually did here.

Here is the problem I tried to solve:  
Just to learn how I should handle signals, I wrote a simple listener for `ticket.create.before` signal to swap user submitted message for a predefined message:  

```
Signal::connect('ticket.create.before', function($object,$data){
    // Regardless of user's input, change ticket body to this sample text
    $data['message'] = 'Generated By Listener';

});
```

`ticket.create.before` is sent in [class.ticket.php:2526] (https://github.com/osTicket/osTicket-1.8/blob/develop/include/class.ticket.php#L2526) and as far as I tested, if you do not put a reference to $data in arguments array passed to `call_user_func_array` (called in [class.signal.php:98] (https://github.com/osTicket/osTicket-1.8/blob/develop/include/class.signal.php#L98)) then $vars['message'] value is not changed after this signal is sent (I simply tried `print $vars['message']` before and after signal being sent).  
When calling `call_user_func_array($callable, $args)` a copy of $args is passed to `call_user_func_array`, so when the actual function we intend to call ($callable) is called, it will not have access to the original data, _Therefore, it is **NOT** possible to propogate changes in the signal handlers back to the originating context_ as written in Signal::send() doc-block.  
But if we put a reference to original data in $args array, the problem is solved and changes could be propagated back to original context. And I think this does not violate the what is said in the note you referred to, because we are not using call-time pass-by-reference at all.  
Correct me please if I'm mistaken because being able to propagate changes back to original context is very important to me as I need this for some plugins I intend to create.  
Thanks

greezybacon:
I think your signal handler just need to *receove* by reference,
```php
Signal::connect('ticket.create.before', function($object, &$data) {
    # you can change $data here
});
```

nimamo:
That was a typo. My signal handler actually *does* receive data by reference (edited my comment to correct the typo).  
To make sure how this works, I wrote another simple piece of code not using any code from osTicket:  
```php
<?php
$callable = function(&$data){    // Receive by reference
    $data['message'] = 'Generated By Listener' ;
};
$data = array('message' => 'Original Text');    // Value before calling function
call_user_func_array($callable, array($data));   // Got a warning here
print $data['message'];   // Value stays the same. Output is: Original Text
```

I got this warning (on PHP 5.3.3 CLI):
`PHP warning:  Parameter 1 to {closure}() expected to be a reference, value given on line 7`  
And value of `$data['message']` is not changed.  

But when I call the function like this:
```php
call_user_func_array($callable, array(&$data)); // Put reference to $data in arguments array, no warnings
print $data['message'];  // Works as expected. Output is: Generated By Listener
```
There is no warning and the value of `$data['message']` changes as expected. So it seems it is a must to pass data by reference  
My guess is PHP documentation is missing some notes about passing arguments by reference when using special function *call_user_func_array*.

greezybacon:
Well, it looks like pass by reference is replaced in PHP 5.4 with a fatal error and a crash. However, it turns out that your code does work and it works in PHP 5.4.

nimamo:
Thanks for approving changes. It was my first time editing a file using Github editor and didn't know everything I write here as first comment is going to be used as log message! Sorry for ugly, lengthy log message.

greezybacon:
@nimamo no worries. You've done nothing wrong

PR #2223, opened 2015-07-06
variable: The HTML balancing system urlencodes @href

### LINKED ISSUE #2201: v1.9.9: %{recipient.ticket_link} wrong formatting after saving changes
Hi,
I am trying to create a link to a text by using the variable %{recipient.ticket_link}. System saves the changes but when switching to a different template and then going back to the recent template where the variable was added the system shows the following variable instead: %%7Bticket.staff_link%7D 

The problem is that any variable that I touch gets scrambled like the one above and all the links in the email are unusable.

greezybacon:
The DomDocument::saveHtml() method will properly urlencode() values in the a[@href] attributes when saving. This patch allows for the sanitization while also allowing the variable replacement system to work properly with the sanitized values.

Fixes #2201

### DIFF
--- a/include/class.variable.php
+++ b/include/class.variable.php
@@ -26,7 +26,7 @@ class VariableReplacer {
 
     var $errors;
 
-    function VariableReplacer($start_delim='%{', $end_delim='}') {
+    function VariableReplacer($start_delim='(?:%{|%%7B)', $end_delim='(?:}|%7D)') {
 
         $this->start_delim = $start_delim;
         $this->end_delim = $end_delim;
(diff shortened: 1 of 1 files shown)


include/class.variable.php
```diff
@@ -26,7 +26,7 @@ class VariableReplacer {
 
     var $errors;
 
-    function VariableReplacer($start_delim='%{', $end_delim='}') {
+    function VariableReplacer($start_delim='(?:%{|%%7B)', $end_delim='(?:}|%7D)') {
```
greezybacon: Perhaps the contents of the %{..} should be `url_decode()d` to fix the same issue for contents

balojs:
+1

PR #2241, opened 2015-07-09
cli: Add concept of cli applet to upgrade

greezybacon:
This patch adds the ability to trigger an upgrade from the command line

```sh
php manage.php upgrade
```

And can hopefully help with recovering failed upgrades in the future.


ntozier:
Great idea!

PR #2303, opened 2015-07-25
Pass Thru Auth and Destination

jdelhome3578:
Possible fix for Pass Thru Auth and Destination
https://github.com/osTicket/core-plugins/issues/76


jdelhome3578:
merged with #2916

PR #2311, opened 2015-07-27
tasks: Ticket tasks access

protich:
Restrict view access depending on agent's perm.

### DIFF
--- a/include/staff/ticket-tasks.inc.php
+++ b/include/staff/ticket-tasks.inc.php
@@ -73,6 +73,7 @@ if ($count) { ?>
     <?php
     foreach($tasks as $task) {
         $id = $task->getId();
+        $access = $task->checkStaffPerm($thisstaff);
         $assigned='';
         if ($task->staff)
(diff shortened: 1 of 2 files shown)


include/staff/ticket-tasks.inc.php
```diff
@@ -84,8 +85,10 @@ class="green button action-button ticket-task-action"
         $threadcount = $task->getThread() ?
             $task->getThread()->getNumEntries() : 0;
 
-        $viewhref = sprintf('#tickets/%d/tasks/%d/view',
-                $ticket->getId(), $id);
+        if ($access)
+            $viewhref = sprintf('#tickets/%d/tasks/%d/view', $ticket->getId(), $id);
+        else
+            $viewhref = '#';
```
greezybacon: It looks like the `else` case isn't used, right?

protich: It is -- just setting the href to "#"

greezybacon: But it looks like `$viewhref` is not used in the following code (if `$access` is false)

PR #2312, opened 2015-07-27
Fix auth-ldap Full Name: [object Object]

jdelhome3578:
This fix worked for me. -  https://github.com/osTicket/core-plugins/issues/71


greezybacon:
Can you try the patch in #2335, I think it's a more direct fix

jdelhome3578:
#2335 This fix work here. :+1: 
image

jdelhome3578:
#2335 Is a more direct fix.

PR #2314, opened 2015-07-27
list: Fix add and edit for "Ticket Status" list

### LINKED ISSUE #2292: 1.10 RC1 Issues - Custom List Ticket Status
Unable to create new Status. The ajax window comes up but after saving the new status does not display anywhere in the system. If I go back to create the same one the system says that is already exists.

image
image

greezybacon:
This patch shores up the new list management feature with the custom list handler feature written to support lists being saved in a table other than %list_item, namely the built-in ticket status list.

Fixes #2292

### DIFF
--- a/include/ajax.forms.php
+++ b/include/ajax.forms.php
@@ -99,35 +99,18 @@ class DynamicFormsAjaxAPI extends AjaxController {
     }
 
 
-    function _getListItemEditForm($source=null, $item=false) {
-        return new SimpleForm(array(
-            'value' => new TextboxField(array(
-                'required' => true,
-                'label' => __('Value'),
-                'configuration' => array(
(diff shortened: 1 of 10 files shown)


jdelhome3578:
Does not correct the issue, New Status still being saved in `list_items` instead of `ticket_status`

jdelhome3578:
I am know able to open current Statuses but when creating a new status the `value` and settings do not save correctly.

greezybacon:
@jdelhome3578 thanks for testing. I tried harder and submitted some updated code. Can you test again?

jdelhome3578:
That look better!! Confirmed Fix!

protich:
Reviewed-By: Peter Rotich <peter@enhancesoft.com>

PR #2325, opened 2015-07-29
lock: Implement lazy locking system

greezybacon:
This proposes a new, more flexible locking system, which waits to acquire a lock until the user starts entering information into the system, or performing an action which immediately requires a lock. Under this scheme, simply viewing a lockable object, such as a ticket, does not require a lock; however, starting to type a response to the end user will trigger a lock acquisition. If a lock cannot be acquired, the user is notified with an annoying warning banner.

### Outstanding
* [x] Fix double AJAX posting

##### New slide out "message bar" with error instead of the annoying popup
screen shot 2015-07-29 at 4 00 34 pm


jdelhome3578:
:+1:

Chefkeks:
That's a really neat improvement :+1:

jdelhome3578:
As soon as I start typing a response the `Your lock is expiring soon.` message drops down. I click on `renew` and it just spins. Is this the intended operation?

greezybacon:
> As soon as I start typing a response the Your lock is expiring soon. message drops down. I click on renew and it just spins. Is this the intended operation?

Maybe you mis-patched something? Are there any notable javascript logs?

jdelhome3578:
I retested from the tip and I am still having issues #2343

PR #2334, opened 2015-07-30
avatar: Make avatars more of a utility (with settings)

greezybacon:
Initially, include support only for Gravatar. This will also allow for more flexible avatar sources, including LDAP and Active Directory.

### Outstanding
* [x] Fix avatar sizing on the client portal


include/class.avatar.php
```diff
@@ -0,0 +1,134 @@
...
+**********************************************************************/
+
+abstract class Avatar {
+    var $user;
+
+    function __construct($user) {
+        $this->user = $user;
+    }
+
+    abstract function getUrl($size);
```
greezybacon: The interface should add `__toString()` since the system coerces them to strings

Chefkeks:
Seems like Avatars are growing up :+1: :)

jdelhome3578:
Ticket Thread Client View Error

image

include/class.avatar.php
```diff
@@ -0,0 +1,238 @@
...
+
+    function __construct($mode) {
+        $this->mode = $mode;
+    }
+
+    function makeAvatar($uid) {
+        if (!extension_loaded('gd'))
+            Http::redirect(ROOT_PATH.'images/mystery-oscar.png');
+
+        $sprite = self::$sprites[$this->mode];
```
greezybacon: Show mystery oscar for unknown mode

jdelhome3578:
Ticket Thread Client View after message post.

image

PR #2335, opened 2015-07-30
search: Fix malformed results on remote user search

greezybacon:


jdelhome3578:
Make sure this is in osTicket:develop-next

PR #2357, opened 2015-08-03
topic: Sort localized list case insensitively

### LINKED ISSUE #2350: Help Topic List Sorting
When sorting Alphabetically the system does not sort upper case and lower case together.

example: The system will sort A-Z then a-z
Suggestion: Sort A-Z and a-z together not separate.

greezybacon:
Fixes #2350

### DIFF
--- a/include/class.i18n.php
+++ b/include/class.i18n.php
@@ -461,6 +461,28 @@ class Internationalization {
         return $locales;
     }
 
+    static function sortKeyedList($list, $case=false) {
+        global $cfg;
+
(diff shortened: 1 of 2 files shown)


include/class.i18n.php
```diff
@@ -461,6 +461,28 @@ static function allLocales() {
         return $locales;
     }
 
+    static function sortKeyedList($list, $case=false) {
+        global $cfg;
+
+        if (false && $cfg && function_exists('collator_create')) {
```
greezybacon: oops

PR #2368, opened 2015-08-04
autoload: Add concept of autoloading include/ classes

greezybacon:
Initially, this doesn't seem to have any affect on page load time whatsoever. However, systems without APC will likely benefit from the concept.

We should likely check the autoload index (`__autoload.php` in this patch) into Git so that folks that just download or `git checkout` from GitHub will have access to the autoload index. Otherwise, they would need to run `php manage.php package --autoload`


bootstrap.php
```diff
@@ -189,23 +189,15 @@ function connect() {
...
-        require_once INCLUDE_DIR.'class.search.php';
+        $classes = (include '__autoload.php');
+        spl_autoload_register(function($class) use ($classes) {
+            $class = strtoupper($class);
+            if (!isset($classes[$class]))
+                return;
+            require ROOT_DIR . $classes[$class];
+        });
+        require INCLUDE_DIR . 'mysqli.php';
+        require INCLUDE_DIR . 'smörgåsbord.php';
```
greezybacon: oops

include/cli/modules/package.php
```diff
@@ -103,6 +109,42 @@ function print_dns() {
...
+                    }
+                }
+            }
+        };
+        $classes = array();
+        $findClasses($path, $classes);
+
+        $classes = var_export($classes, true);
+        file_put_contents(INCLUDE_DIR . '/.autoload.php',
+            '<?php return '.$classes.';');
```
greezybacon: Perhaps we should add a AUTO-GENERATED warning for readers of the file

Chefkeks:
> Perhaps we should add a AUTO-GENERATED warning for readers of the file
Yes ;)

greezybacon:
Actually, after more thought, I would like to omit the `.autoload.php` file from the repo, because otherwise, regenerating it would be required in the process of every change request involving adding, removing, or renaming classes. I added some notes to the `README.md` file for folks running form the repo directly.

greezybacon:
I rebased this one on v1.10 if it is still of interest

PR #2377, opened 2015-08-06
Issue/ui consistency

nfebuary:


greezybacon:
This won't work. You suggested merging to `develop` and you need to suggest merging to `develop-next`

PR #2391, opened 2015-08-11
Very minor fixes spell mismatches - fast to merge - (even in app name osTicet :)

pauluz:


include/cli/modules/unpack.php
```diff
@@ -139,7 +139,7 @@ function copyFile($src, $dest, $hash=false, $mode=0644) {
     }
 
     /**
-     * Copy from source to desination, perhaps recursing up to n folders.
+     * Copy from source to destination, perhaps recurring up to n folders.
```
greezybacon: *recursing* is correct

PR #2393, opened 2015-08-11
staff: Allow numeric usernames

### LINKED ISSUE #2378: V 1.9.11/ RC 1.10 r1 ... no login with number as username
hey

i make a test with a number as username, he say access denied.

is it normal?

greezybacon:
Fixes #2378

### DIFF
--- a/include/ajax.content.php
+++ b/include/ajax.content.php
@@ -121,7 +121,7 @@ class ContentAjaxAPI extends AjaxController {
         case 'none':
             break;
         case 'agent':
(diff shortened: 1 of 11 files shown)


greezybacon:
I think we should create `Staff::lookupByLogin` which would find Staff members by the username or email address and then migrate `Staff::lookup` to work on the `id` only.

matze-biss:
Does not work anymore in the newest version. Is there a solution?

mckaygerhard:
@matze-biss what do you mean with does not work more?

PR #2396, opened 2015-08-11
search: Fix malformed results on remote user search

jdelhome3578:
same as #2335 but for develop-next


greezybacon:
Can't we just merge `develop` into `develop-next`?

jdelhome3578:
Sure, I just noticed that it wasn't in the develp-next code yet.

PR #2400, opened 2015-08-11
Queue sort template

protich:
Utilize the same template for tickets and tasks queues


greezybacon:
Looks like `=undefined` will show up as a query argument if the query string is initially empty. I would suggest filtering it out:
```patch
diff --git a/scp/js/scp.js b/scp/js/scp.js
index 7d2bc01..76638fd 100644
--- a/scp/js/scp.js
+++ b/scp/js/scp.js
@@ -1177,7 +1177,8 @@ function addSearchParam(data) {
     var i=kvp.length, x, params = {};
     while (i--) {
         x = kvp[i].split('=');
-        params[decodeURIComponent(x[0])] = decodeURIComponent(x[1]);
+        if (!!x[1])
+            params[decodeURIComponent(x[0])] = decodeURIComponent(x[1]);
     }
 
     //this will reload the page, it's likely better to store this until finished
```

PR #2401, opened 2015-08-11
boostrap-typeahead.js / typeahead.css should NOT be loaded from scp/css and scp/js

ghost:
Since many people (like me) would like to have http auth enabled on the /scp/ directory, I see no point in including css and js files from the scp/css and scp/js directory on the client frontend.

I have created the two new files  js/boostrap-typeahead.js  and css/typeahead.css  and changed the include/client/header.inc accordingly.

Thanks

Referenced issue: #2402

### DIFF
--- a/css/typeahead.css
+++ b/css/typeahead.css
@@ -1,0 +1,58 @@
+/* FROM bootstrap.css - just what is needed for typeahead */
+.dropdown-menu {
+  position: absolute;
+  top: 100%;
+  left: 0;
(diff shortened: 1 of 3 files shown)


greezybacon:
I think we should move the files from the `scp/` folder so that there are not two instances of the files in the code base

JediKev:
This has been rewritten here:
- https://github.com/osTicket/osTicket/pull/6541

Issue #2399, opened 2015-08-11
1.10RC1 Filter Actions Team Bug


kest874:
When you user the "Assign Team" filter action it's not settign the correct team_id in the database:

![image](https://cloud.githubusercontent.com/assets/12860950/9209209/edda2dd0-403a-11e5-9964-424dd528d014.png)

ost_filter:

![image](https://cloud.githubusercontent.com/assets/12860950/9209225/ffa5e9b4-403a-11e5-9380-45a6ae9daedc.png)

ost_team:

![image](https://cloud.githubusercontent.com/assets/12860950/9209233/0b7c3e96-403b-11e5-8853-0b7a646971a1.png)

ost_filter_action:

![image](https://cloud.githubusercontent.com/assets/12860950/9209267/2e69d116-403b-11e5-86f2-9bb9eb6290e9.png)

kest874:
This: https://github.com/kest874/osTicket-1.8/commit/c509029187d6b0a36ef10ab22c51e4573abbb1ca 
resolves the issue.

PR #2406, opened 2015-08-12
Resolves issue #2399

kest874:


greezybacon:
Thanks for the patch. I fixed it for both teams and departments in 84c84d44afd4a4335f09236fb7293d5a61962b14

PR #2411, opened 2015-08-13
bug: Zombie tickets

protich:
Department create now considers the possibility of preset id (primary key)
Default help topic routing to floating system's default department

PS: The bug only affects users using DB cluster.

### DIFF
--- a/include/class.dept.php
+++ b/include/class.dept.php
@@ -530,7 +530,7 @@ implements TemplateVariable {
     }
 
     static function __create($vars, &$errors) {
-        $dept = self::create();
+        $dept = self::create($vars);
         $dept->update($vars, $errors);
 
         return isset($dept->id) ? $dept : null;
--- a/include/class.i18n.php
+++ b/include/class.i18n.php
@@ -64,8 +64,6 @@ class Internationalization {
             'ticket_status.yaml' => 'TicketStatus::__create',
             // Role
             'role.yaml' =>          'Role::__create',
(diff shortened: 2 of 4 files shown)


greezybacon:
Looks like dropping the `dept_id` requires some attention in `Topic::__create`:
```patch
diff --git a/include/class.topic.php b/include/class.topic.php
index 9571bf7..140be92 100644
--- a/include/class.topic.php
+++ b/include/class.topic.php
@@ -282,6 +282,8 @@ implements TemplateVariable {
 
     static function __create($vars, &$errors) {
         $topic = self::create($vars);
+        if (!isset($vars['dept_id']))
+            $vars['dept_id'] = 0;
         $topic->update($vars, $errors);
         return $topic;
     }
```

greezybacon:
And we should skip installing initial "groups":
```patch
diff --git a/include/class.i18n.php b/include/class.i18n.php
index 3c0da2b..9163f6c 100644
--- a/include/class.i18n.php
+++ b/include/class.i18n.php
@@ -64,8 +64,6 @@ class Internationalization {
             'ticket_status.yaml' => 'TicketStatus::__create',
             // Role
             'role.yaml' =>          'Role::__create',
-            // Note that group requires department
-            'group.yaml' =>         'Group::__create',
             'file.yaml' =>          'AttachmentFile::__create',
             'sequence.yaml' =>      'Sequence::__create',
         );
```

Issue #2458, opened 2015-08-23
develop-next - latest - Firefox Drop down view


gmt-it:
![dropdown_001](https://cloud.githubusercontent.com/assets/10116166/9427184/c664d434-496e-11e5-8351-3a903f957b93.png)

greezybacon:
I think @nfebuary addressed this in #2476

PR #2476, opened 2015-08-26
Firefox form style fix

nfebuary:
Issue with select list options pushing below the viewable select box in firefox. giving the appearance of being cut in half.

Before:
screen shot 2015-08-26 at 2 47 13 pm

After:
screen shot 2015-08-26 at 2 49 04 pm

And rounded corners on inputs for firefox as an added bonus.
screen shot 2015-08-26 at 2 51 42 pm


greezybacon:
Fixes #2458

PR #2477, opened 2015-08-26
forms: Fix search by datetime fields

greezybacon:
This changes the storage model for datetime fields and stores them in the same format as the MySQL `datetime` column type. This allows for the searching code to be consistent when searching the database for datetime custom form entry data.

It also addresses and inconsistency in the forms code, where the PHP value of form fields was stored in the %cdata tables—instead of the normal database version. This code includes a FormField::getKeys() method to deal with multi-entry choice fields.

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -884,6 +884,10 @@ class FormField {
         return Format::searchable($this->toString($value));
     }
 
+    function getKeys($value) {
+        return $this->to_database($value);
+    }
+
     /**
      * Fetches a list of options for searching. The values returned from
      * this method are passed to the widget's `::render()` method so that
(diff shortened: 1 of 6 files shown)


include/class.forms.php
```diff
@@ -1760,20 +1771,14 @@ function toString($value) {
...
     }
 
-    function export($value) {
-        $config = $this->getConfiguration();
-        if (!$value)
-            return '';
-        else
-            return Format::date($value, false, 'y-MM-dd HH:mm:ss', !$config['gmt'] ? 'UTC' : false);
-    }
-
```
greezybacon: This block may be necessary actaully — but should probably use a standard date format rather than the locale format

protich: Any update on above note? 

Also, what happens to installations upgraded past the database patch in question?

greezybacon: I'm thinking the CSV export should use a more computer-readabale date format (like the one in the red block). Otherwise, the locale-based format from `Format::datetime()` would be used, and spreadsheet programs might not parse the date well. Is there a standard on this?

Unpatched databases won't benefit from the new code. However, the new code will break stuff too. One line could be changed in `class.misc.php`, around line 96 in `dbtime`
```php
            $time = is_int($var) ? $var : strtotime($var) ?: $var;
```
which would allow defaulting back to `$var` (assuming its a string Unix timestamp) if `$var` is neither a PHP integer nor a valid date format. This would allow unpatched databases to keep the existing data which recording new and updated data in the new format. New and update data would also be searchable. Perhaps we could cook something up for the command-line for manual patching

PR #2491, opened 2015-08-28
reduce memory usage when storing file in DB

ericLemanissier:
When saving an attached file in the database, Passing the whole content of the file to bin2hex leads to allocating the memory for the whole hex encoded file while the binary content of the file is still in memory. This leads easily to Out of memory error on limited resources servers. Calling bin2hex on each chunk of the file greatly limits the memory usage.


greezybacon:
This looks nice. Thanks for making the request

PR #2523, opened 2015-09-03
orm: Prefer model relationship over local variable.

protich:
The change helps avoid local variables from overwriting joins.

### DIFF
--- a/include/class.export.php
+++ b/include/class.export.php
@@ -59,7 +59,7 @@ class Export {
         // Reset the $sql query
         $tickets = $sql->models()
             ->select_related('user', 'user__default_email', 'dept', 'staff',
-                'team', 'staff', 'cdata', 'topic', 'status', 'cdata.priority')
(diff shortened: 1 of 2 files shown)


protich:
This is a bad idea -- at least based on my initial test.  My original rationale was to avoid local variables from clobbering relationships... good example is ticket's ``cdata.priority``

include/class.export.php
```diff
@@ -59,7 +59,7 @@ static function dumpTickets($sql, $how='csv') {
         // Reset the $sql query
         $tickets = $sql->models()
             ->select_related('user', 'user__default_email', 'dept', 'staff',
-                'team', 'staff', 'cdata', 'topic', 'status', 'cdata.priority')
+                'team', 'staff', 'cdata', 'topic', 'status', 'cdata__priority')
```
greezybacon: I think this part is correct

PR #2529, opened 2015-09-06
bug: Collaborators Notice

protich:
Fix case where collaborators were not being notified 
Alert admin last on new ticket - just in case admin email address is also used by an alerted agent.


greezybacon:
What's up with the `$charset` hinting in mail parse? Isn't UTF-8 the default?

protich:
``parseAddressList`` gets called statically in ``getAllEmailRecipients`` -- meaning ``$this`` won't be in valid context.

greezybacon:
That makes sense. Looks good

greezybacon:
This was merged with #2549

PR #2549, opened 2015-09-11
Fix posting-user determination in Thread::postEmail

greezybacon:
This patch changes the way the thread entry type and other properties are determined by using the `userClass` determined in the `Mailer::decodeMessageId` method which is preferred over the older heuristic methodology using the `From` email address alone.

It also includes the work of and builds off of Peter's pull request in #2529


greezybacon:
I adjusted the patch as you suggested

PR #2552, opened 2015-09-14
Blank page from 'register' link in notification email to guest user.

bailey86:
We have existing clients who do not have an account (yet) on osTicket.  We create a ticket for them from the admin interface (creating a new guest user account) and this results in an email being sent to them with details of a job carried out.  This user can then click on the link in the notification email and view the ticket as guest - and hopefully opt to register an account.

The problem seems to be this - the user clicks on the ticket number in their notification email which lets them view the ticket as a guest - there is then a link to 'register for an account' linked to http://www.example.com/account.php?do=create - but clicking this link results in a blank page.  The PHP error is:

PHP Fatal error:  Call to a member function getField() on a non-object in /home/osticket/public_html/account.php on line 35, referer: http://www.example.com/tickets.php?id=105

This discussion pointed me to the possible bug.

http://osticket.com/forum/discussion/86076/blank-user-account-registration-page-after-viewing-ticket-first-using-auth-link

This is the code change I'm proposing - I've added curly braces to the foreach loop as well:

    // Guest registering for an account
    if ($thisclient->isGuest()) {
        foreach ($thisclient->getForms() as $f)
            if ($f->get('type') == 'U')
                $user_form = $f;
        $user_form->getField('email')->configure('disabled', true);
    }

Should be changed to:

    // Guest registering for an account
    if ($thisclient->isGuest()) {
        foreach ($thisclient->getForms() as $f) {
            if ($f->get('type') == 'U') {
                $user_form = $f;
                $user_form->getField('email')->configure('disabled', true);
            }
        }
    }

### DIFF
--- a/account.php
+++ b/account.php
@@ -29,10 +29,12 @@ if (!$cfg || !$cfg->isClientRegistrationEnabled()) {
 elseif ($thisclient) {
     // Guest registering for an account
     if ($thisclient->isGuest()) {
-        foreach ($thisclient->getForms() as $f)
-            if ($f->get('type') == 'U')
+        foreach ($thisclient->getForms() as $f) {
+            if ($f->get('type') == 'U') {
                 $user_form = $f;
-        $user_form->getField('email')->configure('disabled', true);
+                $user_form->getField('email')->configure('disabled', true);
+            }
+        }    
     }
     // Existing client (with an account) updating profile
     else {


bailey86:
This change does get the link to work - and for a registration form to appear.

My feeling is that the form which appears should have the email address of the guest account filled in - but this field should not be disabled as this may cause issues when possibly accessing the form again.  If the user sees their email address filled in then that should be clear enough - if they decide to change the email address then that is their decision.

greezybacon:
Good catch. I think you've got the right fix

PR #2559, opened 2015-09-16
Add preview option of threads in a ticket.

nfebuary:
Thread preview tab in the ticket preview popout.

### DIFF
--- a/include/staff/templates/ticket-preview.tmpl.php
+++ b/include/staff/templates/ticket-preview.tmpl.php
@@ -41,6 +41,10 @@ echo sprintf('
             faded"></i>&nbsp;'.__('Collaborators (%d)').'</a></li>',
             $ticket->getThread()->getNumCollaborators());
 }
+echo '<li><a id="thread_tab" href="#threadPreview"
+            ><i class="icon-fixed-width icon-list
+            faded"></i>&nbsp;'.__('Thread Preview (6)').'</a></li>';
+
 echo '</ul>';
 echo '<div id="ticket-preview_container">';
 echo '<div class="tab_content" id="preview">';
@@ -117,6 +121,7 @@ echo '
     </table>';
 echo '</div>'; // ticket preview content.
 ?>
+
 <div class="hidden tab_content" id="collab">
     <table border="0" cellspacing="" cellpadding="1">
         <colgroup><col style="min-width: 250px;"></col></colgroup>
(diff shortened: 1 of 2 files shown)


nfebuary:
New thread preview update removes avatars and changes design layout and function. Now headers are clickable to expand and collapse instead of individual buttons. Color changes to keep the conversation from being too busy visually. Collapsed colors mimic the full thread view email template.

Collapsed:
screen shot 2015-10-13 at 4 51 41 pm

Expanded:
screen shot 2015-10-13 at 4 51 50 pm

Chefkeks:
Looks pretty good :+1: 

Have to admit that I never(!) been a fan of the ticket preview since about half the info you got (on the `Ticket Summary` tab) you already get when looking at the queue (e.g. Create Date, Assignee, etc.) and it never made any sense to look a the preview. But now with this thread preview I guess me (and also my colleagues here at work) will definitely use this feature more often.

Since custom queues are on the way (https://github.com/osTicket/osTicket-1.8/pull/2577) and peeps will then customize the queue to their needs with the info they really need my suggestion for the preview is to make the `thread preview` the default tab. I think it's more common to quickly preview the last messages / responses / notes of a ticket instead of showing more info about the ticket itself (like Create Date, Assignee, etc.).

That said I think it makes sense to also allow/invent a custom ticket header layout option in the future. With "ticket header" I mean that here:
image

With 1.10 you plan to replace the phone field with the organization field. Since we here need the phone field, but don't need the organization field at all I think it's a bad decision. But instead of continuing to complain ;) my suggestion is to allow custom ticket header layouts, so necessary info is shown and other is then hidden. E.g. we would only need the following fields - all unneeded fields are blacked out:
image

So as you can see there is a huge potential to reduce unnecessary/unneeded (but always shown) info and therefore some other needed info (e.g. phone number in our case) is missing. Since I think we're not the only ones who'd love to see a customizable ticket header layout I guess it shall be on the list for future improvements (right behind custom ticket queues and the new/improved ticket preview).

How do you @nfebuary, @protich, @greezybacon think about this?

kest874:
I think it would be nice if the "Thread Preview" showed up while hovering over the subject field

nfebuary:
@Chefkeks @kest874 Agreed on all points. I would like to have something setup like a design system for admins to literally pick what they want for ticket views with maybe a global template that could be overridden with a Help Topic template. It could be either a drag and drop the information you want UI, or use the existing variable system. You would be able to do this with tasks, tickets, knowledge base and Agent, Department profiles. At least in my head this is what I would like to see.

Chefkeks:
@nfebuary
Ambitious goal ;)
I thought that much smaller in my head, but like the idea of such a design system :+1:

balojs:
+1

PR #2563, opened 2015-09-17
Export related fixes

protich:
* Add CDATA support to user and org models.
* Fix export bugs
  - [x] Fix priority on ticket export
  - [x] Fix users export
  - [x] Fix orgs export

## Pending
- [ ] include cdata when searching orgs & users


greezybacon:
You really think we need persistent CDATA tables for users and orgs?

protich:
Perhaps __NOT__, but orm related implementation of objects form views is a little funny at the moment - especially with dealing with the exporter where queryset is expected vs. naked query.

In the interest of pushing the fixes to production, I implanted persistent CDATA.

Issue #2024, opened 2015-05-07
In user interface if register an account (in the up message) is pressed then is loaded a blank page


PeopleInside:
![immagine senza nome](https://cloud.githubusercontent.com/assets/5006150/7520619/3e3037c8-f4e8-11e4-8d65-4ef448ff8d17.png)

If user press register

![immagine senza nome](https://cloud.githubusercontent.com/assets/5006150/7520669/a0b3dae4-f4e8-11e4-9bcd-20942f82b932.png)

ntozier:
That link works in my installation.
![capture](https://cloud.githubusercontent.com/assets/1315768/7520766/21d7c6c6-f4b7-11e4-974a-c50c4240fe3b.PNG)

PeopleInside:
I tried with a fresh install on my server and didn't work so if something
not supported by the server should be a check during install process..

I don't know why is not working on my server but tested with fresh install
of OsTicket, English bit also with Italian language file.

Is required something of particular for make it work?

Inviato da dispositivo mobile. | Sended  by mobile device
Il 07/Mag/2015 18:46, "ntozier" <notifications@github.com> ha scritto:

PeopleInside:
Interesting, from my phone if I put the direct link in browser work.

Did you tried from user interface on open ticket where user is guest or you tried to put direct URL into browser?

ntozier:
I tried direct url, and I used the link on my open page since I don't have that link on my landing page.
http://support.harborhomes.org/open.php

PeopleInside:
Ok so sorry but your test is not valid for diagnosis. Direct link works
also for me.

How to reproduce:
1. Create a ticket from not logged and not registered client side
2. If you follow the link on the email you will se the ticket and in top
you have create an account link. If you click that as showed in the image
is loaded a blank page.

Tested now from my smartphone and I can reproduce also from there

Inviato da dispositivo mobile. | Sended  by mobile device
Il 07/Mag/2015 19:07, "ntozier" <notifications@github.com> ha scritto:

ntozier:
"Create a ticket from not logged and not registered client side"

I cannot possibly do that.  My installations ALL require the person to be logged in to create a ticket.  Clicking on Open a ticket takes me to a login screen with the link to register (create an account) on it.

What version are you running?

PeopleInside:
Last version.
Your is only a settings.

I can also force user to register to see ticket but I want alow also not
registered user.

[...]

PR #2572, opened 2015-09-19
try to fix issue #2024

matlinuxer2:
I found the first entry of "$thisclient->getForms()" is a "DynamicFormEntry",
 which has property "object_type" instead of "type".
So I change it to "object_type", and all go well.

### DIFF
--- a/account.php
+++ b/account.php
@@ -30,7 +30,7 @@ elseif ($thisclient) {
     // Guest registering for an account
     if ($thisclient->isGuest()) {
         foreach ($thisclient->getForms() as $f)
-            if ($f->get('type') == 'U')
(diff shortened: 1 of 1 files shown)


greezybacon:
Nice catch! I think you're correct. 

In the ORM since we wrote this code, we've started using `$object->type` instead of `$object->get('type')`, which will throw an exception if the field doesn't exist in the database - in hopes of catching bugs like this one.

greezybacon:
Thanks for the patch. I manually applied it to the `develop` branch directly (55338e5).

matlinuxer2:
GJ! Thank you very much!

Issue #915, opened 2014-05-19
TLS/SSL failure for thor: SSL negotiation failed


snizzleorg:
Hi,

(osTicket v1.9.0)

my IMAP server s using a self-signed certificate and I cannot connect to it using osTicket. 

`TLS/SSL failure for (mailserver): SSL negotiation failed`

this even with 

```
//Mail server string
            $this->srvstr=sprintf('{%s:%d/%s', $this->getHost(), $this->getPort(), $this->getProtocol());
            if(!strcasecmp($this->getEncryption(), 'SSL'))
                $this->srvstr.='/ssl';

            $this->srvstr.='/novalidate-cert}';
```

in class.mailfetch.php

PR #2577, opened 2015-09-21
Feature: Customizable ticket queues, customizable columns

greezybacon:
### How-To Test

Checkout this branch (`greezybacon/feature/custom-queues`) or [visit my page](https://github.com/greezybacon/osTicket-1.8/tree/feature/custom-queues) and download the ZIP file. 

Visit the setup page and perform a normal installation. You can also attempt an upgrade from any version v1.6ST and forward. Happy testing.

##### Ticket queues are defined in Settings -> Tickets page, "Queues" tab

screen shot 2015-10-07 at 10 22 21 pm

#### Queues inherit the criteria system from the Advanced Search feature, but have the addition of a "Name" and "Description" (todo)

Queues also have the ability to inherit criteria from a parent queue. Therefore, queues defined under "Open" would not need to include criteria to define "Open" tickets again and again

screen shot 2016-06-05 at 2 01 40 pm

#### Queues have designable columns

Any data associated with just about anything can be listed in the ticket queue page. Primary and secondary data sources are selectable. If no data exists for the primary data field, then the data from the secondary is displayed in the queue listing. Columns can be dragged to arrange their order.

screen shot 2016-06-05 at 1 42 59 pm

#### Columns can have _annotations_

**Annotations** are extra flair, such as the attachment count, collaborator count, overdue flag, etc., which can be added to the start or end of the cell (floating), or before or after the text (inline).

screen shot 2016-06-05 at 1 54 58 pm

#### Columns can have _sophisticated_ conditions

**Conditions** are checks run on the data associated with the ticket. For records that match the conditions, extra formatting can be defined. For instance, unanswered tickets can have some cells shown in bold. Any data (custom or otherwise) affiliated with the ticket, user, and organization, can be used in conditions. Even set crazy things like `background-color`

screen shot 2016-06-05 at 1 44 45 pm

#### Preview before saving

All settings for the criteria, column sources, annotations, conditions, etc., are viewable without saving. Click the "Preview" tab and the queue is rendered dynamically.

screen shot 2015-10-07 at 10 24 33 pm

#### Queue pages have hierarchical navigation

Queues are selectable from drop-down navigation in the ticket listing pages. Additionally, edit or delete the queue straight from the navigation.

screen shot 2016-06-05 at 1 42 35 pm

#### Use "Quick Filter" to reduce ticket listing in real-time

Choose field such as Assignee, Department, Help Topic, some custom data field, etc., which can be selected when the queue is viewed. The queue will be filtered to only show tickets matching the quick filter.

screen shot 2015-10-07 at 10 25 35 pm

#### Columns are customizable for (saved) advanced searches too

screen shot 2016-06-05 at 1 48 40 pm

#### Quick access to edit queues, as well as add and remove custom personal sub-queues

screen shot 2016-06-05 at 1 42 45 pm

#### Queue column headings are translatable

screen shot 2016-06-05 at 1 45 37 pm

### Outstanding
- [x] Add screenshots
- [x] Add how-to guide for testing
- [x] Make column headings translatable
- [x] Drop interactive column resizing
- [x] Add core settings for default queue (previously "Open")
- [ ] Add container queues with no criteria and default sub-queues. This will replace the current "Show assigned tickets on open queue" setting and friends. When the container queue is clicked, the default sub-queue will be shown instead
- [x] Implement quick filter
- [x] Implement sorting options
- [x] Add install yaml for queues previous to v1.11
- [x] Implement column delete from queue
- [x] Enable/disable queues
- [ ] Add description (queue page title) field
- [x] Delete queue
- [x] Select column configuration for rendering saved search
- [x] Publish (make public) feature for saved searches
- [x] Add personal queues
- [x] Add queue counts or drop the `(?)` tokens, rectify staff stats functions
- [x] Add navigation for saved searches
- [x] Add queue as a child (via navigation)
- [x] Conditions do not load properly when editing a queue
- [x] Implement sorting options, sortable headers
- [x] Implement staff visibility constraints
- [x] Integrate advanced search and saved searches
- [x] `path` is not set correct at installation
- [x] Integrate user and org searches from ticket view
- [x] Transition advanced search criteria to the new object-base `getSearchableFields` used for queues
- [x] Add recent search history to search dropdown
- [x] Queue management does not work in Firefox
- [x] Show parent queue criteria, grayed out (disabled) for personal queues
- [x] Add custom columns to saved searches
- [ ] Add icon selection to custom queue management page
- [x] Add date formatting options in somewhere (maybe in the filter?)
- [ ] Retire CDATA usage and feature
- [x] Add advanced sorting options
- [x] Add default advanced sort to queue configuration
- [ ] Delete custom column (definition)
- [ ] Delete custom sort criteria (definition)
- [ ] Personal queue columns don't need to be translatable
- [x] Add conditions to the queue itself, allowing for controlling the display of an entire row
- [ ] Make queue titles translatable


include/class.orm.php
```diff
@@ -1553,6 +1580,82 @@ function prime() {
...
+     * This would be used in favor of ::filter() which might trigger another
+     * database query. See ::findAll() for more details.
+     *
+     * Example:
+     * >>> $a = new User();
+     * >>> $a->roles->add(Role::lookup(['name' => 'administator']));
+     * >>> $a->roles->findFirst(['roles__name__startswith' => 'admin']);
+     * <Role: administrator>
+     */
+    function findFirst(array $criteria) {
```
greezybacon: This should be dropped (ie. moved) from the InstrumentedList class

Chefkeks:
> Add screenshots

Definitely! ;)

jdelhome3578:
:+1: :100:  I would love to see this in 1.10

balojs:
+1

greezybacon:
Aight peeps. I've added some screenshots. I hope to near an alpha state tomorrow and will add some notes on how to test the feature.

Cheers

AlexYoung28:
Looks really good!!

Aka4:
Amazing :) You plan to release this feature in 1.10 RC3?

Chefkeks:
Looks pretty good and I know it's far from being finished, but anyway some minor suggestion from my side. Did some quick editw in MS Paint to show you what I mean - not pretty but a picture often says more than a thousand words:

I'd remove the sub menu to edit/delete a queue and display the button directly inline.

image




Additionally I'd really love to see an option to display all or several (select-able) queue in one page:

image




The queue page itself should then look like the following if several/all queues are selected:

image

greezybacon:
> Additionally I'd really love to see an option to display all or several (select-able) queues in one page:

Could you expand on this idea? Bear in mind that the feature will allow you to create as many queues as you want and specify the criteria for each.

If showing more than one, would the results be a union or an intersection (unique results from both queues or results common in both queues respectively). How would the column display differences be rectified between multiple queues?

greezybacon:
@Aka4 sorry, this is a major feature for the next major version

Chefkeks:
> Could you expand on this idea? Bear in mind that the feature will allow you to create as many queues as you want and specify the criteria for each.

> If showing more than one, would the results be a union or an intersection (unique results from both queues or results common in both queues respectively). How would the column display differences be rectified between multiple queues?

Hehe, I get your point and see what the issue it would cause... Have to admit that I didn't see this issue coming.

Basically I'd tend to show a ticket that for example is unassigned + unanswered in both queues (unassigned queue and unanswered queue) since it matches the corresponding criteria of each queue. Reason is that I personally think showing multiple queues should not affect the results of a queue - so just display them together. Even though some tickets will then be displayed in more than 1 queue.

greezybacon:
Yeah, I think I see your point: The criteria system doesn't allow for ` OR ` based criteria

greezybacon:
As a workaround, you could flip the logic ((A or B) == !(A and B)). Therefore, one could create another queue (named "Unassigned + Unanswered") with the criteria `Assignee: assigned` and `Answered: not checked`, then we could maybe just add another option to invert the criteria

Chefkeks:
> Yeah, I think I see your point: The criteria system doesn't allow for OR based criteria

Hadn't that in mind, but yes it's true the system doesn't allow for `OR` based criteria. So I guess either you think too complex or I'm thinking too simple ;)

Basically my thought / suggestion was to allow for multiple queues to be displayed, so you don't have to "click around" every time you want to see another (open/personal/closed/etc.) queue and can see all tickets that are open + <i>some-other-condition<i> on 1 page at a glance (grouped by queue).

Hope I could now better explain what my original intention was behind that suggestion.

Anyway, when I could give you or someone else from the osTicket team some inspirations for more possibilities or additional/better options regarding this coming feature I think all osTicket users will benefit which is surely a good thing (independent of you like or dis-like my suggestion).

greezybacon:
I've shored up the installer and added a How-To guide for installing this branch. Keep up the feedback!

Chefkeks:
Great!
Going to setup a test system as soon as possible... so hopefully tomorrow since I've some jobs left for the rest of this day :(

Chefkeks:
OK, luckily did found some time, so have it running an testing.

Despite some SQL errors I sometimes get to see, everything looks pretty good, but sure got some more detailed feedback below. I know understand a lot better the relationship between queues and sub-queues / top-level queues and so I guess my suggestion from a few days ago doesn't seem to make sense anymore to myself :D 

Randomly starting my feedback with this here:

I'd suggest to remove the width input field. The 3 vertical dots to change the size of a column is all it needs. As you can see I used a ridiculous width of 5000 and to my surprise it worked. I'd suggest to only allow 1 row for columns with a max-width value (which is similar to max-width used for queues) and don't allow re-sizing the columns any further, so don't allow for a second or third row for the columns. (Hope you understand what I mean).
image
In the case that all columns together are as wide as max-width and the user wants to add a new column, I'd suggest to reduce the width of the last column (since I guess a new column is appended after the last column) by 10(px / em / whatever it actually is) and insert the new column then with a width of 10. By the way - is the a min-width for columns?


Assuming the width input field is removed (but also if you want to keep it), currently I don't like the arrangement of the fields, so here's a suggestion how I'd arrange them:
image
The heading should be on top (since it's also on top of the column) and for the data in the column (which is then below the heading) display the options below.


The conditions look a bit lost, my suggestion is to re-work the design, so it looks more like the advanced search - did another painting mod therefore - quick&dirty, but I guess you know then better what I mean:
image

So far from my side for today - will add some more feedback tomorrow since I've now "Feierabend" ;) :D

Chefkeks:
I'd always make sure that at least 1 column is selected - so the user directly sees the options and disable that none can be selected when accessing the columns tab:
image

Chefkeks:
Would be great if there were an option to change the order / sorting of the queue (like it's possible for help topics) somewhere here:
image

Custom queues feature summarized: :+1:
I really like custom queues and the way they are implemented into osTicket - sure in some areas in my opinion it needs some little refinements, but overall it already look and feel pretty good (despite a lot SQL error, but I know it's far from being finished).

Btw. have to revert that feedback below since after playing with custom queues I think it's better to "hide" the edit and delete button in another sub-menu to avoid accidental clicks of these buttons. 
> I'd remove the sub menu to edit/delete a queue and display the button directly inline (maybe with a tooltip when mouse is over the button)
image

Cheers,
Michael

greezybacon:
You can click and drag to sort the queues

Chefkeks:
Hehe, ok definitely didn't try this since I did not see any hints / signs / symbols that it's possible to click and drag the queues. 

Maybe it needs to be more obvious that you can click and drag to sort the queues.

greezybacon:
Fair enough. I realized that I forgot to include the install file for the queues, which probably is responsible for your significant quantity of SQL errors.

Chefkeks:
Just had another idea/thought while reading your response. Regarding options which allow inheritance, it would be nice if you'd show the settings which come from an inherit queue, but making them look read only / disabled so the user knows what settings are coming from an inherit queue.
 If it's already implemented that way I must have missed it ;)

mfelber:
@greezybacon I have just downloaded the zip file from your site and followed the install instructions but some of the features that you are describing in your screenshot tour are missing in my test installation. For example the managing options of the queues in the "Columns"-tab:

columns

When I click on a column it won't get blue and nothing happens (also the configure options are missing).

The "My Tickets" queue is missing. Should it be there?

queues

I know the custom queue feature isn't completed but it looks like the managing options of the columns should be available in this "preview" (in the screenshots of chefkeks I see this options)..

greezybacon:
@mfelber I did notice that it doesn't seem to work in Firefox. Can you try another browser for your initial testing?

mfelber:
@greezybacon Yes, thanks for the hint! With Safari and Chrome I can use the configure options in the "Columns" tab :-)

Aka4:
@greezybacon 
I like @Chefkeks idea about displaying several queues in one page.

Can you add option to define GROUP BY column?

For example I would like to group some queues by Department or Assignee, that way it could be really good thing, below is modified version of @Chefkeks idea mockup:

image

or the same mockup, but grouped by agents:
image

In such way, in each group will be no column display differences.

greezybacon:
@Aka4 that's what the "quick filter" option is about. It allows you to filter the queue by another value such as the department or assignee

Aka4:
@greezybacon Yes, but filter shows us view of only one option.
It would be nice, if we could see all options at once, but grouped in separated sections.
It all about what GROUP BY idea is. Quick filter is great, but it's for different purpose.
If we have 10 agents, then it's faster to show them all at once.
Also another situation - if agent exist in two/three departments, then it would be great if he could group tickets by department.

greezybacon:
@Aka4 my opinion: that's a reporting-, not a real-time feature

philbertphotos:
This is nice VERY nice I am updating to this build ... a nice update would be the ability for the admin to create default views/filters that all staff members will get in addition to the ones they create.

xtheosx:
Will this work in 1.9.12?

I would love to use this.

mfelber:
As far as I know this feature will only be availabe in the next major release after version 1.10 (maybe called 1.11).

xtheosx:
Thanks @mfelber

Chefkeks:
After all those commits, I updated now to the latest version (currently bc3eb65) and here is my continued feedback.

**Icons!**
Queues should have icons - as in 1.9.x - so add a dropdown (or image selection etc.) for a queue icon:
image

Or doesn't that look better below? ;)
image


**Top-Level Queues**
I think the top level queue shall also be listed in the dropdown. I know you're maybe thinking that it's unnecessary, but whenever a dropdown opens I personally as a user tend to choose a option from the dropdown in first place instead of clicking the top/title of the dropdown itself. In addition this would make it also easier to edit/delete a top level queue due to the sub-menu which would then appear on the right. Currently it's not possible with 1 click. Beside that, in other parts of the system the top level option is shown as the first entry of a dropdown too, e.g. Dashboard > Dashboard or Tasks > Tasks (hope you get what I mean).
image

**Personal queues**
Currently personal queues seem to use the default layout of osTicket. My suggestion would be that personal queues rather use the defined layout of the top level queue under which they are. So for example I delete the column "Created" from the "Open" queue, but my "Personal Test queue" still shows it. My expectation would be that personal queues honor the layout of the top-level queue under which they are placed.

**Sub-Queues**
Looks OK in the admin panel:
image

But on the agent panel it's really confusing:
image

Btw haven't tested further but from a certain point, I don't think sub-queues make sense anymore, so maybe there should be a limit for sub-queues.

**Bug**
Organization Link points to queue instead of organization:
Config:
image
Link:
image


That's all so far... only thing I imagine for the future is an option maybe for department specific queues which can then only be accessed and seen from members of a department.

Cheers,
Michael

greezybacon:
### Major Update

Queues now re-use column definitions. When the system is installed, several columns are shipped with the system and associated with the various queues. Those columns can then be reused when defining other queues. This will significantly reduce the amount of work necessary to design new queues as well as make changes to columns used in multiple queues.

#### Columns are now managed as a table
This also avoids right-to-left issues with the previous model. Deleting is done via the trashcan button, when is cancelable (by refresh), and is unlikely to be accidently pressed
image

#### Select a column to add to the queue, or create a new one on the fly
image

#### Column editing is now done in a popup
*Name* is used when the column is shown in the drop-down list for addition to queues. The *heading* can be specified and translated when the column is added to a queue.
image

greezybacon:
@Chefkeks I agree about the icons. I've planned to add that

greezybacon:
I think the `%queue_column` table will need a `root` field as well so that the system can track what columns are valid for which queues, by matching up the `root` fields.

Chefkeks:
@greezybacon
Wow, that's a huge update! Really interested in testing the new look and feel (usability) next week :)

PS: Sounds really good with the icons :+1:

Chefkeks:
Ok, so here's my feedback regarding the major update plus the latest fixes:

I really like that `Columns are now managed as a table` and `Queues now re-use column definitions`, but....

when I played around with major updated queues for the first time now, I took some time to realize the difference between the heading (of a column) and the name (for the column definition) and that I could now easily define/add own columns with an own name (column definition) which I can then re-use later in other queues.

The way to easily re-use columns now is great, but the implementation could be better - unfortunately I do not have an idea how implement it in a way that it's clearer. Have to think about that and may do some MS paintings/mockups (based on time) how this could be clearer / more visible.

Beside that I still think that the top queue name (e.g. Open) should be repeated to be the first entry of the dropdown too and I'm still not a fan of the width field since when changing width I'd have to switch between `columns` and `preview` tab all time time until every column has a good width - so I'd prefer some click and drag mechanism to adjust column width. Maybe that's just a personal feeling, but apart from that custom queues look and feel a lot better with that major update!

Cheers,
Michael

nfebuary:
So I've been messing with ideas on Sorting Criteria UI. This is one of many thoughts. If the column is selected as a sortable column. The sort criteria will appear. With sortable options based off the fields selected only for that column. Utilizing the same system for search criteria with the added feature to drag a sort criteria to the top or anywhere in the list. If a browser does not have the ability to drag then we could degrade it to a number system. (not shown in screens) The top sorted field is the first criteria to be sorted and then the second item and so on and so on. Playing with the idea to minimize so many buttons i.e. delete, save, whatever else that will make it unwieldy I thought maybe a check box in the select box. When the item is checked it automatically appears in the list if unchecked it's automatically removed from the list. If you drag a field it auto saves the position on mouse release.

When the column is selected as sortable:
sort-criteria-screen1
Selecting Sortable fields for the column:
sort-criteria-screen2
Asc Desc dropdown options:
sort-criteria-screen5
After sortable fields are saved with properties. Same functionality as Search (click arrow to change settings).
sort-criteria-screen3
Removing a sortable field:
sort-criteria-screen4

greezybacon:
@nfebuary That looks neat. The checkboxes on the dropdown would require another widget, and there's no ability to sort by multiple columns though.

nfebuary:
I was thinking of adding the sort to the column settings instead of the Custom Queue. Then if you checked a box to allow that column to be sortable you get the sort criteria option for that specific column. This would also save time when configuring the Queue because the sort option would be done when adding the column. Does that make sense? In my head it sounds right. We're already saving the admin the agony of selecting 60+ options every time a custom queue is needed by breaking the columns out and letting them be reused over and over with minimal effort so why not add the sort options to that too? I guess my screens should reflect the columns concept.

BUT

If we prefer to have the sort on the custom queue we could have a list of columns when selected (maybe type ahead?) provides the above screens that would allow for nested sorting. Each selection would replace the current sort criteria with the respective column sort criteria.

We could utilize the popout box instead of dropdown to have a form inside so we don't have to add more plugins.

Chefkeks:
Personally I'd prefer nested sorting option, but I also really like the idea to save it with the columns instead of the custom queue.

... 5 minutes later ... 

Both options are great...

... Another 5 minutes later ...

OK, after some thinking I'd still prefer the nested sorting for a custom queue over the column.

nfebuary:
I'll revise the sort screens to include a way to select the specific queue and add the sort criteria to it.

nfebuary:
Select the Custom Queue form the list the dependent field to Add sort Criteria will appear as well as the title will change to "Sort Option for:" 
sort-criteria-screen0-rev1
The below screen displays numbers in the dropdown list (just an idea I had.) This would show if a Custom Queue has existing sort options or not.
sort-criteria-screen1-rev1
sort-criteria-screen2-rev1 copy
Select the Custom Column to be sorted. Each Queue will only display the columns associated with it for selection. Selection of a Queue can either be a system like the one displayed in the screen shots. Optionally a system similar to what is current when selecting the sort Criteria it will appear in the list with a delete button to the right of the line item. I personally like the check boxes in the dropdown with auto save on selection. Checking or unchecking the box would remove or add accordingly.
sort-criteria-screen4-rev1
Column fields associated with the specified column will be available for selection. If the column has only one field then that field is automatically selected. All options will have Ascending Descending sort options and the First item of that column has a checkbox to expose the sort i.e. arrow up or down in the column header to reverse the sort option for that column. If a column has more than one criteria the top level field will have the checkbox to change the sort order.
sort-criteria-screen5-rev1
sort-criteria-screen6-rev1
sort-criteria-screen7-rev1
The below screen shows the collapsed view displayed when selecting a queue that has sort criteria.
sort-criteria-screen8-rev1

greezybacon:
But if you're currently editing a queue, why would you have to choose a queue to manage sort options for? Should we do something like the custom columns where sorting options are managed separately via a dialog? That way they'd be reusable

nfebuary:
Ok I see. We would need to not have the Queue dropdown then.

nfebuary:
Ok, new screens and new function!

Sort options in the Queue list, it would show if there are any sort options for that queue, if so, how many. It would need a tooltip to let people know if you click on the number it pops up a modal to configure the options.
sort-list1
If a queue does not have any sort criteria this is the display the admin will have.
sort-modal7
If the queue does have sort criteria, it will be displayed like the advanced search. It would have a expand option to edit the existing sort and a dropdown to add additional sorting criteria.
sort-modal8
The rest of the function is much like the previous screens, only in a modal over the Queue List. If the Column has only one field then the field option defaults to that field. If the column has more than one field, the Admin can select the field and the sort priority is dependent on the order the sort criteria appears. Also, the top field for a column with more than one field has the "Expose" sort checkbox.
sort-modal3
sort-modal2
sort-modal5
For the delete icon I think It should fade in from a 20% opacity to 100% opacity on mouseover of the line item. On hover of the actual delete icon it would color red. This way it's not in your face until you need it.

Chefkeks:
Looks pretty good :+1:

But again directly a suggestion ;D
Add a reset button to revert things back to the last saved state of the sorting options (e.g. when admin messed up and wants to start over without having to press cancel and re-open the sorting settings/popup)

nfebuary:
There is a reset button on another modal that was located right next to the Cancel button, I would just locate it in the same spot on this modal. I can see the need for that, instead of selecting cancel and reopening up the modal you simply select reset while in the modal.

protich:
Quick feedback -- I haven't had a chance to test the new queue management update.

@greezybacon: :+1:  The update looks awesome -- love it.

* Not sure about inline edit on column table listing - some suggestions
  -  Change table columns to be  - ``Name (Label/Heading)`` | ``Column Name`` | ``Options``
  -  Clicking ``Name`` popups dialog (1) to set/change the column's ``Label/Heading)`` ``With``...etc
  -  Clicking ``Column Name`` will popup column management dialog (2) -- data source, name ...etc
  -  Adding a new column will result with the ``Label/Heading`` being inherited from column's ``Name`` by default (which will be the case 90% of the time) as well as suggested ``width`` for the column. This will help avoid *forced* localization.

* Complex Data Source (version 2) - I think we should support complex data sourcing like using data from multiple fields and formatting  (this is in addition to annotations...etc) . For example, on search results I might want to include state of a ticket in ``Status Column`` e.g - __Closed__- Frozen. 


@nfebuary: :+1: 
I think as Jared hinted, we need to think about placement as well as making sorting reusable across all queues. Following Jared's columns setup model makes sense to me.

Basically, the queue needs queue-specific sort options which will be available on the top right -- these can include fields or even combo of multiple fields which are not necessarily part of the queue's columns.  Any field can, potentially, can be a sort option but we want the queue designer to decide what makes sense based on data involved. For example, Closed Tickets queue will need to make Closed Date a sort option whereas the Open Tickets queue won't use such option.

* I'm not sure about sort getting its own column on the queue table or even if count # is necessary. When looking at queue listing, do I really care about number of sorting options? If maybe, then perhaps an icon with tooltip will do? 

* The complex sorting that involves multiple fields e.g ``Priority + Create Date`` will require sort order for each field.


That's all for now!

Cheers,

philbertphotos:
Are the changes updated yet?

nfebuary:
### New Sort Option Screens.

The idea is to place the sort in a Sort tab on the Queue edit. This Tab will list the sort options with the option to delete or add from a list or create new sort criteria that can be used for any queue. There will also be default Sort Options out of the box.
sorting options0
When a sort item is added to the list it is removed from the "add sort" drop down for this queue.
sorting options1
The default sort for the queue cannot be removed. This default sort setting will be applied in the queue criteria tab. The sort options will have to be added before setting the default sort. Each sort option title is a link to open a modal and add more fields or remove them.
sorting options2
Selecting "Add new sort criteria" will open a modal with options to add fields and a title.
sorting options2b
sorting options3
sorting options4
sorting options5
After saving, the new sort option will automatically be added to the list for that queue and can be accessed by other existing and future queues, If it is needed.
sorting options6

goobersmooch:
Love the queue system! I'm having an issue trying to add a new column to the default open queue. For example when i add Last Message as a new column and click save changes i get the following error: Fatal error: Call to undefined method QueueColumnGlue::setSortable() in ...\include\class.queue.php on line 168.

greezybacon:
Yeah, oops. I'll fix that up

goobersmooch:
When trying to claim or assign a ticket i just get a blank window.

screen shot 2015-11-17 at 11 14 47 am

greezybacon:
> When trying to claim or assign a ticket i just get a blank window.

What does that have to do with the new queues feature...? Did you get a PHP error?

goobersmooch:
Sorry.. thought it might have something to do with some recent commits. No PHP errors.

greezybacon:
I'll try and check it out. Thanks for testing the feature

goobersmooch:
I think i figured it out. When i tried a fresh install of your custom-queue branch on one of our mac VM servers i was able to see some PHP errors. 

When i tried to claim or assign a ticket to an agent i got this error: Call to undefined method AssignmentForm::getId() in ../include/staff/templates/assign.tmpl.php on line 31.

When i clicked the transfer button i got this error: Call to undefined method TransferForm::getId() in ../include/class.forms.php on line 216.

I noticed in 4c18b47 that you changed line 49 of ../include/class.forms.php from function getId() to function getFormId().

What i did as a workaround was to replace getID with getFormId on the above lines that were throwing the errors.

jdelhome3578:
Error on sort drop down: `Warning:  Invalid argument supplied for foreach() in C:\inetpub\osTicket\DPR\include\staff\templates\queue-sort.tmpl.php on line 14`

greezybacon:
@goobersmooch thanks for your hint. I applied the fix in e147ccd

kest874:
Testing this out, I updated an existing 1.10rc2 database only immediate issue I am having is this error in the search dropdown:

image

greezybacon:
Judging by the CSS issues, it seems you might not have everything applied correctly. However, even so, that is not indiciating an error preventing things from working corrrectly

goobersmooch:
I'm also having similar issues. See screenshots below:

screen shot 2016-01-07 at 8 35 03 am


screen shot 2016-01-07 at 8 35 30 am

greezybacon:
Sorting still isn't done yet. But the help topic list I've yet to see. Thanks for the post. Do you have one or more system secondary languages setup?

kest874:
I checked out class.search.php from your branch and it cleared up the css issue, but it's still throwing the error in this function:

image

greezybacon:
@kest874 did you do a fresh install? We've changed our minds several time on some of the data storage without implementing intermediate upgrades. Also, I think it is still possible to create a queue without any criteria, which might also create that problem.

goobersmooch:
@greezybacon I do not have any secondary languages setup.

kest874:
It was an upgrade of an existing database.

Fresh install seems to work ok with minimal issue:

image

Keywork search appears to be broken (red arrow) however if you used the advanced search and use only the keyword it works fine. Notice however that the non saved searches have nothing to indicate what you searched for (highlighetd area)

kest874:
@greezybacon Playing around with this I've noticed a couple issues:

queue=x isn't added tothe paginate url:
image
So you can't page through the tickets in the non default queue my quick and dirty fix : https://github.com/kest874/osTicket-1.8/commit/6ec074f70bd195d250ad42b89a692b471b120ea1 

Second when you create a new ticket with department notification on you will get an error:

Warning: array_merge(): Argument #2 is not an array in /var/www/ost/include/class.ticket.php on line 1214
image

The wrong object is being returned on line 1215 (ModelInstanceManager) not (Staff)

kest874:
line 1216 above works when changed to:

$recipients = array_merge($recipients, $members->asArray());

martincasalis:
Hi,  this mods works in v1.10?

greezybacon:
For anyone following this thread, I've rebased it on top of the updated `develop-next` codebase

kest874:
image

balojs:
👍

kest874:
@greezybacon Testing from the tip of this commit it's shaping up nicely, however I've noticed a couple issues which all might be related.

When you attempt to create a new ticket you pick user etc and fill out the summary/description etc and submit, you will get an error: "Missing or invalid data — Correct any errors below and try again" the cause is the email address and name are for some reason missing or incorrect even though it's displayed.

I have also noted that if you go to users or organizations and try to "edit" one of them you will never get the popup and no errors are logged.

Have any thoughts as to where to look to fix either of these?

greezybacon:
@kest874 thanks for the heads up. I'll look into those issues

kest874:
Looked again at this and getting error: PHP Fatal error:  Unsupported operand types in /var/www/osttest/include/class.orm.php on line 128, referer: http://osttest/scp/users.php and orgs.php

image

kest874:
@greezybacon Also note that 3827731... orm: Add support for MySQL FOUND_ROWS() breaks the highlighed:

image

and they no longer appear.

kest874:
@greezybacon 04f2066... queue: Add support for basic search, user- and org-id searching 

is broke ticket creation:

> When you attempt to create a new ticket you pick user etc and fill out the summary/description etc and submit, you will get an error: "Missing or invalid data — Correct any errors below and try again" the cause is the email address and name are for some reason missing or incorrect even though it's displayed.
>

kest874:
@greezybacon  queue: Add personal queue support  04f2066 broke the user and org popup:

This change breaks it:

image

kest874:
@greezybacon Also note that 3827731... orm: Add support for MySQL FOUND_ROWS() breaks the highlighed:

image

and they no longer appear.

Moving the following to the original position resolves:

image

kest874:
@greezybacon one last issue I have found and not been able to work around, when I have %{message.files} in the template class.mailer.php is throwing:

> Fatal error: Call to undefined method ModelResultSet::getFile() in   var/www/osttest/include/class.mailer.php on line 429
> 

Any clue as to where/why this is happening?  I can't reproduce with develop. Can you provide any guidance on where to look?

kest874:
@greezybacon noticed in this section where the error is happeing:

image

Develop is returning: object(Attachment)
Custom Queues is returning: object(ModelResultSet)

Thoughts?

kest874:
@greezybacon looking through this when you try to attach a file on new ticket it's throwing an error that argument 2 is not an array in class.variable.php line 139:

image

greezybacon:
@kest874 thanks for your patience, and your terrific insight. I fixed the first three issues you posted in f2141fb. The last, I think is a regression which was fixed in the `develop` branch, so I rebased on the `develop` branch. Would you mind testing again?

Also, beware there was an issue with the SQL migration scripts which created the primary key incorrectly on a new table. You can review the changes in a7744ef to make the changes to your database manually, and then change your `schema_signature` config to match the one in `include/upgrader/streams/core.sig`.

include/mysqli.php
```diff
@@ -181,11 +181,6 @@ function db_create_database($database, $charset='utf8',
 function db_query($query, $logError=true, $buffered=true) {
     global $ost, $__db;
 
-    if ($__db->unbuffered_result) {
-        $__db->unbuffered_result->free();
```
protich: Curious to know why you removed the concept of auto-free.

greezybacon: It's freed by the ORM when the cursor reaches the end of the result set. So the double free is part of the issue. This was only used by the export system for database backup. It should be rewritten to use the ORM...

kest874:
@greezybacon first 3 issues are resolved, the file attachment is still an issue:

In develop the array is:

> array(1) { [0]=> object(Attachment)#915 (6) { 

In custom-queue the array is inside another array:

> array(1) { [0]=> array(1) { [0]=> object(Attachment)#923 (6) { 

**Error thrown :**  Call to a member function getFile() on array in var/www/osttest.nasg.net/include/class.mailer.php on line 431
image

kest874:
Found the issue with attachments,  I had a workaround for an early issue where class.variable wasn't passing content as a variable

> $this->extras[$type] = array_merge($existing, array($content->getContent()));

Should be

$this->extras[$type] = array_merge($existing, $content->getContent());

I'll continue to look for issues, but the 4 I mentioned above are working as they should now!

kest874:
If you setup a queue and you use "TicketStatus / Status Name"  nothing is returned in the preview, traced to:

class.search line 967:

image

The name is coming as the status__name, however the resultset needs the ID if you force to $name = 'status__id'; you get the result your looking for.

kest874:
When the predefined sorts are added via upgrade the root needs to be set as T for them to appear in the dropdown, they are currently inserted into the queue_sort table with the root set as Null

kest874:
Also noticed that usersearch typeahead does not return just the typeahead users tickets.

kest874:
When using advanced search and using "User / Email Address" it produces the following error:

image

kest874:
Noticed that the cache of tables for some reason keeps dropping I get this error: 

image

Restarting apache seems to resolve it for a little while.

kest874:
Noticed that the tasks tab is broken:

> Fatal error: Call to a member function getSortOptions() on null in /var/www/osttest/include/staff/templates/queue-sort.tmpl.php on line 2

rbukovansky:
How fast can we get this to main repo? That's just awesome! @greezybacon Nice job! Well done!

rbukovansky:
I have created pull request https://github.com/greezybacon/osTicket-1.8/pull/7 to temporarily resolve issue with tasks queue.

kest874:
@greezybacon 

image

Looks like this is an issue with a missing column. Commenting this in class.queue.php resolves:

image

kest874:
Noticed issue with date due as a column, if the field is null it throws an error when it tries to format the date.

Added the highlighted to the following function in class.queue to resolve:

image

greezybacon:
@kest874 thanks again for the problematic feedback. I think I've addressed the items I added a 👍 reaction too. For the `default_sort` bugs. I haven't seen them. Are you sure you upgraded correctly? There was a mishap in the upgrade script I rewrote. Perhaps you should try a clean install and compare the database schemas with your semi-production one?

rbukovansky:
@greezybacon I have few issues with this:

1. Agent Panel -> Tickets: Edit and Delete options for custom queues are not shown for me. Am I doing something wrong? (installed greezybacon:feature/custom-queues branch)
custom_queues_missing_edit_delete

2. In Admin panel -> Settings -> Tickets -> Queues -> Columns: There is no "management" of custom columns. I can't edit or delete any of my columns in this combobox (I'm guessing that system defined columns are not going to be editable/deleteable).
admin_columns_missing_edit_delete

3. The queues "bar" is not alligned properly.
alignment_queues_menu

4. When setting new column up, there is no possibility in Conditions to say: Assignee = CurrentlyLoggedInUser() -> make background red. I would like to make Open queue with highlighted tickets assigned to currently logged-in user.

5. There is no possibility for Agents without Supervisor privilegium to define their own columns.

6. No possibility to define Conditions for rows. For example, over due ticket would have the whole row with red background.

Could you please take a look? Thanks a lot.

greezybacon:
@rbukovansky I updated the screenshots. The current implementation deviated from the original RFC and originally posted screenshots. 
 1. should be exampled by the current screenshots now.
 2. click the [Config] button after the column is added, and you can "edit" it. You're correct. There is no delete functionality currently. I added it to the outstanding list.
 3. I've noticed this as well. I'm no CSS expert, and we used a third-party library to handle the menu overflow. Looks like a bit more customization will be necessary
 4. Select "Assignee" is "Me" (CurrentlyLoggedInUser())
 5. That is correct. The idea was to make the system more flexible without making it more complicated. Allowing anyone to create new columns for use personally would clog up the UI, IMO
 6. I like you're idea. I'll see about adding a checkbox for that.

rbukovansky:
@greezybacon Thanks.

1. But those screenshots don't explain how to do that, if it's possible at all.
2. I see... So we can't add that [Config] button (as well as the Delete one) next to Add button which is next to that combobox with columns? This could be easier, I guess, than adding column to queue, configuring it, then deleting from queue...
Well, it still leads me to separate tab called Columns next to Queues tab... (Just a thought...)
3. OK
4. Well, in that case it's some issue with i18n, because it doesn't show Me/Já or One of my teams/Jeden z mých týmů (Czech translations) when I'm using Czech language (I have generated messages.pot from your custom-queues branch, merged with my current 1.10 messages.po, translated the new strings, generated new message.mo.php and then cs.phar).
column_conditions_assignee_missing_me_one_of_my_teams
5. I see, but it's more about allowing them use different colors for column backgrounds as they wish etc. But I can't probably live with that.
Hm... If you think that those "personal" columns would show only to them, not for other users or in main queues definition (admin panel), it wouldn't be that clogged up. Again, just a thought...
6. OK.

Additional:
7. Is this some kind of empty placeholder for future icon? Or is that text just misaligned?
columns_menu_missing_icon
8. Is there a plan to add this funcionality (custom queues) for Tasks as well?

kest874:
> @kest874 thanks again for the problematic feedback. I think I've addressed the items I added a :+1: reaction too. For the default_sort bugs. I haven't seen them. Are you sure you upgraded correctly? There was a mishap in the upgrade script I rewrote. Perhaps you should try a clean install and compare the database schemas with your semi-production one?

I compared a new install to the db I'm using and it looks correct.  I will continue to test.

kest874:
@greezybacon quick question, is there any way to set the standard columns? 

Also notice that when you search at times the checkbox is not editable:

image

kest874:
Found the issue with the default sort, the sort_id in _queue is set to NULL instead of 0, also it's impossible to set the default sort back to none.

kest874:
One suggestion I can submit a PR for it if you like, it's nive to have a breadcrumb for the queue your in:

image

One enhancement to the above would be to make each crumb clickable, to take you to that queue.

kest874:
@rbukovansky You can edit queues by clicking the gear:

image

You can also delete queues in the admin panel:

image


@greezybacon still to do is the ability to delete personal queues.

rbukovansky:
@kest874 Oh, well... I would prefer those edit/delete icons next to name of queue in queues menu, but OK. Thanks.

The personal queues are editable/deleteable using cog wheel next to name of displayed queue. But the delete function of personal queue is not working. @greezybacon Could you please take a look?

Thanks a lot.

kest874:
@rbukovansky They were originally there but made the drop-down too busy.

greezybacon:
@kest874 from the CustomQueue class definition, `getColumns` method, around line 471, the queue with `id` == 1 is used as the "standard". This is the "Open" queue (although it could be renamed and repurposed now). Perhaps we could add an option somewhere? We also added a feature in the database model to use a drop-down list. So we could do something like:
```
Use columns from queue:  [ Open          | v ]
                         | Closed            |
                         | My Queue          |
                         `-------------------'
```

rbukovansky:
@greezybacon I have found that you are autocapitalising the names of the queues. Please, don't do that, we are not using it in Czech language and if you do that it looks really weird (see screenshot). I believe this would be valid for other languages too.

queues-names-autocapitalised

Thank you.

ntozier:
Ummm.. Every menu item, every configuration item, Agent names, Priorities, across all of osTicket has the first letter capitalized...  Why wouldn't it be capitalized?

kest874:
I agree, see no need to change the capitalization.

rbukovansky:
First letter is fine, but look at those items marked with !!!. Mé Požadavky should be Mé požadavky, Priorita S Podbarvením should be Priorita s podbarvením...

kest874:
@greezybacon I like the drop-down idea for the standard columns.  I did some testing and it appears that it's always bailing to the last resort columns.

greezybacon:
@kest874 the columns list is currently only editable for *saved* searches. Ad hoc (on the fly) searches have to use the "standard columns" currently. I think that restriction can go away with the drop-down list.

greezybacon:
Personally, I oppose capitalizing even the first letter. It works in most languages, it kinda works, but considering the CJK languages (Chinese, Japanese, Korean), it falls apart. I think we should remove the code

rbukovansky:
Please do. Thanks.

kest874:
@greezybacon  hopefully the standard columns can also be sortable :)

kest874:
Something I added in my production implementation is an additional "Ticket Filter" int he column setup called "Ticket Link with Thread Preview", I'm using it on the subject field:

image

When you click the thread row it will expand:

image

The title at the top also takes you to the full ticket.

If there is interest I can package it up for a PR.

samogon13:
b9bd86e4-608e-11e6-8795-2a1217e2049a

ipAnything:
I have read somewhere that this will be added to v1.10. Is this the case or would one have to wait for v1.11? Will it be included in 1.10 RC4 or 1.10 final release?

rbukovansky:
@ipAnything Best guess it's gonna be in 1.11. osTicket 1.10 feature-wise is complete (no new features in this version are allowed).

ntozier:
@ipAnything It will not be in 1.10 RC4 (if there is an RC4) or 1.10 final release.  It should be in whatever version is released after those.

ipAnything:
What a pity 🙁
This is a very useful feature.

Get Outlook for iOS

		_____________________________

philbertphotos:
Will the pull be compatible with with 1.10 final

greezybacon:
I've rebased it on v1.10 final, if anyone would like to upgrade and continue testing

markus4000:
HI Jared, GREAT FEATURE! 

Testing:

Assigned Staff: "does not have a value"
assignedstaff
does not work: Result is empty. => expected: all tickets that are not assigned to an agent, but can be assigned to a Team
Assignee: "uassigned" works. => shows all tickets that are not assigned to a team and not to an agent

"Save changes" on the criteria page does not work. it only works after activating the preview tab or another tab.

Deleting a "Personal Search" or "Personal Queue" does not work. 
=> I added a search, activated it, when i click "Delete" nothing happens.
deletequeue

Personal Queue and Search, filter by "Ticket Status Name" does not work. No results.:
ticketstatusname
Status State works

groovybluedog:
I am trying to set up a condition on a due date column, so any time in the near future will show up with conditions. For example, all due dates/times with less than 6 hours to be coloured yellow... however no condition seems to work. I thought "In the next X days" > "6 Hours" would work, but it doesn't. "More than X days ago" and all other ones also don't seem to work for this. Hmm... :c

greezybacon:
@groovybluedog nice catch. I was able to fix that one. @markus4000 I was able to implement the queue delete link. I'll continue to work on verifying and fixing the other noted items. Also, I can't find who requested it, but I added conditions to the queue itself, so one condition can apply to the whole row.

kest874:
@greezybacon noticed with commit(queue: Fix search by status name https://github.com/greezybacon/osTicket-1.8/commit/1ad9028eef67015dda97a4da2d5a441427bb4010) that it blows up.

I get:
'Fatal error: Uncaught exception 'InconsistentModelException' with message 'Unable to prepare query: SELECT SQL_CALC_FOUND_ROWS A1.`number` AS `number`, A1.`ticket_id` AS `ticket_id`, A1.`source` AS `source`, A3.`priority` AS `cdata__priority`, A2.`name` AS `status__name`, A1.`created` AS `created`, A1.`topic_id` AS `topic_id`, A6.`name` AS `user__org__name`, A5.`org_id` AS `user__org_id`, A3.`subject` AS `cdata__subject`, A1.`isoverdue` AS `isoverdue`, A5.`name` AS `user__name`, A1.`user_id` AS `user_id`, A1.`staff_id` AS `staff_id`, A1.`team_id` AS `team_id`, A1.`isanswered` = '0' AS `2GkBZxz`, AS `B1xWyh0`, AS `Dhz0GIp`, AS `wTTxn5N`, (SELECT COUNT(R0.`id`) AS `count` FROM `ost_thread` Q7 JOIN `ost_ticket` Q8 ON (Q7.`object_type` = 'T' AND Q7.`object_id` = Q8.`ticket_id`) LEFT JOIN `ost_thread_entry` R0 ON (Q7.`id` = R0.`thread_id`) WHERE Q8.`ticket_id` = A1.`ticket_id` AND NOT R0.`flags` & 4 != 0) AS _thread_count, (SELECT COUNT(R1.`id`) AS `count` FROM `ost_thread` Q7 JOIN `ost_ticket` Q8 ON (Q7.`object_type` = 'T in /var/www/osttest/include/class.orm.php on line 3398 '

The change here is causing it: 

image

Looks like class.search doesn't contain anything for TextboxField

philbertphotos:
I get a lot of those

groovybluedog:
@greezybacon I downloaded the ZIP again and overwrited (I know, bad grammar) over the old copy, and if I have both the "in the past" and "in the next X days"conditions set on one column at the same time, I get this error:

Warning: Illegal offset type in /osticket/include/class.forms.php on line 2048

Warning: Cannot modify header information - headers already sent by (output started at /osticket/include/class.forms.php:2048) in /osticket/include/staff/header.inc.php on line 2

Does it support having two or more conditions on the same column?

greezybacon:
@kest874 @philbertphotos @groovybluedog I cannot reproduce any of your issues. Could you maybe send some other information about how you're using the features and where you're triggering the errors?

> Does it support having two or more conditions on the same column?

No, that wasn't a design goal; however, I was able to add two conditions for the same column, and it worked without any fuss.

greezybacon:
@protich I'm reading back through the thread and found your comments on the complex data source:

> Complex Data Source (version 2) - I think we should support complex data sourcing like using data from multiple fields and formatting (this is in addition to annotations...etc) . For example, on search results I might want to include state of a ticket in Status Column e.g - Closed- Frozen.

Is this something you're looking for at this point in the request?

philbertphotos:
Ok i guess custom queues does not upgrade an existing 1.10 final. 
I made a sql copy of my environment and did a upgrade and nothing worked right.
But doing a clean install did the trick.

When will this be able to update a running version of 1.10 final????

greezybacon:
@philbertphotos as far as I know, it is able to perform the upgrade. Can you help shed any more light on what went wrong during the upgrade so I could address it?

kest874:
> Ok i guess custom queues does not upgrade an existing 1.10 final.
I made a sql copy of my environment and did a upgrade and nothing worked right.
But doing a clean install did the trick.

Doing some testing it's definitely DB related because a new install works normally.  I'll do some digging to find out whats going on.

@greezybacon I get the error when it starts to display the queue /scp 
image

kest874:
@greezybacon small issue with dates in the columns when they are null you get:

image

Might make the change here: https://github.com/kest874/osTicket-1.8/commit/6d097887cc7ef422bf35a0bcd79b3fe75474c2c0 to correct.

kest874:
@greezybacon traced the DB issue to the conditions on the "Status Name" column, and sure enough if I try to edit the conditions set as follows:
image

I can not pick anything from the list.  This change breaks it:

image

I also see this error in the log when I switch to the conditions tab: ` PHP Warning:  Missing argument 1 for QueueColumnCondition::__construct(), called in /var/www/osttest/include/ajax.search.php on line 316 and defined in /var/www/osttest/include/class.queue.php on line 1378, referer: http://osttest/scp/queues.php?id=1`

greezybacon:
Okay, @kest874 I was able to address the date column crash, and I fixed the Warning output from editing the queue.

greezybacon:
@kest874 I read your post about the search on ticket status name several times, and it finally soaked in. So I got it fixed as well. I also fixed the issue where editing the search criteria might not save right.

kest874:
@greezybacon I can confirm that both are fixed no issues.

I've been using custom queues since the beginning, have nothing but praises!  Will continue to test and report back.

philbertphotos:
I will try another upgrade and see what happens...
I think it messes up because I already had searches in the old system but i am not yet sure.

kest874:
@philbertphotos  you might need to drop the current queues and recreate them.

kest874:
@greezybacon noticed an issue with deletion of a personal queue:

image

I do not get an error or anything in the log so I'm not sure what the issue is.  This is with a new db.

kest874:
@greezybacon queue-delete-confirm.tmpl.php is missing from the template directory....

greezybacon:
I probably forgot to add the template file ...

greezybacon:
@kest874 okay, I fixed it. I also fixed an issue with inherited columns on a queue with conditions

markus4000:
@greezybacon
Status Name Filter => works
Assigned Staff: "does not have a value" => works
Deleting Queue => works
Save Changes => works

but the column, "Status Name" is still empty.
I tried it with a clean new installation:
statusname

kest874:
@greezybacon Confirmed that it's all working now.

@markus4000 try removing that column and then re-add it.  It's working fine here:

image

Chefkeks:
Following this PR since the very beginning and regarding the recent activity just wanted to let you know that y'all doing a great job getting the bugs found and squashed! 👍 🎉

markus4000:
@kest874  I installed a totally new instance. How can I remove and re-add ?
i did a personal Queue. I added the column and removed it and added it again. but it's not working.
how can I check that i have the Version with the fix?

kest874:
@markus4000 Do you have all of the commits?  You might download (or pull) again from https://github.com/greezybacon/osTicket-1.8/tree/feature/custom-queues and overwrite your setup, I can't duplicate the issue.

kest874:
@greezybacon I am seeing one oddity, for some reason the counts are double in the dropdown menu It's counting 8 and there are only 4 (verified):

image

kest874:
@greezybacon The count issue appears to be db related, as removing and recreating the queues resolves the issue.

kest874:
Suggestion, and it may have been mentioned above, but it would be nice if the search used the columns and sorts for the default primary queue instead of the hard coded ones. (This is just using the test search, not advanced)

kest874:
Suggestion might be handy to allow "Clone" from the admin side via a dropdown.

markus4000:
@kest874 @greezybacon  : i downloaded the latest pull. but the status name was still not there. then I
turned back this one: 
https://github.com/greezybacon/osTicket-1.8/commit/e7b178805d6e8866866d013b545f83d7678d0da5
manuallly. and the status name appeares and it works.
But in the aearchfield, Status Name isn't a dropdownmenu anymore. so it is not useful too.
Wher do I make a mistake? I cant find... even if I delet the column and add it again, it doesn't help.
Any ideas?

kest874:
@greezybacon Another suggestion/issue

When you add conditions:

image

after adding an element like background color you can then again add the same element, which causes confusion and may not give you the result you want.

Also it's not possible to chose the same field again without saving or choosing another field and deleting the field.  

I would like to chose the same field an apply different conditions and elements without the workaround.

Chefkeks:
And I'd still like to see that conditions look like this as commented some time ago: https://github.com/osTicket/osTicket/pull/2577#issuecomment-147736335

image

JimssonJim:
Hey @kest874, did you find out the reason for the ticket amount being wrong on the queue drop-down? I'd really like to get my numbers correct.

And if it's any help, the numbers on my setup are not double, but it varies. Sometimes 1 ticket shows up as 84, sometimes 6 tickets as 179. And it seems to change from day to day even if the actual ticket amount doesn't change. This is running on a version of this branch that was downloaded a couple of months ago.

markus4000:
I found the reason why the Status Name was not displayed. 
I dont know why, but this field was empty. after choosing "Status Name" it works fine.
statusname1

kest874:
@JimssonJim I had to start over with my queues to get the numbers correct.  The format of the criteria and what not is different than it was, I found it easier to recreate than to manually change them in the db.

JimssonJim:
@kest874 Do you think I need to re-download and update the whole branch?

kest874:
@JimssonJim  I would at a minimum pull all changes or download from that branch and see if it them works, if not you may need to recreate the queues.

markus4000:
The export function does not export the current view with the current collumns. it would be great to have the actually displayed columns exported too.
export

ntozier:
I feel that is outside of the scope of this PR and should be handled in a separate topic.

The export feature has some open issues currently last I checked, maybe on of those.

greezybacon:
I started a feature some time ago to for exporting configuration. Perhaps we should have a special queue which defines the columns to be exported? I think currently way more columns are exported as what is shown on the screen. Then, perhaps the queues could have a setting to use the default export columns, or those defined in the queue?

How should the annotations and conditions be exported? Should those be separate columns or just ignored in export?

markus4000:
@greezybacon  I like the idea to have a special queue that defines the default fields and the option to choose if the default or the actually showed fields should be exported.

Annotations and conditions can be exported as a keyword, but I think it is not that important.

@ntozier you're right, its not in the scope, but perhaps @greezybacon can bring it together. but its more for a reporting and statistics and not for realtime work.

kest874:
@greezybacon another suggestion for "Department" for teams we have:

image

Can we do the same same for departments, offering "Mine" or "One of my Departments".  I have several individuals that are in more than one department.

kest874:
@greezybacon Enhancement with defining the criteria of a queue it would be nice to be able to use "And"  or "Or" between selections.

By default they are all **AND**, but I might want a queue that shows a specific department **OR** assigned to me, by default its only specific department **AND** assigned to me.

JimssonJim:
The Dashboard isn't working quite as it should, even though the graphic plot seems correct.

image

-Jim

PS. Same as in issue #3313

rbukovansky:
@greezybacon Would you be please so kind and merge latest `osticket:develop` changes? Thank you.

rbukovansky:
@greezybacon When I did fresh install from this branch, when switching to Agent Panel I'm getting this error:

```
Fatal error: Uncaught Error: Call to a member function getId() on null in /var/www/support-test.vahynetto.cz/scp/tickets.php:121 Stack trace: #0 /var/www/support-test.vahynetto.cz/scp/index.php(17): require() #1 {main} thrown in /var/www/support-test.vahynetto.cz/scp/tickets.php on line 121
```

Could you please take a look?

Thank you!

freeday100:
Hi, I've got following issue: osticket was working fine til last week, when it started to showing "Unable to create ticket. Correct any errors below and try again." All required fields are cleared and they have to filled once again. After that ticket is created succefully. So every time we need to filled required fields once again.

AndrewJordaan:
@greezybacon, To me this is the greatest since sliced bread, the install (Fresh 10.1) was painless and the zip upgrade when in like a dream.
The Que system for Status works great, as soon as I understood  the logic and its easy.
The custom columns are magnificent.
However,
I am trying to get a hierarchical colour code using "Create Date " with a "more than n days ago " and it looks like I cannot use the condition  numerous times as it seems to only register the last " n day "
The idea I am trying to get is say :
One Day     = green background,
Two days    = Yellow Background,
Three Days = Red Background

Aslo , a minor but after the upgrade the system is telling me
" Server Information     osTicket Version 	1.8-git (?)      — 901e5ea is available  "

the Zip is Two days old

groovybluedog:
**"greatest since sliced bread"**

A TOAST to queue management! 

Hahahahaha... sorry... xD

How would this system be handled when osTicket eventually releases a responsive SCP? Would this too be responsive, and contents simply wrap? What about mobiles? For example, a different version for mobiles, where I would select less columns?

rbukovansky:
@greezybacon [FeatReq] Could we please get possibility of showing tickets which are for organisations I'm manager of? Our sales department would like to be able to see tickets this way as they are in our osTicket installation assigned as organisations managers.
Thanks.

ntozier:
Are you asking for the possibility to add an "Organization" field on the custom Queue?

freeday100:
Hi. I'm using version osTicket (1.8-git) - it's Custom Quenes version.
Osticket was working fine til last week, when it started to showing "Unable to create ticket. Correct any errors below and try again." All required fields are cleared and they have to filled once again. After that ticket is created succefully. So every time we need to fill required fields once again.

ntozier:
Please keep in mind that the "custom queues" feature is alpha software and for testing only.  You should not be using in a production environment.

If you are having errors with this feature please provide PHP, Browser, Webserver errors, and maybe a screen shot indicating whats going on.  Thank you.

rbukovansky:
@ntozier No, I'm asking for "Tickets of organisations I'm manager of". Makes sense?

freeday100:
@**ntozier** Same issue appears on different browsers. Here are server specs:
Oprogramowanie serwera WWW 	Apache/2.4.10 (Debian)
Wersja MySQL 	5.5.53
Wersja PHP 	5.6.29-0+deb8u1

Apache error log does not show any errors during creating new ticket (error log in attachment).
[error.txt](https://github.com/osTicket/osTicket/files/884677/error.txt)

kest874:
Please alloow this be a 1.11 milestone....

TetraAsh:
ISSUE: 

OK, I have a custom list (pending quote, quoted, accepted. invoiced, installed)

from here, i have a custom form thats added into new tickets. 

next i created a custom queue (active) its got many standard columns and at the end, the data from the form/list. 

i created a couple of sub queues under 'active', these all work fine. 

lastly, i created a queue where the criteria is based on 'sales status / sales status' has a value, some of my tickets do and some dont.

when i create teh last column the numbers for all the ticket counts go very wrong, 

i click the field and it shows 6 tickets but counts 152 (i dont even have 152 in the system yet i have a total of about 60) 

same that all the other queues also display incorrect numbers now,

active queue with 54 tickets counts 128
sub queue with 3 tickets counts 16
sub queue with 51 tickets counts 112
sub queue (sales) with 6 tickets counts 152


have created a databased sql dump if you want to contact me. its all demo data anyway

LithiumNL:
After some search work i found this branch and it sound really good, just what i was looking for!
Last week i introduced this ticket system to my company and one for the questions was, can we have the organization on the column. This branch does the job!

After using it for a few days i am overloaded with error messages from the sql database:
afbeelding

I tried to find out what the problem is but i am not able to find it. Can you help me with this problem?

greezybacon:
Looks like expressions q31, q32 and q33 aren't compiling correctly. Could
you submit some screenshots of your queue configurations so I could try and
reproduce the error?

LithiumNL:
@greezybacon of course i can. Do you prefer a database screenshot or just the one on the webpage?
I edited the que that came with the upgrade, the open que list, added the organization in the column.
afbeelding
afbeelding
Annotations and Conditions are empty.

greezybacon:
Ok. I think the query you posted is for queue count collection. Did you add
some new queues? Or advanced searches? What do the "Criteria" pages look
like?

LithiumNL:
Yeah this error comes from the counting. 
It doens't count: 
afbeelding
I only added one queue, called my tickets. I didn't change anything about the Criteria (this one is from Open/Unanswered:

afbeelding

Just after installing this branch i got these error's.

greezybacon:
Ok. Thanks again. Did you do an upgrade or a fresh install? If an upgrade,
what version / git commit did you start from originally?

LithiumNL:
It was a fresh install, downloaded the files from osticket. After that I downloaded your branch and pasted the file on the server and did the upgrade.

greezybacon:
Ok. So you started with 1.10.1 or a different version?

LithiumNL:
To be exactly i think it was 1.10 and didn't upgrade to 1.10.1.

SolRoffe:
Hi! this code I can use for 

v1.10.1 (9ae093d)??

greezybacon:
@LithiumNL I have not yet been able to reproduce your issue. I think you can use a database editor and edit the `ost_queue` table. Delete queues with IDs 31-33, and I think it should start working again.

@SolRoffe and everyone else: I rebased the code on v1.10.1. Feel free to update your branches. Feedback welcome.

I added publish support for saved searches. They can be shared with other team members or with everyone (if you're and administrator).

image

@protich @JediKev I reused the `staff_id` column to contain the team ID. I think it would be nice to add a `team_id` column so the original creator will not be lost when sharing with a team. Would you be in support of changing the database patch or adding another? Or do you disagree with the sharing feature?

LithiumNL:
@greezybacon thanks! Will delete them and let you know if it works.
What's the best way to update? First update to 10.1 and then pull your branch or just pull your branch over 10.0?

greezybacon:
My branch includes 10.1 now, so you could just pull it again.

LithiumNL:
@greezybacon deleting id 31 to 33 did the job. That were queues from an older search.
Thanks for looking and helping me out :)

mahendrasahani:
@greezybacon I am using osTicket (v1.10.1) and I am getting me all queries from my customer through email. All emails are getting unique no of 16 digit as order no as e.g ( 1577-6060-1969-4357 and 1576-2573-3299-1165 etc). While I am searching order id of 16 digits I am not getting search result of exact order id. 
Please see the screenshot and my queries.

1) I have created rule as saved search and I have mentioned source from Email as per your suggestion.
image

2) As per below images I got a ticket which has been raised by customer and reference no on subject as order no .
image

3) As per below images I have copied order ID which I got in the subject line of the ticket then I searched the same ticket no.
When we open each ticket to search order ID (for checking the history of order ). It Showing different order no.
As I have observed its searching only first 4 characters as unique rest of character is uncommon. 
I have marked the ticket no in the screenshot.

image
I would like to request you whatever we search like character or number it should reflect the same not should be the type of.

Please feel free to reach us if you required more information.

Thanks
Mahendra Sahni

greezybacon:
@mahendrasahani this is not a forum, and your post has absolutely nothing to do with this topic. Please refrain from hijacking threads like this. And, as a general rule, when searching for things which include operators like `-`, you should quote your search term (eg. `"1577-4402-6187-5906"`)

ipAnything:
Hi,

Is this ready for a production environment or is not yet recommended?

greezybacon:
@ipAnything I would suggest upgrading a copy of you production database with the feature and test it out. At this point, I think you would be the best judge of its stability. Feedback is appreciated

ipAnything:
@greezybacon Is there a special upgrade procedure for this?

jdelhome3578:
@greezybacon The upgrader does NOT support upgrading from the current patch [98ad7d550c26ac44340350912296e673]! message. Do I need to do something different?

lewishackfath:
@greezybacon is there an expected release date for this feature?

ntozier:
@lewishackfath I dont think that he will answer you, but when 1.11 is released.

luiizfenix:
@greezybacon Hi, Upgrader error as @jdelhome3578 commented on Oct 27:

> @greezybacon The upgrader does NOT support upgrading from the current patch [98ad7d550c26ac44340350912296e673]! message. Do I need to do something different?

need to do something else? i'm on v1.10.1

Best Regards,

lewishackfath:
Hello,

I am currently using OSTicket 1.10 with a custom Bootstrap Theme for the Client Side of the Site.
Am I able to install this as well somehow?

thanks in advance.

rhew90:
hello, unfortunately, the search does not work.

error

ntozier:
@rhew90 I'm not sure what your reply to this thread has to do with this pull request (PR) to add customizable ticket queues.   This thread is not about search...

rhew90:
@ntozier I know, but it's a bug because if I install version 1.10 the problem does not exist.
In this version the search does not work so I can not use it.

ntozier:
@rhew90 So you are saying that: 
You installed this PR on 1.10 and search works.  
You installed this PR on 1.10.1 and search does not work.
This PR has never been updated for 1.10.1, and is intended for 1.10.
Did you try search in 1.10.1 prior to installing the PR?

rhew90:
I have not made any updates.

I have performed a clean installation with this. I have downloaded the complete Zip and in the information is written 1.8 Git. The search does not work.

So I deleted everything and installed the official stable version from site 1.10.1

ntozier:
Sorry, i have no idea what your saying...  If you used this PR then you updated 83 files... and this PR still has nothing to do with search.

rhew90:
schermata 2018-02-06 alle 18 19 00

I downloaded the zip package and installed. Stop

ntozier:
@rhew90 so we are back to this.

You are saying that:
You installed this PR on 1.10 and search works.
You installed this PR on 1.10.1 and search does not work.
Did you try search in 1.10.1 prior to installing the PR?

rhew90:
leave it

ntozier:
Is there some reason why you cannot simply answer the question?

rhew90:
I downloaded the zip package and installed. Stop
Version 1.8-Git and search not function.

schermata 2018-02-06 alle 18 40 40
 
Now I have installed 1.10.1 from here http://osticket.com/download 
without Customizable ticket queues. 

I do not want to make the change anymore because for me it does not work.

ok?

ntozier:
Q: Did you try search in 1.10.1 prior to installing the PR?
Your Answer: no.

Seaev:
Hello!
Do you know which files i have to integrate on OSTicket 1.10.1 to have your feature ?

Thanks a lot!

olly3gs:
Hello!

 

I'm afraid you didn't ask the right person.

 

I didn't develop any osTicket feature.

 

Regards,

 

Robert

greezybacon:
@rhew90 sorry for the long silence. I changed my system to Italian and attempted to recreate your crash and I cannot. Are you still able to available to comment on the issue?

mikshepard:
@greezybacon :  I'm running into an issue, where I can't edit teams.  Going into the Admin Panel --> Agents -->  "Teams" tab and trying to select an existing team puts a grey overlay over the page.  Here is the error log:

> [Thu May 10 15:22:13.456973 2018] [:error] [pid 24191] [client X.X.X.X] PHP Warning:  Invalid argument supplied for foreach() in /var/www/osticket/include/staff/team.inc.php on line 235, referer: https://ticket.mysite.org/scp/teams.php

I'm using a direct pull of your repo for this.  Running Debian 8, with php 5.6.33, mysql 5.7.22

Are you able to recreate?  Any idea of the issue?

Many thanks on this PR, it's been working great otherwise

ntozier:
Download the current version from osticket.com/download.

lewishackfath:
@ntozier I just did that and get the following error message on the updater page:
The upgrader does NOT support upgrading from the current patch [ad9d0a5f1bd9f83a4ecb3f590b334eac]!

ntozier:
Then highly unlikely.  If it made changes to the database you would have to find out what they were, and reverse them to go backwards.

JediKev:
@lewishackfath 

Because your DB still has the upgraded tables and new `schema_signature` in the `ost_config` table. Please revert the DB to the backup in order to restore the database to its original state. (you should've backed up the db and files before making these changes)

If you don't have a backup (which is bad news bears) you should compare your DB to the DB of a new installation of your version and restore the tables and `schema_signature` to its original state.

Cheers.

lewishackfath:
I have a backup but it was some time ago and have had people using it since, don't want to lose any of the data/user accounts etc. Might just have to leave it as is.

JediKev:
@lewishackfath 

Actually if you look at the "Files" tab of this pull request you can see all of the `.patch.sql` and `.cleanup.sql` files that contain the changes that were made to the database. All you have to do is revert those changes and update the `schema_signature` with the old one from the `include/upgrader/streams/core.sig` file in the "Files" tab.

Cheers.

lewishackfath:
I just deleted the 4 Extra columns created, changed the Schema_signature and replaced the files with those of the latest build and its all working.

JediKev:
@lewishackfath 

Awesome Possum! Just to let you know, the full/stable version of this feature will be released in v1.11rc-1 (coming soon).

Cheers.

lewishackfath:
@JediKev Awesome can't wait another company I do work for really wants custom queues but wanted to wait until it was in the official release.

MatD73772:
so sorry for being a bit slow. I'm trying to do a test of this mod as it's really important that we can view other fields in a ticket list.

So I'm running Apache And MySQL off my laptop using XAMPP. I've downloaded the zip linked at the top of this page, extracted as per usual and run the setup (no other osticket version installed). 

If I've installed PHP 5.6, then when I look at settings/tickets, I don't see any tabs
If I've installed PHP  7.0.30, then when I run the setup, I get Fatal Error: Cannot declare class Error, because the name is already in use in c:\xampp\htdcos\osticket\include\class.error.php on line 43

If I install 1.10.1, that works fine on PHP 7.0.30, but obviously without the important ticket view options..
Am I doing something obviously wrong here ?

thanks

greezybacon:
Thanks for all the comments thus far! This is being released in v1.11 and further issues should be discussed in #4342

Rockwebsupport:
Is there a way to allow "users" to input a "due date" and have it populate in the ticket queue at the /scp/tickets.php page?

JediKev:
@Rockwebsupport

This is a question best suited for our community supported Forum [here](https://forum.osticket.com). There you can search existing discussions for the same question, if this question hasn’t been asked yet (which it has), create your own discussion.

To answer your question though, you’d create a custom field, make it viewable/editable to Users, and then create a new column for queues that has this custom field as the Primary Data Source.

Cheers.

Issue #2557, opened 2015-09-16
v1.10-RC2: Avatar rendering in Safari 5.1.10


ehk20:
Avatars are not rendering correctly in V1.10-RC2 on Safari 5.1.10, as per the attached image (wrong size, vertically stretched, messing up the chat timelines), on this and other pages. While this is an older version of Safari, it is in use for 10.6 clients, which we have a few users on older Macs that can't upgrade beyond it, and while staff aren't too bothered, users find it confusing.
![screen shot 2015-09-16 at 09 26 18](https://cloud.githubusercontent.com/assets/14161662/9900052/e5cae85a-5c55-11e5-8326-420ccd7760f5.png)

nfebuary:
@ehk20 check to see if b4e1e9c fixes your issue. I don't have a way to test Safari 5.1 Sauce labs only has Safari 6 to test.

ehk20:
Fix confirmed as working for Safari 5.1. Thanks!

ehk20:
Leaving open as I'm not sure what the etiquette is for closing issues!

nfebuary:
You can close it since it works.

PR #2580, opened 2015-09-22
Fix for issue #2557

nfebuary:
#2557 Avatar height issue in Safari.


ehk20:
This is a problem for end users too, so needs the same fix in assets/default/css/theme.css

nfebuary:
ok, I'll make the fix.

nfebuary:
Try this #2609 for the end user avatars.

PR #2585, opened 2015-09-27
orm: Add MySQLi executor without prepared statements

greezybacon:
There are [plenty of reasons](http://joshduff.com/#!/post/2011-05-10-why-you-should-not-be-using-mysqli-prepare.md) not to use prepared statements. I originally chose to because of increased security, because better performance on some systems (e.g. Oracle), and because the prepared statement API is able to fetch correct types (ints, not strings).

However some tests show that as much as 30% of the overall query execution time for osTicket is preparing prepared statements. In theory this 30% could be avoided on the server by avoiding prepared statements. There is also a significant overhead in PHP to adapt the `bind()` and `fetch()` methods of the statement handling.

This patch adds and uses by default a simplified executor which uses the mysqli_query() function to process queries. This method is (alledgedly) faster on MySQL as it doesn't require the PREPARE overhead, nor require two trips to the database per query. All parameters are escaped and placed directly into the SQL statement.

This executor extends the traditional mysqli_fetch_xxx() methods by casting non-null numeric field values to PHP numeric types.

### Outstanding
  * [ ] Demonstrate through Xdebug profiling and database performance analysis that this method is indeed faster overall than the current executor.

### DIFF
--- a/include/class.orm.php
+++ b/include/class.orm.php
@@ -2699,7 +2699,7 @@ class MySqlCompiler extends SqlCompiler {
     }
 }
 
-class MySqlExecutor {
+class MySqlPreparedExecutor {
 
     var $stmt;
     var $fields = array();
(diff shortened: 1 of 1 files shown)


greezybacon:
It looks like MySQL 5.7 [reuses the output of the parse and resolution](http://mysqlserverteam.com/re-factoring-some-internals-of-prepared-statements-in-5-7/) phases of processing, so some performance could be gained using prepared statements if the statement were held and reused in PHP. However, this probably won't be beneficial as the connection is established for every request which would mean prepared statements could only benefit from reuse in the same request, which isn't much. Furthermore, this would require some sort of hashing and caching system in PHP to find an existing prepared statement rather than creating a new one at the database.

infectormp:
in my opinion better look to Mariadb futures because most distros switch to this db. Mariadb backport some futures from MySQL.

include/class.orm.php
```diff
@@ -2868,6 +2868,89 @@ function($m) use ($self) {
...
+ * are escaped and placed directly into the SQL statement. With this style,
+ * it is possible that multiple parameters could compile a statement which
+ * exceeds the MySQL max_allowed_packet setting.
+ */
+class MySqlExecutor
+extends MySqlPreparedExecutor {
+    function execute() {
+        if (!($this->stmt = db_query($this->__toString())))
+            throw new InconsistentModelException(
+                'Unable to prepare query: '.db_error().' '.$sql);
```
greezybacon: `$sql` is undefined

greezybacon:
> in my opinion better look to Mariadb futures because most distros switch to this db. Mariadb backport some futures from MySQL.

Watch out. MariaDB 10.1 will include new MySql 5.7 features

infectormp:
I alredy use MariaDB 10.1

greezybacon:
This was merged in another pull request

PR #2586, opened 2015-09-27
Upgrade pear Mail libraries, use header charset for attachment filenames

greezybacon:
This includes upgrades to the Mail_Mime and Net_SMTP upstream libraries. It also includes a patch to use the defined header charset (`utf-8`) for attachment filename encoding

### DIFF
--- a/include/pear/BUNDLE
+++ b/include/pear/BUNDLE
@@ -1,5 +1,11 @@
 Log of pear packages bundled with osTicket
 
+* osTicket v1.9.13, v1.10 *
+===========================================================
+Mail_Mime-1.10.0
+Mail_mimeDecode-1.5.5
+Net_SMTP-1.7.1
+
 * osTicket v1.7.*
 ===========================================================
  PEAR-1.9.4 - core class
--- a/include/pear/Mail/mime.php
+++ b/include/pear/Mail/mime.php
@@ -679,6 +679,9 @@ class Mail_mime
         if (!empty($value['headers_charset'])) {
             $params['headers_charset'] = $value['headers_charset'];
         }
+        elseif (isset($this->build_params['head_charset'])) {
+            $params['headers_charset'] = $this->build_params['head_charset'];
+        }
         if (!empty($value['language'])) {
             $params['language'] = $value['language'];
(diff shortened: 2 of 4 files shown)


matlinuxer2:
Thanks! I have applied this patch and tested it. All goes well, GJ!!

infectormp:
Need bump osTicket PHP version to 5.4 because Net_SMTP v1.7.1 have Dependencies PHP Version: PHP 5.4.0 or newer.

or maybe better bump version to 5.5 or 5.6 because 5.4 not supported anymore
for 5.5 active support end at 10 July 2015, security support 10 July 2016
for 5.6 active support end at 28 Aug 2016, security support 28 Aug 2017

greezybacon:
I agree that PHP 5.4 is a fair requirement for v1.10

Chefkeks:
@greezybacon
When bumping the minimum requirement anyway, why not skip 5.4 and go to 5.5 or 5.6 in favor of the active/security support?

greezybacon:
CentOS / RedHat 7 ship with PHP 5.4, and are still supported. Since that's the most current stable version of the popular server OS, I think we should support it. What versions are available with Windows Web Platform Installer?

ntozier:
Web Platform Installer 5.0 lists the following options for PHP
5.6.0
5.5.11
5.4.24
5.3.28

Chefkeks:
Ok makes sense to not skip PHP 5.4 since some distributions are still shipping with it as the default which I wasn't aware of. And as long as osTicket doesn't require any PHP 5.5 or 5.6 exclusive features/fixes I agree that 5.4 is a fair requirement.

greezybacon:
As a side note, I would point out that we will be labelling a **requirement**. I think we should move our development testing and certification to PHP 5.6 and label that version as the *recommendation*.

Chefkeks:
@greezybacon  :+1:

greezybacon:
@protich was merging this a mistake? Doesn't it require PHP 5.4 for osTicket v1.9.x? Perhaps we should revert the changes to the PEAR libraries on the `develop` branch

PR #2588, opened 2015-09-29
search: Try harder to match MySQL full text boolean query

greezybacon:
As currently written, if a full text query is attempted in BOOLEAN mode and the query cannot be compiled by MySQL, it causes an OrmException and a crash. This patch adds a more straight-forward and comprehensive grammer-based approach to checking for valid boolean syntax.


greezybacon:
Changing WORD to `[\w][\w-]+` seems slightly faster for strings that do not match the overall regex, and generally searching for one letter words is futile

PR #2590, opened 2015-09-29
Add option to use dept/agent name as FROM-NAME on replies

protich:
Enable the ability to optionally use/show the agent's full name, department's name or emails address name on responses/replies. Email address of the response would still be designated outgoing email address of the Department to which the ticket is assigned or the system default outgoing email.

* Agents can set the default FROM Name in Profile > Preference.
* Private department's name won't be used regardless of the setting.
* Agent's name won't be used if identity masking is enabled.

### Pending 
* [x] Move ``Agent Identity Masking`` from Settings > Tickets to Settings > Agents 
* [ ] Add DB patch to add ``default_from_name`` field or convert agent to use config table.

### DIFF
--- a/include/class.mailer.php
+++ b/include/class.mailer.php
@@ -66,10 +66,14 @@ class Mailer {
         $this->ht['from'] = $from;
     }
 
-    function getFromAddress() {
+    function getFromAddress($options=array()) {
 
-        if(!$this->ht['from'] && ($email=$this->getEmail()))
-            $this->ht['from'] =sprintf('"%s" <%s>', ($email->getName()?$email->getName():$email->getEmail()), $email->getEmail());
+        if (!$this->ht['from'] && ($email=$this->getEmail())) {
+            if (($name = $options['from_name'] ?: $email->getName()))
+                $this->ht['from'] =sprintf('"%s" <%s>', $name, $email->getEmail());
+            else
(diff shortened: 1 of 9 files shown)


include/class.mailer.php
```diff
@@ -66,10 +66,12 @@ function setFromAddress($from) {
...
     }
 
-    function getFromAddress() {
+    function getFromAddress($options=array()) {
 
-        if(!$this->ht['from'] && ($email=$this->getEmail()))
-            $this->ht['from'] =sprintf('"%s" <%s>', ($email->getName()?$email->getName():$email->getEmail()), $email->getEmail());
+        if (!$this->ht['from'] && ($email=$this->getEmail())) {
+            $name = $options['from_name'] ?: ($email->getName() ?: $email->getEmail());
+            $this->ht['from'] =sprintf('"%s" <%s>', $name, $email->getEmail());
```
greezybacon: I don't think `$name` here should fallback to the email address, although I see it was previously written that way. Perhaps
```php
    if ($name = $options['from_name'] ?: $email->getName())
        $this->ht['from'] = sprintf('"%s" <%s>', $name, $email->getEmail());
    else
        sprintf('<%s>', $email->getEmail());

protich:
@greezybacon -- made the change you suggested.  Thanks!

protich:
This PR is now part of #2700

PR #2600, opened 2015-09-30
users: Fix remote user import

greezybacon:


include/ajax.users.php
```diff
@@ -61,13 +65,16 @@ function search($type = null) {
                 return $this->search($type);
             }
 
+            // Omit already-imported remote users
             if ($emails = array_filter($emails)) {
-                $users->chain(User::objects()->filter(array(
```
greezybacon: This line crashes

protich: @greezybacon - Is the crash ORM related?

greezybacon: `chain` method doesn't exist. It's called `union`. That's what the patch changes

PR #2609, opened 2015-10-01
End user avatar fix for #2557 #2580

nfebuary:
Older browser height fix for the end user avatar. Changing the height
100% property to height auto; corrects the avatar size on older
browsers, in this case Safari 5. The fix will keep the image from
stretching 100% for the end user.


ntozier:
See you got this commit description thing down now. :)

nfebuary:
;) I'm doing my best!

PR #2612, opened 2015-10-02
login: Client SSO Destination

jdelhome3578:
Consider the destination clicked prior to SSO authentication.


protich:
@jdelhome3578  -- thanks again! PR #2913 merged.

PR #2616, opened 2015-10-02
Uploaded image preview fix

nfebuary:
Sets a static height to the row since it truncates it shouldn’t have
any odd behavior. Content in the span is centered on the row and sets
image height to fit inside the row.

Before:
screen shot 2015-10-02 at 4 28 22 pm

After:
screen shot 2015-10-02 at 4 14 59 pm


greezybacon:
:+1:

ntozier:
Looks good.

PR #2617, opened 2015-10-03
locks: Exclude expired locks via constrain

protich:
Fixes a bug where a lock icon is shown on tickets listing even when the lock in question is expired.


greezybacon:
Your profile pick is cracking me up

Chefkeks:
@greezybacon
Thought the same some hours ago :D :D

protich:
It was taking in the aroma of my first cup of coffee after 15 days of no-coffee!

PR #2622, opened 2015-10-05
Add location for ticket response

nfebuary:
Adds an additional location for the ticket response above the ticket
thread for a reverse order view of the ticket thread. Additionally the
reply/internal note button would be selected and the page would auto
scroll to the bottom.

To do: Add backend code.

screen shot 2015-10-05 at 9 47 48 am

screen shot 2015-10-05 at 9 46 40 am

### DIFF
--- a/include/staff/templates/status-options.tmpl.php
+++ b/include/staff/templates/status-options.tmpl.php
@@ -33,13 +33,12 @@ if (!$nextStatuses)
 ?>
 
 <span
-    class="action-button"
-    data-dropdown="#action-dropdown-statuses">
+    class="action-button pull-right"
+    data-dropdown="#action-dropdown-statuses" data-placement="bottom" data-toggle="tooltip" title="<?php echo __('Change Status'); ?>">
     <i class="icon-caret-down pull-right"></i>
     <a class="tickets-action"
         href="#statuses"><i
-        class="icon-flag"></i> <?php
-        echo __('Change Status'); ?></a>
+        class="icon-flag"></i></a>
 </span>
 <div id="action-dropdown-statuses"
     class="action-dropdown anchor-right">
--- a/include/staff/ticket-view.inc.php
+++ b/include/staff/ticket-view.inc.php
@@ -190,12 +190,8 @@ if($ticket->isOverdue())
             <?php             
             // Status change options
             echo TicketStatus::status_options(); ?>
-            <span class="action-button ticket-response pull-right"data-placement="bottom" data-toggle="tooltip" title="Internal Note">
-                <a href="#reply"><i class="icon-file-text"></i></a>
-            </span>
-            <span class="action-button ticket-response pull-right" data-placement="bottom" data-toggle="tooltip" title="Post Reply">
-                <a href="#note"><i class="icon-mail-reply"></i></a>
-            </span>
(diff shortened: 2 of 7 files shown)


protich:
I like the idea of it, but I don't like the location. 

I personally think we should consider changing the main action bar to be icon-based buttons, with took-tip, instead of the worded action-buttons we have now. This will give us enough room to add more options and solve inconsistency issues related to multi-lingual support.

nfebuary:
Agreed, If we make everything icons with tool tip we will have room in the sticky bar to have the reply and internal note too. Then it wont matter where you are on the page the reply button will always be accessible. I'll get on that.

Chefkeks:
> I like the idea of it, but I don't like the location.

Agree :+1: 

> I personally think we should consider changing the main action bar to be icon-based buttons, with took-tip, instead of the worded action-buttons we have now. This will give us enough room to add more options and solve inconsistency issues related to multi-lingual support.

Sounds also good to me, but in addition to the tool-tip it may would be nice to have a mouse-over effect. So when the mouse is placed over the icon, the button will grow and show the icon + text - that way I think it would be more obvious what action a button has (if people don't get what symbol/icon stands for what action).

nfebuary:
New commit changes the location of the Reply and Internal Note to be inline with the action buttons at the top of the ticket.

screen shot 2015-10-06 at 3 24 40 pm

Buttons are also added to the sticky header for constant use anywhere on the thread.

screen shot 2015-10-06 at 3 24 52 pm

Tooltip was z-indexed below the dropdown I made this small change to place it above. (Not sure why EVERYONE insists on 999999999 as a number but 10 or 15 or even 20 are perfectly acceptable z-index numbers).

screen shot 2015-10-06 at 3 25 37 pm

screen shot 2015-10-06 at 3 25 48 pm

Additionally I added the tooltip to hide when the respective button is clicked for ALL buttons that have a tooltip.

@Chefkeks I really like the idea of having the names open on hover I'll have to explore that for the next commit.

@protich TODO: change the tabs when the Reply or Internal Note is selected. Not sure why my other attempts did not work so I reverted to a basic href before commit, if you could take a look.

Admin side: a toggle to reverse order the threads.

nfebuary:
Latest Commit removes some unnecessary spans and adds a color to Internal Note button to differentiate between Reply and Internal Note.
screen shot 2015-10-07 at 1 48 29 pm

Chefkeks:
I think this is a eye catcher, but not a positive one. It puts too much focus on the internal note button in my opinion.

nfebuary:
Peter was looking for some way to show that there is a difference between the two, we were thinking a yellowish color. Maybe it needs to be more subtle, or simply a colored border instead.

Chefkeks:
Ok I understand now what you want to achieve and have to admit that I currently have not really an other suggestion (thought about it since my previous post), but I'm confident you all will find a good solution by trying the several possibilities ;)

Btw. this here remembers me a little bit about what Jared once taught me: Half the world will like it, half not :D ;)

Chefkeks:
Oops and I forgot to tell you that the subtle version is definitely better in my mind :+1:

nfebuary:
Haha! Jared taught me the same lesson. :D Agreed on the subtle version.

protich:
Pull requests #2651 implements pending items. I decided to only highlight the post internal note button on hover.

PR #2633, opened 2015-10-07
Provide the option to preselect a help topic

ajzanotti:
This pull request provides the option to preselect a help topic on ticket open by specifying the topic ID in the URL request string.

Example:
http://example.com/open.php?ht=5

This will fill the topicID field on open.php with the option that has a value of 'ht' or 5 in this case.

If a topic ID isn't specified the default action is to fill the field with the "Select a Help Topic" text.


ajzanotti:
You're absolutely correct. I've made your recommended code change.

greezybacon:
Merged #2735 instead. Thanks for the request

PR #2635, opened 2015-10-09
Colour rules in agent panel v.1.10 RC-2

Technospain:
I propose to developers create three color rules to tickets assigned in agent panel to see at a glance the priority of open cases. In my opinion this option are basic and very important for team work

Red - critical
Orange- medium
Green - low


ntozier:
I think that you meant for this to be an issue report.

Issue #2397, opened 2015-08-11
1.10 RC1 Issue - Login Page Object Moved


jdelhome3578:
![image](https://cloud.githubusercontent.com/assets/5774055/9201512/7dec6d7c-401b-11e5-9fff-0d345f7c0da7.png)

`Object Moved`
`This document may be found here`

Refreshing the page will login correctly.

LDAP Enabled
HTTP Pass Thru Enabled

ntozier:
I cannot replicate that.
Can you provide the page source?
Are you doing any web server redirect type things?

jdelhome3578:
All files are from the tip of develop-next. I have no redirection. I don't have this issue with 1.9.11

jdelhome3578:
Issue goes away when `// Browsers shouldn't suggest saving that username/password
Http::response(422);` is commented out

jdelhome3578:
Also if I disable Windows Auth on IIS issues goes away.

jdelhome3578:
@greezybacon still an issue in 1.10 rc2

jdelhome3578:
#2641 fixes this issue more directly.

PR #2641, opened 2015-10-10
Fixes bare header redirect

kevinoconnor7:
This fixes a bug where the browser will fail to redirect away from the login page if the client's browser doesn't allow for location mutations with the 422 response code. This was triggered since there was no exit call after the bare header mutation. Using the HTTP#redirect call solves this problem.

This issue only impacted SSO auth plugins.


jdelhome3578:
Fixes #2397

PR #2671, opened 2015-10-22
Fixes #2642

### LINKED ISSUE #2642: devnext: fatal @ getDbTimezone
Get `Fatal error: Call to a member function getDbTimezone() on null in /xxx/include/class.misc.php on line 99` @ install step with last devnext branch

mariadb 10.1.7

c0nfused:


greezybacon:
I think the best solution to this problem would be to convert the Config class to the ORM. Then the dbtime() call would not be necessary in the Config::update method. Second best I think is just to add a check in the Config::update method (around lind 127)

```php
global $cfg;
if (isset($cfg))
    // Config uses the global $cfg to determine the database timezone
    $setting['updated'] = Misc::dbtime();
```

protich:
I agree with Jared, but perhaps we should also not assume ``$cfg`` will always be set.

Diff of what we have in production... 

```diff
diff --git a/include/class.misc.php b/include/class.misc.php
index 69063c4..0af9257 100644
--- a/include/class.misc.php
+++ b/include/class.misc.php
@@ -85,19 +85,24 @@ class Misc {
         static $dbtz;
         global $cfg;
 
-        if (is_null($var) || !$var) {
-            // Default timezone is set to UTC
-            $time = time();
-        }
-        else { //user time to UTC
-            $tz = new DateTimeZone($cfg->getTimezone());
+        if (is_null($var) || !$var)
+            $time = time(); // Assume default timezone is set to UTC
+        else
             $time = is_int($var) ? $var : strtotime($var);
+
+        // Bail out if we don't have $cfg loaded e.g during install
+        if (!$cfg) return $time;
+
+
+        if ($var) { // user time to UTC
+            $tz = new DateTimeZone($cfg->getTimezone());
             $D = DateTime::createFromFormat('U', $time);
             $time -= $tz->getOffset($D);
         }
-        if (!isset($dbtz)) {
+
+        if (!isset($dbtz))
             $dbtz = new DateTimeZone($cfg->getDbTimezone());
-        }
+
         // UTC to db time
         $D = DateTime::createFromFormat('U', $time);
         return $time + $dbtz->getOffset($D);
```

c0nfused:
Yeah, that's the problem I was having during installation was that `$cfg` was not set.

My solution was more of a quick fix than anything, and I just took it overly cautious to make sure it wasn't just set but also an object.. not that it would or should ever be a string or anything, was just handling each possibility.

Will be attacking some more bugs!

codekmarv:
Hi guys, thank you for all your hard work. So was this fixed on the develop-next branch already?

Regards,

Marvin M

balojs:
same problem develop-next (11-17-2015)

greezybacon:
I think this was addressed in #2827

PR #2673, opened 2015-10-23
Feature/task enhancements

kest874:
Changes to the task tab including:

Added the "Ticket" column to the table to see the ticket the task was created from
Added the "Ticket" sort option
When a task is created in the tack view the "Ticket" is blank

image

image

Ideas from https://github.com/osTicket/osTicket-1.8/issues/2578 and https://github.com/osTicket/osTicket-1.8/issues/244


kest874:
Would like to add the count of tasks to the tickets grid (annotations) at some point as well

Chefkeks:
Interesting...
I don't remember exactly whether I wrote publicly or directly to the osTicket developers via mail about my "tasks/ticket relationship" ideas and suggestions but my idea goes basically in the same direction - just had a couple more stuff I'd have added too, but instead of writing about all that again here, I'm curious what the devs think about this pull request (especially since custom queues are on the way and would made this PR kinda unnecessary in my opinion).

Chefkeks:
For the record, the custom queues feature pull request: https://github.com/osTicket/osTicket-1.8/pull/2577

protich:
I strongly disagree with adding ticket number as a column -- adding it as suffix to the task number might be a better solution. Tasks is primarily designed as generic internal todo list that can be created independently or can be attached to other models such as tickets, organizations, users...etc.

PS: @Chefkeks - I have a pull request pending to address the relationship linkage.

kest874:
Updated the second screen capture as it was incorrect.

I agree with the custom queue route, one complaint with our use is that in the task view you have no idea what the parent is.  

I also like the idea of adding the ticket number as a suffix. 

Another solution might be to add a link to the parent in the preview of the task.

PR #2695, opened 2015-10-30
Add missing icons in Assigned To column

antriver:
There was an element next to the assigned agent/teams's name to display an icon, but those icons didn't exist. I created a user icon and set the team icon.


scp/css/scp.css
```diff
@@ -344,7 +344,7 @@ a.emailSettings { background:url(../images/icons/emails.png) }
 a.emailDiagnostic { background:url(../images/icons/email_diagnostic.gif) }
 a.banList { background:url(../images/icons/ban_list.gif) }
 
-a.users { background:url(../images/icons/list_users.gif) }
+a.users { background:url(../images/icons/user.gif) }
```
greezybacon: Looks like the `list_users.gif` graphic currently exists. Why the change?

antriver: The list_users.gif icon  shows a person with a "ticket" in front of them. That's appropriate when it's shown next to the "My Tickets" link [](http://ctrlv.in/669494), but the same icon was also used for the "Agents" and "Access" pages in the admin panel. Since I added that new icon (user.gif ) to the assigned to column, it seemed more appropriate to use that one when referring to agents as individual people. [](http://ctrlv.in/669495)

greezybacon:
Fair enough. I like the change

protich:
I was thinking it was a good thing to retire the assignment icons -- we already have plenty of icons with addition of thread count, collaborators ...etc. Just a thought...

PR #2696, opened 2015-10-30
Display message type on message

antriver:
This was more of an internal request than useful feature for every user. But I thought I'd see what you think anyway.

Displays what kind of entry each message is on the ticket thread:


greezybacon:
Seems like we should do an icon with a tooltip title

PR #2703, opened 2015-11-03
Thread preview

protich:
Implements PR #2559


kest874:
Thumbs Up

kest874:
On suggestion with some use, default all of them to collapsed, add an option for sort.  It's possible that agents might want to see the first thread first.

nfebuary:
Thread Sort can be a user setting in their profile preferences.

kest874:
Agree, profile setting would be best.

protich:
Thread entries sort order as an agent's preference implemented in #2700 -- commit 0f814d5

PR #2710, opened 2015-11-04
UI Utility navigation and profile block.

nfebuary:
UI Prototype changes the layout and function of the top right corner
navigation. Creates a profile panel with information related to the
Agent and a link to the Agents full profile. Also moves the Agent,
Admin, and Account navigation into a drop down with the possibility for
future expansion to the navigation. i.e. Report a bug that allows
admins to submit tickets directly to the osTicket Team. Or access help
information. Ideally admins could add their own custom navigation under
the primary links.

TO-DO:
1. needs Php, it’s only HTML and CSS right now.

Screen shots:
screen shot 2015-11-04 at 3 25 38 pm
screen shot 2015-11-04 at 3 25 10 pm
screen shot 2015-11-04 at 3 24 58 pm
screen shot 2015-11-04 at 3 25 18 pm
screen shot 2015-11-04 at 3 25 27 pm
screen shot 2015-11-04 at 3 25 55 pm

### DIFF
--- a/include/staff/header.inc.php
+++ b/include/staff/header.inc.php
@@ -58,21 +58,37 @@ if ($lang) {
         echo sprintf('<div id="notice_bar">%s</div>', $ost->getNotice());
     ?>
     <div id="header">
-        <p id="info" class="pull-right no-pjax"><?php echo sprintf(__('Welcome, %s.'), '<strong>'.$thisstaff->getFirstName().'</strong>'); ?>
+        
+        
+        
+        <ul id="info" class="pull-right no-pjax">
+            <li>
+                Welcome, <a class="no-pjax" data-dropdown="#profile-dropdown"><strong>John</strong> <i class="icon-sort-down"></i></a>
+            </li>
+            <li><a class="no-pjax" data-dropdown="#utility-dropdown"><i class="icon-th"></i></a></li>
+            <li><a class="no-pjax"><i class="icon-signout"></i></a></li>
+        </ul>       
+        
+        
+        <!--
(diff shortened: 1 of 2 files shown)


Chefkeks:
Looks awesome and I really like the idea in general and also the details like the "report a bug" option! :+1:

As usual I directly have some feedback :D
My suggestion is to remove the "Welcome, " before the users first name and instead add the "single user icon" which is used for Agent panel on the drop down. Saves a translation and would look better in my opinion.

Cheers,
Michael

nfebuary:
Actually that sounds like a plan I'll do that. It makes sense since the other options are icons.

Hannibal226:
Nice, handy feature!

ntozier:
Very very swank. +1

nfebuary:
Here is a screen with the user icon. If we do it this way it might be better to have the users first and last name. I added a screen to see what that would look like.

With only the icon being added:
screen shot 2015-11-05 at 1 32 48 pm

This has the full name with icon.
screen shot 2015-11-05 at 1 38 52 pm
screen shot 2015-11-05 at 1 36 19 pm

Chefkeks:
I agree the full name looks better, but don't think it needs a setting to define if the first, the full or may the last name should be shown. Most other services I use in my company or privately mostly show the username or the full name - so full name should be fine in my opinion.

kest874:
Have a future plan to make it reactive?  I've been working on the client side for awhile now.

nfebuary:
@kest874 I can't exactly say when it will happen(because I don't know) but yes there is discussion on this topic. A template based system that allows the admins to literally do whatever they want to the template would be ideal. Then... any template/theme could be applied and solve issues such as responsive layouts for client and agent side.

PR #2714, opened 2015-11-09
Make Authentication Tokens Optional

protich:
Add ability to disable use of authentication tokens on ticket links. 

__Note__: Disabling auth-tokens effectively kills Guest Ticket Access email verification mechanism. Use of difference link variable might be necessary.

### DIFF
--- a/include/class.client.php
+++ b/include/class.client.php
@@ -41,12 +41,14 @@ implements EmailContact, ITicketUser, TemplateVariable {
         $tag =  substr($name, 3);
         switch (strtolower($tag)) {
             case 'ticket_link':
+                $qstr = array();
+                if ($cfg && $cfg->isAuthTokenEnabled()
+                        && ($ticket=$this->getTicket()))
+                    $qstr['auth'] = $ticket->getAuthToken($this);
+
                 return sprintf('%s/view.php?%s',
                         $cfg->getBaseUrl(),
-                        Http::build_query(
-                            array('auth' => $this->getTicket()->getAuthToken($this)),
-                            false
-                            )
+                        Http::build_query($qstr, false)
                         );
                 break;
         }
@@ -64,7 +66,7 @@ implements EmailContact, ITicketUser, TemplateVariable {
         return array(
             'email' => __('Email address'),
             'name' => array('class' => 'PersonsName', 'desc' => __('Full name')),
-            'ticket_link' => __('Auth. token used for auto-login'),
+            'ticket_link' => __('Link to view the ticket'),
         );
     }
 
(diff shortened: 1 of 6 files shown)


include/class.client.php
```diff
@@ -41,12 +41,14 @@ function __call($name, $args) {
         $tag =  substr($name, 3);
         switch (strtolower($tag)) {
             case 'ticket_link':
+                $qstr = array();
+                if ($cfg && $cfg->isAuthTokenEnabled()
+                        && ($ticket=$this->getTicket()))
+                    $qstr['auth'] = $ticket->getAuthToken($this);
+
```
greezybacon: I think this can be moved to `::getVar($tag)`, which would simplify the code and perhaps remove the dependency on the __call trampoline

greezybacon:
```patch
diff --git a/include/class.client.php b/include/class.client.php
index 23b64d1..932a5bd 100644
--- a/include/class.client.php
+++ b/include/class.client.php
@@ -36,25 +36,7 @@ implements EmailContact, ITicketUser, TemplateVariable {
                 ? call_user_func_array(array($this->user, $name), $args)
                 : call_user_func(array($this->user, $name));
 
-        if ($rv) return $rv;
-
-        $tag =  substr($name, 3);
-        switch (strtolower($tag)) {
-            case 'ticket_link':
-                $qstr = array();
-                if ($cfg && $cfg->isAuthTokenEnabled()
-                        && ($ticket=$this->getTicket()))
-                    $qstr['auth'] = $ticket->getAuthToken($this);
-
-                return sprintf('%s/view.php?%s',
-                        $cfg->getBaseUrl(),
-                        Http::build_query($qstr, false)
-                        );
-                break;
-        }
-
-        return false;
-
+        return $rv ?: false;
     }
 
     // Required for Internationalization::getCurrentLanguage() in templates
@@ -70,6 +52,22 @@ implements EmailContact, ITicketUser, TemplateVariable {
         );
     }
 
+    function getVar($tag) {
+        switch (strtolower($tag)) {
+        case 'ticket_link':
+            $qstr = array();
+            if ($cfg && $cfg->isAuthTokenEnabled()
+                    && ($ticket=$this->getTicket()))
+                $qstr['auth'] = $ticket->getAuthToken($this);
+
+            return sprintf('%s/view.php?%s',
+                    $cfg->getBaseUrl(),
+                    Http::build_query($qstr, false)
+                    );
+            break;
+        }
+    }
+
     function getId() { return ($this->user) ? $this->user->getId() : null; }
     function getEmail() { return ($this->user) ? $this->user->getEmail() : null; }
 
```
Here's what I meant

PR #2725, opened 2015-11-12
Feature/thread event tagging

kest874:
Different approach to https://github.com/osTicket/osTicket-1.8/pull/2696  (develop-next)

image


greezybacon:
Can we add spacing between the labels?
```css
.label + .label {
    margin-left: 1em;
}
```

kest874:
Done:

image

ntozier:
The spacing between the Collab button and the Message button does not appear to be the same amount of space between the Message Button and drop down arrow box.

Chefkeks:
I agree with @ntozier. But in general that PR looks pretty good :)

kest874:
Fixed the alignment all around:

image

PR #2726, opened 2015-11-12
Issue/2685

kest874:
This resolves issue https://github.com/osTicket/osTicket-1.8/issues/2685

Emails will be processed from oldest to newest.


greezybacon:
This original code seems to work properly on some IMAP system, and incorrectly on others. Is there a property from the server than can be polled?

kest874:
I'll have to look into this, the IMAP were using here is exchange 2013.

mfelber:
We are using dovecot. With this patch emails will be processed from oldest to newest. With the original code not. Maybe it helps to have more information which server type is affected and which not.

mfelber:
It seems that the uid of the mails will be used for the fetch order.

If I send two emails to my inbox the first mail (older one) gets the uid 1 and the second the uid 2.

With the original code the second mail (newer one) will be fetched first.

With the code from the patch the first mail will be fetched first.

I can manipulate the uid of the mails if I move for example the first mail to another folder and move it back to the inbox.

Then the second mail has uid 1 and the first mail has uid 2.

The original code will then work as expected.

I guess there is no way using the uid to ensure that the order will ever be as expected.

But it looks like that the code from the patch (with raising the increment) is fitting better to most imap setups.

As far as I know normally the uids are assigned in ascending fashion (higher the value newer the message).

kest874:
Was thinking maybe the there should be an option in the email setup that is a simple checkbox "Reverse Email Fetch" that can be toggled, this way it can be toggled if your not getting the result you expect.  Thoughts?

bunkobugsy:
Great! Can confirm that with dovecot forward for clause works as expected (oldest first, FIFO).

greezybacon:
@protich confirms that email is fetched in reverse, because if the emails are not being deleted or moved into a folder, then no new emails would be fetched after some point in time, because the deleted emails at the end of the inbox would be fetched, detected as already fetched, and marked as error. Then, the fetching system would have 100% error rate after fetching its 30 or so max emails (configurable), and would break down.

This needs to detect if deleting or archiving is enabled before swapping the fetch order

bunkobugsy:
clipboard01

And another thing: ever since I reversed  mail read order (oldest first), tickets are shown as in attahced image, even tough ticket id-s are ok, ascending. Probably another for clause needs to be reversed.

kest874:
I'm currently moving the fetched emails to another folder.  I'll play around with it and see what I can come up with taking into account the archive or delete function.  

@protich did you confirm against Exchange?

PR #2735, opened 2015-11-16
Patch so that topicId can be specified in $_GET

flotwig:
Allows topicId to be specified as a URL parameter when opening a ticket.

### DIFF
--- a/include/client/open.inc.php
+++ b/include/client/open.inc.php
@@ -9,6 +9,9 @@ if($thisclient && $thisclient->isValid()) {
 
 $info=($_POST && $errors)?Format::htmlchars($_POST):$info;
 
+if(array_key_exists('topicId',$_GET) && preg_match('/^\d+$/',$_GET['topicId']))
+    $info['topicId'] = intval($_GET['topicId']);
+
 $form = null;
 if (!$info['topicId'])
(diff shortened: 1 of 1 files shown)


include/client/open.inc.php
```diff
@@ -9,6 +9,9 @@
 
 $info=($_POST && $errors)?Format::htmlchars($_POST):$info;
 
+if(array_key_exists('topicId',$_GET) && preg_match('/^\d+$/',$_GET['topicId']))
+    $info['topicId'] = intval($_GET['topicId']);
+
```
greezybacon: I think you should move this code in the block below:
```php
if (!$info['topicId']) {
    if (array_key_exists('topicId',$_GET) && preg_match('/^\d+$/',$_GET['topicId']))
        $info['topicId'] = intval($_GET['topicId']);
    else
        $info['topicId'] = $cfg->getDefaultTopicId();
}
```
You could also perform a stronger check to ensure valid topic id:
```php
if (array_key_exists('topicId',$_GET) && Topic::lookup($_GET['topicId']))
```

flotwig:
Good points. Made those changes.

greezybacon:
@flotwig, @ajzanotti between this and #2633, which implementation is better? Can we consolidate the ideas into one?

PR #2749, opened 2015-11-23
Feature/custom queue dropdown

nfebuary:
Changes to custom queue frontend dropdown and it's functionality. Removed frontend queue manager to make way for a better sub queue ui. Sub queues are now right drop down menus instead of the file tree list style. This will provide a better way to group queue types within the main queue. Additionally IE fixes have been made and tested in sauce labs IE9 Windows 7. Design and some layout changes have been made to accommodate IE's rendering. :/ Other minor design changes have been made. Screen shots below.

screen shot 2015-11-23 at 3 57 32 pm
screen shot 2015-11-23 at 3 57 48 pm

Windows7 - IE9 Test

screen shot 2015-11-23 at 3 56 48 pm
screen shot 2015-11-23 at 3 57 02 pm


ntozier:
:+1: 
This looks great!
My only concern (which isn't a huge concern) is that with all that padding (examples: between Open and test, Dashboard and Users, Testing This Queue and Unanswered, etc.) makes me feel like a lot of valuable on screen real estate is being wasted with blank space aka spacing/padding.  I don't want it to feel crowded either, but I think that you could shave a few pixels (top/bottom) [and 10-20 left/right] and it would still be just as aesthetically pleasing.  Something to think about I guess. :)

nfebuary:
@ntozier We have icons that will fill that space, we are discussing options for admins to select the icon they want. I left the space for the icon. @protich has a pull request with temp icons in place for now as well as a condition for the break between admin queues and personal queues. I could shave some margin off though.

greezybacon:
It looks like you dropped the padding around the popout edit/delete box. I liked it. Perhaps we should have just made it smaller?

greezybacon:
Should we add the top-level queue as the first item in the drop down (perhaps with light-grey background or something to set it apart from the other items in the list)? Such would allow EDIT for administrators.

greezybacon:
Should we allow personal queues at any level?
screen shot 2015-12-17 at 10 18 47 pm
*Yeah, that's right, I got the queue counts working*

greezybacon:
Another question: where should the top-level queue counts go?

Chefkeks:
> Should we add the top-level queue as the first item in the drop down (perhaps with light-grey background or something to set it apart from the other items in the list)? Such would allow EDIT for administrators.

Yes! Please! :+1: 
(Btw. already suggested that a while ago in one of my feedback posts here on github)


> Should we allow personal queues at any level?
Makes sense to me and would be a neat improvement.

nfebuary:
> It looks like you dropped the padding around the popout edit/delete box. I liked it. Perhaps we should have just made it smaller?

If I remember correctly I had to remove it. The padding was conflicting with something in the dropdown menu after I converted the system to use drop downs in favor of a menu tree. I can revisit this since I have a clear head on it.

> Should we allow personal queues at any level?

Yes, I agree with adding personal queues to any level, the dropdown system makes that more feasible. But would it make more sense to have a select list on the Add Personal Queue modal that gives you a list of all the queues you could use as a parent? I think that would eliminate the need to have "Add Personal Queue" on every level of the dropdown system. Agents would have "Add Personal Queue" twice or more on a page depending how deep they go with the dropdown menus. I think it needs to be once on the main parent dropdown and have the choice of all the queues to choose from even possibly their own personal queues in the Add Personal Queue modal. 

I would have to extend the dropdown system to include personal queues, but that should be easy enough.

> Should we add the top-level queue as the first item in the drop down (perhaps with light-grey background or something to set it apart from the other items in the list)? Such would allow EDIT for administrators.

Yes, I agree. However, I'm not sure about setting it apart though. If it looks like the other buttons I feel people will be confident in knowing it's a button and not possibly a header, or random duplicated text, ect... Sure enough you can click on it and it will go where it is supposed to go but the agents need consistency and confidence in the UI. I think it should be the same as the other buttons.

> Another question: where should the top-level queue counts go?

The tabbed main element is just a container so I think the duplicated main level queue should have the count.

We removed the gear from the admins custom queues.
screen shot 2015-12-18 at 10 07 35 am
We chose to remove the gear on admin queues since they can go to the admin side and make the changes they need. BUT that throws the counts out of alignment because the personal queues still have the gear. I have an idea to accommodate this. TO THE BAT CAVE!!!

greezybacon:
> We chose to remove the gear on admin queues since they can go to the admin side and make the changes they need

Sure, after 4 more clicks... I'm not sure I like it better without the gears.

greezybacon:
> The tabbed main element is just a container so I think the duplicated main level queue should have the count.

I agree. That way we don't have to count the tickets for every page load. As implemented, it only counts the tickets when the queue navigation drop-downs are hovered.

nfebuary:
Touché the thought did cross my mind. Maybe I just need to figure out a better way of handling the editing of custom queues. I don't want the interface to be bloated with buttons everywhere but then I don't want the admin to click more than they need to get to something that could have just been a single click.

greezybacon:
I think we should just be consistent. If the gears are annoying, then drop them alltogether. Perhaps instead there could maybe just be a single gear on the queue-listing page after the queue "heading". (The heading one can click to refresh the page). Then there's only one button on the page for edits, deletes, etc.

greezybacon:
> But would it make more sense to have a select list on the Add Personal Queue modal that gives you a list of all the queues you could use as a parent? I think that would eliminate the need to have "Add Personal Queue" on every level of the dropdown system. Agents would have "Add Personal Queue" twice or more on a page depending how deep they go with the dropdown menus.

I agree. Less is more

nfebuary:
You mean like this:
screen shot 2015-12-18 at 3 57 58 pm

That is here: a2574bdc35700b36a1f8824b4ae056c8a92b0b37
(please disregard the 102 commits, sorry :/ )

Chefkeks:
I really like it that way with the gear behind the queue heading / name :+1:

greezybacon:
@nfebuary, yeah, I think that would allow for the best consistency and leanness of the UI. I'll pull that into my pull request

nfebuary:
Sounds good, now this is just HTML only. As for the nav elements, use developers discretion on what will be displayed. This was simply to show what it would look like.

greezybacon:
> As for the nav elements, use developers discretion on what will be displayed. This was simply to show what it would look like.

That's the relationship I would like to see; and visualization is the part I have difficulty with

greezybacon:
Looks like the step CSS made it into that commit. I'm going to strip that part out when I merge it. No action required of you; but I'd like to make sure we don't lose it

nfebuary:
Right, I forgot to stash that when I made the commit. It weaseled it's way in. It's basically a just incase you want to use it to turn it into a step I don't think I have anything actually using it at the moment. The modal I was going to use it for Peter wanted the tabs instead.

greezybacon:
Gotcha. Here's the queue system with arbitrary nesting, no more cogs per item, and repeating the top-level queue in the first-level drop-down with the count.

screen shot 2015-12-21 at 5 32 03 pm

I still feel like the top-level item needs some distinction from the *sub* queues of itself. I think the light gray bar like what we have on "Recent Searches" would be appropriate.

greezybacon:
Should we make the cog beside the queue title show only on hover?

greezybacon:
For the records, here it is with the light  background

screen shot 2015-12-21 at 5 42 48 pm

nfebuary:
I dig it!

nfebuary:
> Should we make the cog beside the queue title show only on hover?

Maybe lighter but not invisible. I don't think people will know to mouse over the title without some indication that there are options.

greezybacon:
More UI feedback:

Maybe it will look different with the icon in the navigation, but I don't particularly like the border around the top-level queue navigation (the "Open" queue in the nav bar in the picture). Perhaps the background and text color change is enough?

screen shot 2015-12-22 at 2 17 53 pm

I also think the sub-level indication (beside the words "My Tickets") should be an arrow to the right; not to the bottom.

nfebuary:
> Maybe it will look different with the icon in the navigation, but I don't particularly like the border around the top-level queue navigation (the "Open" queue in the nav bar in the picture). Perhaps the background and text color change is enough?

It looks different with the icons. I think the border helps distinguish what's being selected and starts the dropdown off by connecting the top nav to the dropdown. I think it might look disconnected without it. Now that I said that it might look better with less padding, OR maybe take off the border-radius and just have it boxed. 

>I also think the sub-level indication (beside the words "My Tickets") should be an arrow to the right; not to the bottom.

Arrow pointing right is a better indicator of what is going on in the menu however this was discussed a while back and the down caret was chosen to indicate that there were more options. BUT the menu has changed so much since we started this that I think we should do icon-caret-right. Go for it! :smile:

ipAnything:
Hi has this been abandoned? Is this compatible with v1.10rc3?

Any plans to include it in v1.10?

ntozier:
It is mu understanding that this feature was pulled from the 1.10 release and was planned to be included in the version after that.

PR #2755, opened 2015-11-26
Dirty hack for redirect header after output. Fixes #2667

### LINKED ISSUE #2667: Develop-Next: Cannot search by ticket number in scp
When I try to search by ticket number I get this error:

image

I am on the latest develop-next commit 1aaa3139d1dfd7236c576b1278261bc820690a1d.

Message reads:
`Warning: Cannot modify header information - headers already sent by (output started at osTicket-1.8\include\staff\templates\sub-navigation.tmpl.php:27) in osTicket-1.8\include\class.http.php on line 62`

t-oster:
As described in #2667 the problem is, that the header include contains HTML, which is outputted before the HTTP-Redirect method is called. The redirect method sends a redirect header, which cannot be sent anymore, because there was already HTML outputted. This is an indicator of a bigger structural problem, since the logic (e.g. redirect) should have been executed before the HTML is generated.

However since this would require major changes, a quick and dirty solution is, to buffer the output instead of sending it directly. This way the headers can be modified while on the way.

### DIFF
--- a/css/redactor.css
+++ b/css/redactor.css
@@ -153,6 +153,7 @@ body .redactor-box-fullscreen {
   background: #fff;
   border: none;
   box-shadow: 0 1px 4px -2px rgba(0, 0, 0, 0.4);
+  z-index: 1 !important;
 }
 .redactor-toolbar:after {
   content: "";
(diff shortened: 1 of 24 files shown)


greezybacon:
I suggested #2830 to address this directly by bypassing the search page

PR #2760, opened 2015-11-30
Task Revisited

protich:
This PR address various task related issues...

1) Make tasks actions icon based.
2) Fix tasks counts on refresh
3) Tasks assignment bugs


t-oster:
I use it in my customized version and it seems to work.

PR #2761, opened 2015-11-30
Columns Heading Sort

protich:
Column heading based sorting got removed in favor of drop down based sorting in commit
number fa327c. It's being returned due to negative feedback we received from users.


greezybacon:
Looks like a bit of pork. Should the tasks actions be split out separately?

greezybacon:
Actually, it looks like it includes the commits from #2760. It should be rebased to only include 60979ee

PR #2762, opened 2015-11-30
cli: Agents Access and Permission CLI importer/exporter

protich:
Initial concept for agents access and permissions importer -- it's coded to take in file in YAML format for now, but the plan is to allow CSV import, via admin panel, with yet to be defined format.

## Pending 
* [ ] Allow CSV import via admin panel
* [ ] Finish exporter -- csv output

### DIFF
--- a/include/class.staff.php
+++ b/include/class.staff.php
@@ -1463,3 +1463,68 @@ extends AbstractForm {
         return $clean;
     }
 }
+
+class StaffAccess {
+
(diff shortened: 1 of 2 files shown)


greezybacon:
I would argue this CLI app should go as a verb under the `agent` module:

```sh
$ php manage.php agent set-access -f /path/to/file
...
$ php manage.php agent dump-access > /path/to/file
```

PR #2767, opened 2015-12-02
Ticket Status Properties

protich:
Allow changing/saving properties on internal ticket statuses, with the exception of the state.


greezybacon:
This is extremely abstract. Could you snap a couple screenshots?

PR #2768, opened 2015-12-02
Check agent's access on reopen auto-assignment

protich:
Auto-assign to closing staff or the last respondent if the agent is available and has access. Otherwise, put the ticket back to unassigned pool.


include/class.ticket.php
```diff
@@ -1525,28 +1525,21 @@ function onMessage($message, $autorespond=true) {
         // confused with autorespond on new message setting
         if ($autorespond && $this->isClosed() && $this->isReopenable()) {
             $this->reopen();
+            // Auto-assign to closing staff or the last respondent if the
+            // is agent available and has access. Otherwise, put the ticket back
```
greezybacon: agent **is** available

PR #2772, opened 2015-12-03
Default Role

protich:
Add support for default role for assigned tickets/tasks outside the scope of agent's department.

### DIFF
--- a/include/class.staff.php
+++ b/include/class.staff.php
@@ -428,14 +428,20 @@ implements AuthenticatedUser, EmailContact, TemplateVariable {
         return isset($this->locale) ? $this->locale : 0;
     }
 
-    function getRole($dept=null) {
+    function getRole($dept=null, $default=true) {
         $deptId = is_object($dept) ? $dept->getId() : $dept;
         if ($deptId && $deptId != $this->dept_id) {
             if (isset($this->_roles[$deptId]))
                 return $this->_roles[$deptId];
 
+            // Check extended access.
             if ($access = $this->dept_access->findFirst(array('dept_id' => $deptId)))
                 return $this->_roles[$deptId] = $access->role;
+
+            // Agent doesn't have access to the department
+            // return default role - assume assignment
+            if ($default && ($a = $this->dept_access->findFirst(array('dept_id' => 0))))
+                return $this->_roles[$deptId] = $a->role;
 
             // View only access
             return new Role(array());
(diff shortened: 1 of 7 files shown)


include/class.staff.php
```diff
@@ -1069,38 +1075,47 @@ function update($vars, &$errors) {
...
+
+        if ($errors)
+            return false;
+
+        $this->dept_access
+            ->filter(Q::not(array( 'dept_id__in' => $depts
+                            ? array_keys($depts)
+                            : array(-1))
+                        ))
+            ->delete();
```
greezybacon: Maybe
```php
if (count($depts))
    $this->dept_access
        ->filter(Q::not(array('dept_id__in' => array_keys($depts))))
        ->delete();
```
No need sending an ineffective statement to the database

include/class.task.php
```diff
@@ -261,6 +261,13 @@ function getTitle() {
         return $this->__cdata('title', ObjectModel::OBJECT_TYPE_TASK);
     }
 
+    function isAssignee($staff) {
```
greezybacon: I would suggest `isAssignedTo()` as the name

include/i18n/en_US/help/tips/staff.agent.yaml
```diff
@@ -73,11 +73,20 @@ assigned_group:
 primary_department:
     title: Primary Department
     content: >
-        Choose the primary <span class="doc-desc-title">department</span> to which this Agent belongs.
-
+        Choose the primary <span class="doc-desc-title">department</span> to which this Agent belongs and an affective <span class="doc-desc-title">Role</span>.
```
greezybacon: *effective* role

include/staff/staff.inc.php
```diff
@@ -282,8 +283,37 @@ class="staff-username typeahead"
...
             <?php echo __('Extended Access'); ?>
+            <div><small><?php echo __(
+            "Select the departments the agent is allowed to access and the effective roles."
+          ); ?>
+            </small></div>
           </th>
         </tr>
+        <tr>
+          <td>
+            <span class="faded">Assignment Access</span>
```
greezybacon: Use the `__` function here

greezybacon:
It looks like we're adding a record to the staff-dept access table where `dept_id` => 0 to represent the "assignment" role. I have to wonder, since `%staff` table has an `extra` column used to stash JSON data, should we add the role_id there?

```php
class Staff {
    ...
    function getAssignmentRoleId() {
        return $this->getExtraAttr('assn_role_id');
    }

    function setAssignmentRoleId($role_id) {
        $this->setExtraAttr('assn_role_id', $role_id);
    }
}
```
We could move it to a column in a later version

greezybacon:
I think we should add a checkbox option to the "Primary Department and Role" section on the agent access tab. The option should configure the agent to use its primary role for directly assigned tickets when agent has no prescribed access the assigned department. In such a case the agent could have access according to its primary role (if the option is set) or read only access (if not set).

PR #2788, opened 2015-12-10
Fix or Issue/2669

### LINKED ISSUE #2669: 1.10 tip of devevlop-next 1aaa313  Client search and count is broken
When searching as an enduser the search returns all of their tickets:

image

I searched "findme" and  both tickets for user were shown.  Note that the ticket count is incorrect, showing 3 when only 2 tickets are for the logged in user.  This is a green install from the tip commit 1aaa313.

Search last worked at commit f6ccc3f

kest874:
This resolves #2669 (Client Search and Count)


kest874:
Closing this not sure why it has all the additional commits.  Will resubmit

PR #2807, opened 2015-12-18
Properly support scp and admin app plugins

greezybacon:
This patch fixes several issues with creating a plugin to add a "Application" to either the scp or the admin panel. The main issues are:
- URLs in the navigation assume the URL ends with `.../scp/page.php`, when the application is served from `.../scp/apps/name/page`, then the navigation links will be broken
- The PHP CLI server does not respect special dispatchers like the one used to dispatch apps, the api, and the pages URLs. `.../scp/apps/name/page` is rewritten by `.htaccess` to `.../scp/apps/dispatcher.php/name/page`. 
- The apps module did not have the concept of both _admin_ apps and _scp_ apps. Different signal collectors are added here for `apps.scp` and `apps.admin` with appropriate checks for `staff->isAdmin()` to ensure proper access.

This patch includes a new CLI module to facilitate properly running osTicket via the built-in CLI server, available as of PHP 5.4. It includes compatible logic to that used in the shipped `.htaccess` and `web.config` files to facilitate magic URL rewriting.

Usage:

``` sh
$ php manage.php serve --help
```

### DIFF
--- a/include/cli/modules/serve.php
+++ b/include/cli/modules/serve.php
@@ -1,0 +1,95 @@
+<?php
+
+class CliServerModule extends Module {
+    var $prologue = "Run a CLI server for osTicket";
+
+    var $options = array(
+        'port' => array('-p','--port',
+            'default'=>'8000',
+            'help'=>'Specify the listening port number. Default is 8000',
+        ),
+        'host' => array('-h','--host',
+            'default'=>'localhost',
+            'help'=>'Specify the bind address. Default is "localhost"',
+        ),
+    );
+
+    function make_router() {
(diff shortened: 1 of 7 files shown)


greezybacon:
Now we just need to complete the ORM usage and add a SQLite backend. Then it would be possible to run osTicket in a development environment with *only* PHP!

Chefkeks:
> Now we just need to complete the ORM usage and add a SQLite backend. Then it would be possible to run osTicket in a development environment with only PHP!


Awesome / Great! :+1:

Hoimar:
Same experience here, currently you can't register an application without modifiying the ``dispatcher.php` file. This, and the missing documentation for plugins makes it qutie hard to get something running.

Issue #1959, opened 2015-04-14
FEATURE REQUEST: ticket scheduling


rp-tech:
Good day,

It would be awesome if there was a feature where techs could schedule their tickets throughout the day.

While I understand OST is not a calendar / scheduling tool per se, it would be great to have the ability for the system to send out reminders throughout the day for techs to start working on specific tasks.

Example: User calls in @ 8:30 AM, tech is swamped. Tech schedules the ticket for 1:00 PM, ticket system sends out an email reminder a few minutes before for the tech to start  working on the issue.

ntozier:
What is most calendar apps, with email notification (like Google Calendar) FTW?
Seriously though, you can do this pretty easily like I just suggested.

rp-tech:
Hmm, I'm not sure I follow. Are you suggesting just manually have to add calendar entries into my personal calendar every time I schedule a ticket?

Yes that is a workaround but still fairly inconvenient for a busy team of techs.

The reason I asked is, it would be a great addition for busier environments, and it doesn't seem that for off from the existing 'overdue' feature, to have the ability for the system to simply say 'send me an email reminder to work on this ticket @ this time'. In fact, I am using the 'due date' feature to do this atm but it is a bit kludgy and fudges with the response time / ticket close stats since it makes it look like a ticket is going overdue every time.

mckaygerhard:
this feature are a great , remedy has already this! this its a wonderfull feature ovbiously for large enterprise deployments!

harrisyn:
+1

t-oster:
I would suggest it a bit different:
There are some tasks/tickets, which cannot be done right now, e.g. because I am waiting for something. I do not want them to fill my list, but if I close them, I will forget them.
Thus I would like to postpone them like "close and reopen at <datetime>". Then the ticket should be closed and on the specified time just be re-opened again.

ntozier:
So you want a snooze button for your tasks/tickets.  That sounds neat.

t-oster:
was not as hard to implement as I thought. But of cause this needs to be cleaned up.

rollopack:
+1

PR #2811, opened 2015-12-21
Feature: Snooze tasks and tickets

t-oster:
This is my first draft of the feature requested in #1959.
# Description
It adds the possibility to automatically reopen a task/ticket at a specific point in time.

a

b

# Install
In order to use this, you have to add the column autoreopen to ost_ticket and ost_task like

````
ALTER TABLE  `ost_ticket` ADD  `autoreopen` DATETIME NULL DEFAULT NULL;
ALTER TABLE  `ost_task` ADD  `autoreopen` DATETIME NULL DEFAULT NULL;

````

It is based develop-next, but I have some patches from @protich s task-actions branch (https://github.com/protich/osTicket-1.8/tree/feature/task-actions), so it may only work with tickets, not tasks if you add this to develop-next.

# Future
I am not that into the OsTicket codebase, so the code is rather hacky. If someone of the dev-team likes this feature, I think it should be re-implemented in a clean way. At least the database columns have to be created on install/update, but I did not find the place in the code.
However it would be cleaner to add the "autoreopen" as a property of the Task and Ticket objects and allow proper handling through the ORM.

### DIFF
--- a/include/ajax.tasks.php
+++ b/include/ajax.tasks.php
@@ -487,6 +487,81 @@ class TasksAjaxAPI extends AjaxController {
         include STAFFINC_DIR . 'templates/delete.tmpl.php';
     }
 
+   function changeStatus($tid, $status) {
+        global $thisstaff;
+        $statuses = array(
+                'open' => __('Reopen'),
+                'closed' => __('Close'),
+                );
+
+        if(!($task=Task::lookup($tid)) || !$task->checkStaffPerm($thisstaff))
+            Http::response(404, __('No such task'));
+
+        $perm = null;
+        $info = $errors = array();
+        switch ($status) {
+        case 'open':
+            $perm = Task::PERM_CREATE;
+            $info = array(
+                    ':title' => sprintf(__('Reopen Task #%s'),
+                        $task->getNumber()),
+                    ':action' => sprintf('#tasks/%d/reopen',
+                        $task->getId())
+                    );
+            break;
+        case 'closed':
+            $perm = Task::PERM_CLOSE;
+            $info = array(
+                    ':title' => sprintf(__('Close Task #%s'),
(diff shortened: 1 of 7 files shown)


jdelhome3578:
This looks to rely to much on open and close status and will cause issues with SLA and reopend time stamp.

t-oster:
Well, in order to be more generic one could change it to "change status to [v Statusdropdown] at [Date] [Time]". Then of cause there should be an extra table saving all those "timers".
The timestamps should be correct, since I am using the very same code to reopen a ticket, which is used when manually reopening it.

FG-Lars:
I see this is a few years old, but still a feature which would be excellent to have.

I asked for this last year myself, in the osTicket forum: https://forum.osticket.com/d/91684-ticket-reminder-followup

> The tickets can be regarding everything from new employees or the need of a new switchboard number to the planned maintenence on one of our buildings or removing an e-mail forwarder on a spesific date.
> 
> The timespan before action is needed on the ticket, can range all form a couple of hours to several months ahed in time.
> 
> The method we use today is that the staff member who claims the ticket, creates a task in he's Outlook with a reminder and then closes the ticket with the comment that he will followup at the appropriate time.
> 
> The reason we close them, is that the ticket list would be very long if we did not. So we close them to keep it tidy, and then reopen when the Outlook reminder pops up.

As you can see. A snooze would be great.

So is anyone working on this?

t-oster:
Well Not working on it but still using my above Patch successfully

FG-Lars:
Does it still work on v1.11?

t-oster:
I am still in the process of merging my custom Version to 1.11, but its low on my priority list. Maybe you give it a try and see what happens?

FG-Lars:
Think I'll wait until someone comfirms stabile conditions.

PR #2812, opened 2015-12-22
bug: Set due date based on user's timezone

protich:


include/class.ticket.php
```diff
@@ -2740,7 +2740,7 @@ function update($vars, &$errors) {
             elseif (strtotime($vars['duedate'].' '.$vars['time']) === false)
                 $errors['duedate']=__('Invalid due date');
             // FIXME: Using time() violates database and user timezone
-            elseif (strtotime($vars['duedate'].' '.$vars['time']) <= time())
+            elseif (Misc::usertime($vars['duedate'].' '.$vars['time']) <= Misc::usertime())
```
greezybacon: I would suggest naming it `user2gmtime` since the output is GMT from the function

protich: Done

PR #2841, opened 2016-01-04
collabs: Add avatars to the popup mgmt dialog

greezybacon:
And, add to the collabs preview popup too, including collaborators tab in the ticket preview.


greezybacon:
I think it would be neat in the custom columns feature to add a decoration of an avatar listing like Github does in the notification window

Chefkeks:
> I think it would be neat in the custom columns feature to add a decoration of an avatar listing like Github does in the notification window

Yes! :+1:

Android00700 (approved):

Issue #2823, opened 2015-12-29
Incorrect "Thank you" pages displayed


GoodbyeWorld:
I think I found a bug. 

I have a use case that requires multiple Help Topics, each with a different "Thank you" page. 

This should be doable by simply creating multiple "Thank you" pages, each with the content they require, and configuring each Help Topic to use the corresponding "Thank you" page. 

The issue is that they are not displayed to the user. Instead, the "Thank you" page configured at the Company level is used, even if that "Thank you" page is disabled. 

This is simple to reproduce, just create a new "Thank you" page, configure a Help Topic to use it, and without signing-in as an agent or user, create a new ticket. You'll be taken to the Company configured Thank you page. 

I reproduced this on both the osTicket Download page version of 1.10 RC2, and a current version I retrieved via `git clone -b develop-next https://github.com/osTicket/osTicket-1.8 .` and then deploying etc..

GoodbyeWorld:
Just wondering if there's anything I can do that would help the development team with this one.  I notice there's been no response - is that because you need more info?  Just let me know what you need to tackle this and I'll do what I can.

greezybacon:
I see the problem is `class.topic.php` line 63.
```php
var $page;
```
should be deleted. That's what's breaking the system

GoodbyeWorld:
@greezybacon - I tested your suggestion and it appears to have worked.  Thanks!

I'm new to github - are you planning to correct that, or would it be helpful for me to submit a pull request with that adjustment?

PR #2842, opened 2016-01-05
Fix incorrect Thank-you page being shown issue #2823

GoodbyeWorld:
See https://github.com/osTicket/osTicket-1.8/issues/2823

Thank-you pages configured at a Help Topic level should override the Thank-you pages configured at a Company level.

With this fix, osTicket correctly shows the Thank-you page configured in the Help Topic.


greezybacon:
So it looks like the problem must be in Topic::getPage

protich:
@GoodbyeWorld  -- Thank you for the PR. As @greezybacon mentioned the issue is with Topic::getPage. Pull request #2915 addresses the root cause.

PR #2852, opened 2016-01-10
Feature :: File content processors

greezybacon:
In the software, they're called `FileStorageBackendAdapter`, and they can be written as plugins. The idea is that the content of file might be modified between the user interface and the backend storage, and visa versa. The best example of this would be to encrypt the file contents.

### Encryption
This adds the ability to encrypt file contents when saving them in the storage backend. This would be very useful for compliance requirements; or to guard against data leakage if the backend (S3 in AWS, for instance) cannot be trusted or might be backed up to an lower security medium.

### Compression
Some files attached to the system are very compressible, such as bitmap images and text documents. This adds the option to compress such documents, or to apply quick compression on all file contents uploaded.

##### Add settings to compress and encrypt files
screen shot 2016-01-10 at 5 07 39 pm

##### Encrypted and compressed file contents viewed in the database (a text file)
screen shot 2016-01-10 at 5 12 32 pm
*This is a 160k PDF document which was compressed to 130k and encrypted using AES-256(CBC) with a system-configurable password*

### Thoughts
The actual encryption and compression adapters might work better as plugins, which would allow them to have far more elaborate configuration without cluttering up the admin panel. For instance, selecting which files should be compressed or encrypted, or which encryption algorithm or mode should be used. However, there's always the option to add these two basic options to the product and create plugins for more elaborate setups.

The encryption option should probably be used with another option to avoid sending attachments in emails and replace them with links which would direct the email recipient back to the system. Since the links have an expiration, it provides a maximum lifetime of the attachment for all recipients; even if the email is accidentally forwarded.

### Outstanding
  * [ ] Verify backend migration
  * [ ] Verify CLI import / export works
  * [ ] Verify system backup and file export do not leak encrypted file contents
  * [ ] Add CLI option to add adapter to an exiting file. For example, compress one or more files currently stored in the system

### DIFF
--- a/include/class.file.php
+++ b/include/class.file.php
@@ -113,7 +113,10 @@ class AttachmentFile extends VerySimpleModel {
     function open($raw=false) {
         $bk = FileStorageBackend::getInstance($this);
         if (!$raw && ($adapters = $this->getAttribute('stack'))) {
-            $bk = $this->pushAdapters($bk, $adapters);
+            // Reverse the order of adapters used when creating the file's
+            // storage data. If it was compressed, then encrypted, we want
+            // to decrypt, then decompress (ie. reverse).
+            $bk = $this->pushAdapters($bk, array_reverse($adapters, true));
         }
         return $bk;
     }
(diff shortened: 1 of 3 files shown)


include/class.file.php
```diff
@@ -799,6 +858,330 @@ function getSize() {
...
+    }
+    function setupUploadStream($stream) {
+        stream_filter_append($stream, 'mcrypt.'.$this->algo,
+            STREAM_FILTER_READ, $this->getKeyAndIv());
+    }
+
+    static function isEnabledForFile($file) {
+        global $cfg;
+
+        if ($cfg->get('encrypt_files') && $cfg->get('file_encrypt_pwd'))
```
greezybacon: Also verify `mcrypt` extension is loaded

include/staff/settings-system.inc.php
```diff
@@ -359,6 +359,40 @@ class="help-tip icon-question-sign" href="#attachments"></i></em>
...
+foreach (CompressionFileAdapter::allModes() as $v=>$desc) {
+    $selected = $config['compress_files'] == $v
+        ? 'selected="selected"'
+        : '';
+    echo sprintf('<option value="%s" %s>%s</option>',
+        $v, $selected, $desc);
+} ?>
+                </select>
+        </tr>
+        <tr>
```
greezybacon: Verify `mcrypt` extension is loaded

PR #2857, opened 2016-01-12
Fix for issue 2855

kest874:
Fix for https://github.com/osTicket/osTicket-1.8/issues/2855


protich:
Thanks for the suggestion - the root cause fixed in develop-next via 9816835.

PR #2864, opened 2016-01-15
Users & Orgs export

protich:
* Fix users export
* Fix orgs export
* Convert orgs list page to ORM


protich:
## Why CDATA tables for Users and Orgs?
I'm not excited about creating more materialized view tables, but it is the quickest solution I could come up with at the moment without database signature/models changes.  Longterm, we would want to retire CDATA tables and rely on `form entry value` table -- which requires major enhancement(s) on `forms api` data storage model to  __store multi-entry values separately instead of JSONed format we use now__. 

The change will allow for direct data query without the expensive joins and parsing that forced use materialized view tables (cdata).

PR #2865, opened 2016-01-15
Feature: Auto Close for v1.10

jdelhome3578:
This is an updated PR for the Auto Close feature. This PR will work with v1.10 and also allows for admin to select the status from which will be used to automatically close tickets.

Need help:
 - [ ] set status using status name instead of id in AutoClose function

![image]
(https://cloud.githubusercontent.com/assets/5774055/12360897/a879218c-bb88-11e5-94d5-5e3bfbc1fba6.png)


kest874:
Works well needs a little cleanup and has a few typos.

jdelhome3578:
> $criteria = array('states' => array('closed')) 
> really should be 
> $criteria = array('states' => array('open'))
> 
> Otherwise you can only auto-close the closed tickets.....

@kest874 I considered your suggestion and opened the drop down to all statuses. We believe that even though we mark it as pending that the state of the Pending status would be closed this helps with the reopened ticket status. Please keep in mind states and statuses.

kest874:
@jdelhome3578 True, didn't think of it that way.  We set a status of "Waiting for Submitter Reply" which is an "Open" state, our agreement is that we close the ticket if there isn't a response in 72 hours.

include/class.ticket.php
```diff
@@ -3620,5 +3620,34 @@ static function agentActions($agent, $options=array()) {
...
+
+		// select all tickets marked as selected status from config page where updated is older than ($grace) hours ago
+		$sql  = 'SELECT ticket_id FROM '.TICKET_TABLE.' ticket '
+			 .' INNER JOIN '.TICKET_STATUS_TABLE.' status ON (status.id=ticket.status_id AND status_id = '. $graceStatusId .') '
+			 .' AND TIME_TO_SEC(TIMEDIFF(NOW(),ticket.updated))>='.$grace.'*3600';
+
+	        if(($res=db_query($sql)) && db_num_rows($res)) {
+	        	while(list($id)=db_fetch_row($res)) {
+	        		if($ticket=Ticket::lookup($id)) 
+				$ticket->setStatus('3', 'Ticket Closed by the SYSTEM after '.$grace.' '.$plural.' of no activity.',$errors, false);
```
jdelhome3578: Need help: set status using status name instead of id
@greezybacon @protich

jdelhome3578:
I removed extra commits in #2917

PR #2891, opened 2016-01-29
Missing global from the function that actually uses it

timwhite:
Seems to already be fixed in develop, but the global $cfg needs to be in the getVar function in develop-next otherwise we get a fatal error


protich:
@timwhite -- Thanks! Already fixed via efe3a3e6

Issue #2902, opened 2016-02-02
global $ost not defined when $plugin->bootstrap()


quazardous:
Hi,

can you just add one line to allow plugins to access $ost ?

``` php
class osTicket {
...
    function start() {
        global $ost; // <--- this line
        // Prep basic translation support
        Internationalization::bootstrap();

        if(!($ost = new osTicket()))
            return null;

        //Set default time zone... user/staff settting will override it (on login).
        $_SESSION['TZ_OFFSET'] = $ost->getConfig()->getTZoffset();
        $_SESSION['TZ_DST'] = $ost->getConfig()->observeDaylightSaving();

        // Bootstrap installed plugins
        $ost->plugins->bootstrap();

        // Mirror content updates to the search backend
        $ost->searcher = new SearchInterface();

        return $ost;
    }
}


```

quazardous:
i think it can fix #1074

You can inject headers...
```php
...
$ost->addExtraHeader('<!-- BOUYA -->');
...
```

fnbf:
Ping.

PR #2907, opened 2016-02-03
Allow plugins to access $ost global in bootsrap()

quazardous:
fix for #2902

### DIFF
--- a/include/class.osticket.php
+++ b/include/class.osticket.php
@@ -495,6 +495,7 @@ class osTicket {
 
     /**** static functions ****/
     function start() {
+    	global $ost;
         // Prep basic translation support
         Internationalization::bootstrap();
 


quazardous:
@greezybacon it's a no brainer, maybe you could merge it :)

clonemeagain:
:+1: 

Much better for simple CSS/JS things. Looking forward to the merge conflict!

clonemeagain:
@protich @greezybacon Pinging you guys so you can see why we want this. (Or something like it)

Currently: Adding a stylesheet or script to osTicket requires modifying core, adding it via a plugin still requires modifying core, unless we can use a Signal, which there isn't for Every pageload. 

Check out the requirements to inject a script to a ticket:https://github.com/clonemeagain/osticket-plugin-notefirst/blob/master/NoteFirstPlugin.php It has two bootstrap methods, both doing the same thing, one requires another entire plugin just to inject the script, the other uses the $ost method.

Simply having access to `$ost` in the bootstrap method is a boon as it lets us inject a new header with stylesheet, script, <whatever>Yay!</whatever> contents we want. Great if that is the sole point of a plugin!

The problem with this PR is: $ost isn't fully loaded into the global scope until _after_ the `Plugin::bootstrap()` methods are run, so that object isn't available as a global on bootstrap.. (considering that it is loading itself at the time, that's understandable, however, we still want to use methods from it). 

Now, we could run our bootstrap plugins and add Signal hooks in them, then use global $ost in those Signal functions, however, this isn't always possible, as you know, there aren't a lot of Signals, and the ones that are there aren't designed for editing stylesheets or scripts. 

Maybe instead we can get a `Signal::send('bootstrap',$ost, &$errors);` at the end of main.inc.php?
I'd be onboard with that. We'd still have $ost, we'd have a fully loaded osTicket object instead of a partially bootstrapped one.

clonemeagain:
@quazardous 

Example using Signals only:

Added to end of main.inc.php:
```php
// patch related to https://github.com/osTicket/osTicket/pull/2907#issuecomment-312804184
Signal::send('bootstrap', $ost, $errors);
```

Then in the plugin:
```php
function bootstrap() {
	Signal::connect ( 'bootstrap', function ($ost) {
		$ost->addExtraHeader ( '<style>/* CSS */</style>' );
	} );
}
```

quazardous:
$ost is not fully loaded but plugins stuff is a the end of the bootraping.
In my experience it worked like a charm. Not very elegant but he have you
seen the MVC implementation ? Yeah have you seen it ? Have not lol.

clonemeagain:
> bootraping

 lmfao

Close enough to the end of Bootstrap I suppose. In practice, it runs all Plugin::bootstrap()'s before it checks the config.. so the config could fail to load and our code wouldn't know.  https://github.com/osTicket/osTicket/blob/develop/main.inc.php#L33

quazardous:
yes it's a risk...
but my point is not if the osTicket boostrap is messy (or not ?), it is
that plugins NEEDS a way to access $ost in early stage directly or not
(signals is good too).
global $ost does a great job with a minimalistic effort ;p

clonemeagain:
I agree that plugins need $ost, I'm just ideologically opposed to globals. :-) Dependency injection is more elegant for testing and my IDE as well.

PR #2936, opened 2016-02-23
emails: Allow duplicate tickets

protich:
osTicket uses message-id to detect previously "seen" emails in order to
prevent duplicate tickets from being created from a single email. This is
necessary when fetched emails are left in the INBOX.

Scenario:
* We have multiple departments with strict access control on tickets e.g
  members of Billing department do not have access to tickets in Legal
  department
* Joe sends an email to billing@company.com and legal@company.com (CC)
 - If a ticket gets created in Legal first (unpredictable - depends on which
   inbox gets processed first)
 - Billing department will be left in the dark
 - No way for legal to know that the same email was also addressed to Billing

 This pull request allows for less strict matching which can allow for
 duplicate tickets when an email is sent to multiple departments / email
 queues.

 To credit the correct email account the system will prefer
 Delivered-To address, followed by the account being fetched and if all
 fails the first system email encountered.

__TODO__
 * [ ]  Add setting to Admin Panel > Emails > Setting


greezybacon:
Should the setting be applied per department rather than globally?

protich:
I thought about it, but I figured global setting makes sense since it's multi-department feature. For example, it doesn't make sense for **only** one department to allow dual delivery.

greezybacon:
Perhaps then per email address?

I guess my hangup is in that this feature just feels like *your're doing it wrong* if you have to use it. Perhaps I'm too closed minded.

PR #2939, opened 2016-02-25
Client interface tweaks

datatecuk:
Added the ability to remove the staff login link from the client login page from the admin panel.
Added the code in the header for the favicon.ico file so users can simply upload their favicon.ico file into the root directory of their osTicket installation.
Also modified the footer copyright so it now includes a hyperlink to the website set in the admin panel.


infectormp:
:+1:

PR #2940, opened 2016-02-26
iFrame 

protich:
This pull request addresses a case where iframe gets dropped as well was potential XSS security exploit in html cleanup.

* Whitelist iframe as allowed html node without content when dom (php-xml) is used to balance the tags.

* Decode html entities before sanitization to avoid potential htmLawed bypass. By default, Format::safe_html will now decode html entities before the balancing and cleanup is done.

### DIFF
--- a/include/class.format.php
+++ b/include/class.format.php
@@ -136,7 +136,8 @@ class Format {
             // Remove empty nodes
             $xpath = new DOMXPath($doc);
             static $eE = array('area'=>1, 'br'=>1, 'col'=>1, 'embed'=>1,
-                'hr'=>1, 'img'=>1, 'input'=>1, 'isindex'=>1, 'param'=>1);
+                    'iframe' => 1, 'hr'=>1, 'img'=>1, 'input'=>1,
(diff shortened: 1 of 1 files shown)


include/class.format.php
```diff
@@ -218,6 +219,17 @@ function html2text($html, $width=74, $tidy=true) {
     static function __html_cleanup($el, $attributes=0) {
         static $eE = array('area'=>1, 'br'=>1, 'col'=>1, 'embed'=>1,
             'hr'=>1, 'img'=>1, 'input'=>1, 'isindex'=>1, 'param'=>1);
+
+        // We're dealing with closing tag
+        if ($attributes === 0)
+            return "</{$el}>";
+
+        // Remove iframe and embed without src (perhaps striped by spec)
+        // It would be awesome to rickroll such entry :)
```
greezybacon: :+1:

greezybacon:
I'm trying to open a ticket with bioinformatics on the Unicode corruption issue. Looks like I can't post on their forum even after confirming my email address...

Issue #2871, opened 2016-01-19
XSS Vulnerability - open.php


mattkoch614:
The open.php (New Ticket) page is vulnerable to XSS attacks. I was able to change my profile "Full Name" field to the following:

```
"<script>alert("vulnerable")</script>
```

This resulted in the script being executed when I loaded the open.php page (see attachment).

![xss_vulnerability](https://cloud.githubusercontent.com/assets/8203609/12430382/528a5754-bebd-11e5-892f-04e2ed3ff0d5.PNG)

Chefkeks:
What version of osTicket are you running?

ntozier:
I can confirm that this can be done on 1.10rc2.

Side note: the name field appears to be limited to 128 characters.

noleafclover614:
@chefkeks
1.9.12

Chefkeks:
Thanks!

I sure @protich @greezybacon will take care of that.

PR #2941, opened 2016-02-29
Textarea input cleanup

protich:
Strip tags on text input field to prevent potential XSS exploit.

Addresses #2871

### DIFF
--- a/include/class.thread.php
+++ b/include/class.thread.php
@@ -1434,7 +1434,14 @@ class ThreadBody /* extends SplString */ {
     }
 
     function getClean() {
-        return trim($this->body);
+        switch ($this->type) {
+        case 'html':
+            return trim($this->body, " <>br/\t\n\r") ? $this->body: '';
+        case 'text':
+            return trim($this->body) ? $this->body: '';
+        default:
+            return trim($this->body);
+        }
     }
 
     function __toString() {
@@ -1476,6 +1483,13 @@ class ThreadBody /* extends SplString */ {
             return new ThreadBody($text);
         }
     }
+
+    static function clean($text, $format=null) {
+        global $cfg;
+        $format = $format ?: ($cfg->isHtmlThreadEnabled() ? 'html' : 'text');
+        $body = static::fromFormattedText($text, $format);
+        return $body->getClean();
+    }
 }
 
 class TextThreadBody extends ThreadBody {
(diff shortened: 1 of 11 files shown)


protich:
@infectormp - Thanks for pointing it out, it was a back port mishap. Commit rebased to fix the issue.

include/class.thread.php
```diff
@@ -1476,6 +1476,22 @@ static function fromFormattedText($text, $format=false) {
             return new ThreadBody($text);
         }
     }
+
+    static function clean($text, $format=null) {
```
greezybacon: Shouldn't this replace `getClean`? The `ThreadEntry::create()` method uses `getClean` to determine if the entry is empty.

include/class.thread.php
```diff
@@ -1484,7 +1500,8 @@ function __construct($body, $options=array()) {
     }
 
     function getClean() {
-        return Format::stripEmptyLines($this->body);
+        return  Format::stripEmptyLines(
+                self::clean($this->body, $this->format));
```
greezybacon: Based on the above comment, seems like this should become
```php
    return Format::stripEmptyLines(parent::getClean($this->body));
```

greezybacon: Isn't the format stored in `$this->type`?

include/class.thread.php
greezybacon: Perhaps
```php
function getClean() {
       switch ($this->type) {
       case 'html':
           return trim($this->body, " <>br/\t\n\r") ? $this->body: '';
       case 'text':
           return trim($this->body) ? $this->body: '';
       default:
           return $this->body;
       }
}

static function clean($text, $format=null) {
    $format = $format ?: ($cfg->isHtmlThreadEnabled() ? 'html' : 'text');
    $body = static::fromFormattedText($text, $format);
    return $body->getClean();
}
```

include/class.thread.php
```diff
@@ -1527,7 +1544,7 @@ function ($m) use ($self) {
     }
 
     function getClean() {
-        return trim($this->body, " <>br/\t\n\r") ? Format::sanitize($this->body) : '';
+        return Format::sanitize(self::clean($this->body, $this->format));
```
greezybacon:
```php
function getClean() {
    return Format::sanitize(parent::getClean());
}
```

PR #2944, opened 2016-03-01
v1.10 rc2++ Bug Fixes

protich:


greezybacon:
is 8d00d8e about the new ticket autoresponse going out if not configured?

include/class.ticket.php
```diff
@@ -3382,8 +3392,12 @@ static function create($vars, &$errors, $origin, $autorespond=true,
...
         // Assign ticket to staff or team (new ticket by staff)
         if ($vars['assignId']) {
-            $asnform = $ticket->getAssignmentForm(array('assignee' => $vars['assignId']));
-            $ticket->assign($asnform, $vars['note']);
+            $asnform = $ticket->getAssignmentForm(array(
+                        'assignee' => $vars['assignId'],
+                        'comments' => $vars['note'])
+                    );
+            $e = array();
+            $ticket->assign($asnform, $e);
```
greezybacon: My personal on this style is that we should use something like this instead:
```php
$asnform = $ticket->getSomeKindaForm(['key' => $value]);
$ticket->assign($asnform->getClean());
```

include/class.ticket.php
```diff
@@ -2012,7 +2016,8 @@ function transfer(TransferForm $form, &$errors, $alert=true) {
 
         // Set SLA of the new department
         if (!$this->getSLAId() || $this->getSLA()->isTransient())
-            $this->selectSLAId();
+            if (($slaId=$this->getDept()->getSLAId()))
```
greezybacon: FYI, you don't need double parentheses here

include/staff/ticket-edit.inc.php
```diff
@@ -70,12 +70,17 @@ function(user) {
...
-                    <option value="Other" <?php echo ($info['source']=='Other')?'selected="selected"':''; ?>><?php echo __('Other');?></option>
+                    <option value="" selected >&mdash; <?php
+                        echo __('Select Source');?> &mdash;</option>
+                    <?php
+                    $source = $info['source'] ?: 'Phone';
+                    foreach (Ticket::getSources() as $k => $v) {
+                        echo sprintf('<option value="%s" %s>%s</option>',
+                                $k,
+                                ($source == $k ) ? 'selected="selected"' : '',
+                                $v);
```
greezybacon: I think this will need to be translated
```php
__($v)
```
because the getSources() method does not translate it, and it is only marked as translatable in the constant definition

greezybacon: As a side note, if we added `Ticket::addSource($k, $translated_name)`, then a plugin could be used to add new sources...

greezybacon: On second thought, since the sources will likely be used in several places, I think they should be translated in the `Ticket::getSources()` method

include/class.report.php
```diff
@@ -35,13 +35,19 @@ function getDateRange() {
...
-
-        if (substr($stop, 0, 1) == '+')
-            $stop = strftime('%Y-%m-%d ', $start) . $stop;
+        // Convert user time to db time
+        $start = Misc::dbtime($start);
+        // Stop time can be relative.
+        if ($stop[0] == '+') {
+            // $start time + time(X days)
+            $now = time();
+            $stop = $start + (strtotime($stop, $now)-$now);
```
greezybacon: Isn't this reintroducing the original problem?

include/staff/ticket-edit.inc.php
protich: It is already translated statically - see `static protected $sources =  array` in ticket class.

protich: As for plugins adding sources -- the problem is the ticket table source field is currently an ENUM. The plan is to move source definition to Admin Panel > Manage > Lists and perhaps allow plugins to add to the list.

include/class.ticket.php
protich: I somewhat agree -- I simply used the form here as a assignment request. I think we should move more towards __request based transaction__ as opposed to hash array (good example is ticket create request). This is highly important for request interception, modification (automation) and uniform data validation that is not tied to html input elements. 

My thinking is internal API, automation and pluggable mechanisms could use direct calls like `setStaffId`.

include/staff/ticket-edit.inc.php
greezybacon: @protich the `/* @trans */` decorator indicates to the translation system that the string should be included in the `messages.po` output and should be translatable. It still has to be translated at runtime with the `__()` function.

greezybacon:
```patch
diff --git a/include/class.ticket.php b/include/class.ticket.php
index b2a2251..3433a8e 100644
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -216,7 +216,12 @@ EOF;
     }
 
     static function getSources() {
-        return self::$sources;
+        static $translated = false;
+        if (!$translated) {
+            foreach (static::$sources as $k=>$v)
+                static::$sources[$k] = __($v);
+        }
+        return static::$sources;
     }
 }
 ```

PR #2951, opened 2016-03-06
Fix page size returned in ajax config

antriver:
The config returned by **/scp/ajax.php/config/scp** would return 0 if the agent had their "Maximum page size" set to "system default". In that case it should return the global system setting for maximum page size.

This was causing the Statistics not to load on the Dashboard page, with this error:
[](http://ctrlv.in/722941)
Because of these lines line **dashoard.inc.js**:
```
                    if (i % pagesize === 0)
                        b = $('<tbody>').attr({'page':i/pagesize+1}).addClass('hidden').appendTo(q);
                    row = json.data[i];
                    tr = $('<tr>').appendTo(b);
```
So this PR also fixes that.

### DIFF
--- a/include/ajax.config.php
+++ b/include/ajax.config.php
@@ -38,7 +38,7 @@ class ConfigAjaxAPI extends AjaxController {
               'lang'            => $lang,
               'short_lang'      => $sl,
               'has_rtl'         => $rtl,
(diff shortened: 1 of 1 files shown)


protich:
@antriver - Thank you for the pull request. I sorta rebased your commit a little (0c29477) to make it clear that we're getting agent's page limit.

Issue #2602, opened 2015-09-30
timeline to support PHP7


jjjj12212:
PHP 7 is around the corner and I tried the application with PHP7 and getting alot of deprecated issues with the Class methods having the same name as the class (the constructor) then nothing happens (a blank screen).

Is there a timeline when osTicket will be supporting PHP 7?

greezybacon:
> Is there a timeline when osTicket will be supporting PHP 7?

Wanna do the work and open a pull request? I have run osTicket v1.10 (`develop-next`) on PHP 7 with just a few changes, so I don't think the codebase is *too* far off

infectormp:
Here is some deprecated warnings for PHP 7:
```Method name "Services_JSON" in class "Services_JSON" is deprecated (@php4_constructors) in file /xxx/include/JSON.php[132].
Method name "Services_JSON_Error" in class "Services_JSON_Error" is deprecated (@php4_constructors) in file /xxx/include/JSON.php[782].
Method name "Services_JSON_Error" in class "Services_JSON_Error" is deprecated (@php4_constructors) in file /xxx/include/JSON.php[796].
Method name "PasswordHash" in class "PasswordHash" is deprecated (@php4_constructors) in file /xxx/include/PasswordHash.php[33].
Method name "AjaxController" in class "AjaxController" is deprecated (@php4_constructors) in file /xxx/include/class.ajax.php[28].
Method name "API" in class "API" is deprecated (@php4_constructors) in file /xxx/include/class.api.php[22].
Method name "Captcha" in class "Captcha" is deprecated (@php4_constructors) in file /xxx/include/class.captcha.php[21].
Method name "Config" in class "Config" is deprecated (@php4_constructors) in file /xxx/include/class.config.php[32].
Method name "OsticketConfig" in class "OsticketConfig" is deprecated (@php4_constructors) in file /xxx/include/class.config.php[182].
Method name "CryptoAlgo" in class "CryptoAlgo" is deprecated (@php4_constructors) in file /xxx/include/class.crypto.php[234].
Method name "CSRF" in class "CSRF" is deprecated (@php4_constructors) in file /xxx/include/class.csrf.php[37].
Method name "Dispatcher" in class "Dispatcher" is deprecated (@php4_constructors) in file /xxx/include/class.dispatcher.php[24].
Method name "UrlMatcher" in class "UrlMatcher" is deprecated (@php4_constructors) in file /xxx/include/class.dispatcher.php[84].
[...]

PR #2828, opened 2015-12-30
php: Support PHP 7, require at least 5.4

greezybacon:
This commit attempts to remove all coding standard warnings emitted by PHP 7.0. There's still some work to be done; but this gets things kinda working.

Fixes #2819
Fixes #2815
Fixes #2602

PR #2970, opened 2016-03-24
Rename Error class

kevinoconnor7:
PHP 7 provides its own Error class which clashes with OsTicket's. Since there's no namespaces being used I just renamed the class to have a localized osTicket specific name.

This should help progress toward #2602


protich:

PR #2971, opened 2016-03-24
Remove improper usage of func_get_args

kevinoconnor7:
In PHP 7 func_get_args will  "do the right thing" if a argument has been mutated, even if it was passed by value.

In this case it ensured that if non-array value was passed in then the resulting criteria would be:
`array($pk => array())`

This should make progress toward #2602


protich:

PR #2972, opened 2016-03-24
Remove deprecated constructor

kevinoconnor7:
In PHP 7 the legacy constructor pattern is deprecated.

This should make progress toward #2602


protich:

PR #2973, opened 2016-03-24
i18n: Eliminate duplicate or incorrect texts

greezybacon:
This is my attempt at removing similar texts from the translator work load. It also makes an attempt at dropping some misspelled words and works toward standardizing some phrases.

### DIFF
--- a/include/staff/slaplans.inc.php
+++ b/include/staff/slaplans.inc.php
@@ -36,7 +36,7 @@ $qs += array('sort' => $_REQUEST['sort'], 'order' => $_REQUEST['order']);
 
 $pageNav = new Pagenate($count, $page, PAGE_LIMIT);
 $pageNav->setURL('slas.php', $qs);
-$showing = $pageNav->showing().' '._N('SLA plan', 'SLA plans', $count);
+$showing = $pageNav->showing().' '._N('Service Level Agreement', 'Service Level Agreements', $count);
 $qstr .= '&amp;order='.($order=='DESC' ? 'ASC' : 'DESC');
 ?>
 <form action="slas.php" method="POST" name="slas">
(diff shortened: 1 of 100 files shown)


scp/roles.php
```diff
@@ -39,11 +39,11 @@
             $msg = __('Role updated successfully');
         } elseif ($errors) {
             $errors['err'] = $errors['err'] ?:
-                sprintf(__('Unable to update %s. Correct error(s) below and try again!'),
+                sprintf(__('Unable to update %s. Correct any errors below and try again.'),
```
Chefkeks: Sometimes its "Correct ANY errors..." and sometimes its "Correct THE errors..." (e.g. /scp/tasks @ Line103). I guess it'd make sense to just use one phrase / sentence and not several to hint at errors which shall be corrected ;)
Rest of this PR looks pretty good so far :+1:

greezybacon:
@Chefkeks thanks for the heads up. I took your lead

include/staff/slaplans.inc.php
```diff
@@ -36,7 +36,7 @@
 
 $pageNav = new Pagenate($count, $page, PAGE_LIMIT);
 $pageNav->setURL('slas.php', $qs);
-$showing = $pageNav->showing().' '._N('SLA plan', 'SLA plans', $count);
+$showing = $pageNav->showing().' '._N('Service Level Agreement', 'Service Level Agreements', $count);
```
Chefkeks: You changed "SLA Plan" / "SLA Plans" to the full length "Service Level Agreement" / "Service Level Agreements" here and in include/class.slas.php, but not in scp/slas.php. Is this for a specific reason or not? If not, you should may change it in scp/slas.php too ;)

greezybacon: Personally, I think Service Level Agreement is more correct; however, honestly, after about six hours, I just got worm out of rephrasing text

Issue #1137, opened 2014-08-04
date/time ticket thread should use time of email not time when ingested into osticket


snizzleorg:
If several emails from same issue are present a the same time in the imap folder watched by osticket these emails tend to show up with a wrong order in the ticket thread.

I noticed that rather using the date when the email was sent osticket uses the date when it ingests the emails. This is problematic if several answers are received in relatively short time intervals.

PR #2975, opened 2016-03-24
Changed ticket create date to be the email date

catbadger:
Changed ticket create date to be the email date as opposed to the ingestion date of the email into osTicket. 
This should happen for a couple reasons
   1. Historical emails must have date preserved on import
   2. If the ticket system goes down or is having problems, it should not throw off dates. This may work as an unintended penalty for downtime.
   3. Multiple emails received at very close times may show as out of order.

### DIFF
--- a/include/api.tickets.php
+++ b/include/api.tickets.php
@@ -39,7 +39,7 @@ class TicketApiController extends ApiController {
 
         if(!strcasecmp($format, 'email')) {
             $supported = array_merge($supported, array('header', 'mid',
-                'emailId', 'to-email-id', 'ticketId', 'reply-to', 'reply-to-name',
+                'emailId', 'to-email-id', 'ticketId', 'reply-to', 'reply-to-name','date',
                 'in-reply-to', 'references', 'thread-type',
                 'mailflags' => array('bounce', 'auto-reply', 'spam', 'viral'),
                 'recipients' => array('*' => array('name', 'email', 'source'))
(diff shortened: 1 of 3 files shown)


catbadger:
Also resolves #1137

[#1137]( https://github.com/osTicket/osTicket/issues/1137)

Chefkeks:
I totally agree with the intention behind this PR!

Hoping the masters @protich and @greezybacon agree with the code changes and implement this into the upcoming release 1.10 ;)

Korrosivo:
Looks great. :+1:

ipso:
This would be extremely helpful for us too.

catbadger:
this should work in both develop and develop-next/

greezybacon:
@catbadger a few comments / opinions
* the `Date` timestamp in the email headers is likely in GMT time. The database time may or may not be in that timezone. The `created` and `lastupdate` timestamps should be changed to the database timezone with `Misc::dbtime()`. For the default, `SqlFunction::NOW()` should be employed to avoid unnecessary timezone conversions.
* We should add a setting to Admin Panel -> Email -> Settings to enable/disable this feature

KiloalphaDE:
Due to a crash of my old environment I am currently installing a new osTicket environment based on the v12.2.2 which I downloaded from osticket.com.

I would like to import the old tickets now to restore the old structure. For that I need this modification.

I have made the changes from @joshr and @catbadger, but the system does not set the time from the email but the import.

Can someone help me with the adaptation to v12.2.2?

The mod for the correct timezone required in the previous post by @ greezybacon is included in v12.2.2.

Thank you in advance!

PR #2985, opened 2016-03-28
FIX: Return-Path header is landing in spam boxes

catbadger:
If the Return-Path header does not match the From header, emails are significantly more likely to be blocked by spam filters.
This patch forces the return header to match the from address without having to change php.ini and therefore works properly on servers that have multiple apps sending from multiple email addresses for example shared hosting, multiple support emails in one instance etc.

### DIFF
--- a/include/class.mailer.php
+++ b/include/class.mailer.php
@@ -561,6 +561,7 @@ class Mailer {
 
         //No SMTP or it failed....use php's native mail function.
         $mail = mail::factory('mail');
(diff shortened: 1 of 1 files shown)


greezybacon:
Thanks for the pull request. I like the edit and think we should use it. However, I don't like manipulating the mailer object's private stuff directly. I think the proper way to do this is:
```php
        $args = array('-f '.$this->getEmail()->getEmail());
        $mail = mail::factory('mail', $args);
```

catbadger:
okie dokie. better now?

PR #2988, opened 2016-03-29
Session: Create SessionData object on demand

protich:
Session backend issues update via write on new session which caused assertion to fail.

### DIFF
--- a/include/class.ostsession.php
+++ b/include/class.ostsession.php
@@ -174,6 +174,8 @@ extends VerySimpleModel {
 
 class DbSessionBackend
 extends SessionBackend {
+    var $data = null;
+
     function read($id) {
         try {
             $this->data = SessionData::objects()->filter([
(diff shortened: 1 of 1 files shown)


include/class.ostsession.php
```diff
@@ -200,7 +202,9 @@ function update($id, $data){
         $ttl = $this && method_exists($this, 'getTTL')
             ? $this->getTTL() : SESSION_TTL;
 
-        assert($this->data->session_id == $id);
+        // Create a session data obj if not loaded.
+        if (!$this->data)
+            $this->data = new SessionData(['session_id' => $id]);
```
greezybacon: Looks like `session_rengenerate_id()` will also change the ID without calling any other functions. I wonder if we should write something like
```php
if (!$this->data)
    $this->data = new SessionData(['session_id' => $id])
elseif ($this->data->session_id != $id)
    $this->data->session_id = $id;
```

greezybacon: Actually, never mind. The ORM does not support updating the primary key at the moment

PR #2989, opened 2016-03-29
orm: findAll bug

protich:
SqlCompiler::evaluate got called with swapped arguments

### DIFF
--- a/include/class.orm.php
+++ b/include/class.orm.php
@@ -1565,7 +1565,7 @@ extends CachedResultSet {
         foreach ($this as $record) {
             $matches = true;
             foreach ($criteria as $field=>$check) {
-                if (!SqlCompiler::evaluate($record, $field, $check)) {
+                if (!SqlCompiler::evaluate($record, $check, $field)) {
                     $matches = false;
                     break;
                 }


greezybacon:
Here's the definition of SqlCompiler::evaluate:
```php
    static function evaluate($record, $field, $check) {
        static $ops; if (!isset($ops)) { $ops = array(
            'exact' => function($a, $b) { return is_string($a) ? strcasecmp($a, $b) == 0 : $a == $b; },
            'isnull' => function($a, $b) { return is_null($a) == $b; },
            'gt' => function($a, $b) { return $a > $b; },
            'gte' => function($a, $b) { return $a >= $b; },
            'lt' => function($a, $b) { return $a < $b; },
            'lte' => function($a, $b) { return $a <= $b; },
            'contains' => function($a, $b) { return stripos($a, $b) !== false; },
            'startswith' => function($a, $b) { return stripos($a, $b) === 0; },
            'hasbit' => function($a, $b) { return $a & $b == $b; },
        ); }
        list($field, $path, $operator) = self::splitCriteria($field);
        if (!isset($ops[$operator]))
            throw new OrmException($operator.': Unsupported operator');

        if ($path)
            $record = $record->getByPath($path);
        // TODO: Support Q expressions
        return $ops[$operator]($record->get($field), $check);
    }
```
It seems like it should have been correct as it was written. What were the other problem symoptoms?

protich:
I agree, but it looks like the definition changed in Custom Queues PR.  This was a result of maddening debugging session at 2am.

greezybacon:
I agree. There is a bit of shoring up required to get the orm changes for the custom quest to play nicely with this code

PR #3003, opened 2016-03-31
files: Require authentication to view attachments

greezybacon:
This feature adds a setting to the control panel to require signing in to view attachments. This is in addition to the security already provided in the download URLs. Currently, download URLs are signed for a specific help desk, and automatically expire after about 24 hours. (The exact timing is the following GMT midnight allowing for at least 12 hours cache time.)

Administrators can impose this extra security feature to refuse serving attachment files if the user is not currently signed in. This could prevent third-party users from viewing an attachment if they were able to get access to the download URL before it expired.

**One caveat** is that the system cannot be sure if the unauthenticated user is an end user or an agent. Therefore, if the system cannot determine the user class, and end user is assumed and the client login page is displayed.

### DIFF
--- a/include/class.config.php
+++ b/include/class.config.php
@@ -207,6 +207,7 @@ class OsticketConfig extends Config {
         'agent_avatar' => 'gravatar.mm',
         'ticket_lock' => 2, // Lock on activity
         'max_open_tickets' => 0,
+        'files_req_auth' => 1,
     );
 
     function __construct($section=null) {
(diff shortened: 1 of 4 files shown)


include/class.config.php
```diff
@@ -207,6 +207,7 @@ class OsticketConfig extends Config {
         'agent_avatar' => 'gravatar.mm',
         'ticket_lock' => 2, // Lock on activity
         'max_open_tickets' => 0,
+        'files_req_auth' => 0,
```
protich: I think we should default to requiring authentication, lockdown by default seems to be the expected norm nowadays.

greezybacon: So I guess that officially retires the idea of sending attachment links for large attachments in emails?

PR #3004, opened 2016-03-31
Department  Access and Membership

protich:
This PR addresses a bug in department membership check and also adds a few enhancements.

__Todo__
- [ ] Rebase PR after #2996 is merged to drop use of `all()-asArray()`.

### DIFF
--- a/include/class.dept.php
+++ b/include/class.dept.php
@@ -305,10 +305,10 @@ implements TemplateVariable {
         if (is_object($staff))
             $staff = $staff->getId();
 
-        // Members are indexed by ID
-        $members = $this->getMembers();
+        $members = $this->getMembers() ?: $this->members;
 
-        return ($members && isset($members[$staff]));
+        return ($members->findFirst(array(
+                        'staff_id' => $staff)));
     }
 
     function isPublic() {
--- a/include/class.dept.php
+++ b/include/class.dept.php
@@ -51,6 +51,9 @@ implements TemplateVariable {
     );
 
     var $_members;
+    var $_primary_members;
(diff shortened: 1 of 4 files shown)


include/class.dept.php
```diff
@@ -205,6 +200,42 @@ function getAvailableMembers() {
...
+
+    function getExtendedMembers() {
+
+        if (!isset($this->_exended_members)) {
+            // We need a query set so we can sort the names
+            $members = StaffDeptAccess::objects();
+            $members->filter(array('dept_id' => $this->getId()));
+            $members = Staff::nsort($members, 'staff__');
+            $extended = array();
+            foreach($members->all()->asArray() as $member) {
```
greezybacon: You can iterate with `->all()` always. This issue was that it used to return an array and it was changed to return an iterator. #2996 converts it back to an array. Either way, asArray() wouldn't be necessary to iterate

include/class.dept.php
```diff
@@ -205,6 +200,42 @@ function getAvailableMembers() {
...
+                if (!$member->staff) continue;
+                // Annoted the staff model with alerts and role
+                $extended[] = new AnnotatedModel($member->staff, array(
```
greezybacon: I worked on a variation to the InstrumentedList in the custom queue feature to automate this trick. Perhaps we come up with a nice way to describe it in the meta data without duplicating code

protich: Perhaps expanded use of broker?

greezybacon: Yeah, it was a special broker subclass

PR #3006, opened 2016-03-31
Add FORCE_HTTPS option to the config

greezybacon:
If enabled, all HTTP requests will be redirected to their HTTPS equivalent. As currently written, the session cookie possibly disclosed over HTTP will also be removed to prevent possible session hijacking.


include/class.ostsession.php
```diff
@@ -90,6 +90,7 @@ function regenerate_id(){
         $oldId = session_id();
         session_regenerate_id();
         $this->backend->destroy($oldId);
+        $this::destroyCookie();
```
protich: `$this::` should be `self::`

greezybacon: Done.

PR #3014, opened 2016-04-05
Update ticket-view.inc.php

SDGPeteBatin:
Have attachments open in new window as default


jdelhome3578:
#2943 is a more direct fix

SDGPeteBatin:
Hey,

I did consider that fix, but it means changing the default behaviour of the method globally which may cause undesired effects elsewhere.  The ability to open attachments in a new window is there it's just not being utilised when the method is being called.

PR #3016, opened 2016-04-06
Fix invalid use of static variable in Config #2954

### LINKED ISSUE #2954: Invalid reset token
Hi, I've downloaded trunk version and staff password reset is not working at all, we always get an "invalid reset token" message. I found that the problem is related with class.config.php. Many changes were included there since the release candidate. I found that in function Items() in that class the variable $items was declared as static, and the problem was that it only gathers the "core" namespace config once, but it never loaded "pwreset" namespace and so it never found the reset token saved.

Changing the line "static $items" to $items=null in the function items() solved the issue.

Regards,
Gabriel.

quazardous:
Fix #2954


protich:
@quazardous - thank you for the PR. I've reimplemented it in #3022.

quazardous:
ok

PR #3022, opened 2016-04-08
Config  items cache

protich:
Make config items caching class instance based.


protich:
Reimplement commit cbea6d03 in #3016  to addresses issue #2954

PR #2995, opened 2016-03-29
forms: Fix attachment security merge conflict

greezybacon:
Is is just me, or did attachments stop working altogether?

Somehow the attachment security checks got duplicated and messed up. Perhaps it was a bad merge?

PR #3031, opened 2016-04-13
fixes issue #101

catbadger:
Fixes issue #101. Notes and responses in threads will be able to have file attachments.

### DIFF
--- a/include/class.thread.php
+++ b/include/class.thread.php
@@ -1398,8 +1398,21 @@ implements TemplateVariable {
             return false;
 
         /************* ATTACHMENTS *****************/
-
         //Upload/save attachments IF ANY
+        if ($vars['attach:response']) {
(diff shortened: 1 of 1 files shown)


greezybacon:
This ends up being a security violation the way you've written it. The code in `class.forms.php` checks to ensure that the references file-ids were actually uploaded by the submitting user in the current session. There was an issue fixed and merged (#2995) which caused attachments to be dropped. Did you check with that branch merged?

I've made a separate pull request, #3035, which fixes a secondary issue where duplicate INSERTs are sent to the database for attachments.

catbadger:
Does #3035 fix the attachment upload then?

greezybacon:
@catbadger perhaps you could test and report on that

catbadger:
Confirmed that #3035 fixes the broken attachment issue. this request is not needed if yours goes through.

PR #3054, opened 2016-04-20
session: Lock sessions on write to prevent thrashing

greezybacon:
#### History

osTicket v1.7 added support for database-backed sessions. This is great as it does not require any configuration, and fixes a number of security issues with PHP in general. However, it also dropped a major feature of the built-in session system: PHP blocks concurrent requests based on the session-data file. Therefore, only one request can be serviced for a given session at a time. The new session handling code allows concurrent requests to be served concurrently and will allow thrashing of the session data. That is one request can clobber session data written by another request. 

As an example, two requests for the same session are started at approximately the same time. They will fetch previously-saved session data from the database. Let's assume the content is the same for both requests. Then the first request makes changes and commits the session data back to the database. Then the second request does the same. Now the changes made by the first request are lost.

#### Implementation

The _SESSION variable is now a special ArrayObject derivative, which will monitor write access to itself. When writes are attempted, a lock is acquired based on the current session_id. If multiple requests are being handled on the same server for the same session, the subsequent requests will be blocked if they attempt to write to the session as well.

If a wait was required when acquiring the lock, the data is re-fetched from the storage backend and rebuilt into the array before allowing the write. This should help guard against write thrashing for concurrent requests, while also allowing read-only requests to proceed unblocked.

This commit also departs from the built-in session engine and handles the session serialization in the osTicket session class directly. This is necessary because it is not predictable when PHP will decide to commit the session, and PHP does not support the _SESSION variable from being anything except a plain array, which is not extensible.


greezybacon:
I added another part to the implementation in hopes of getting it stable. If the session lock was able to be acquired without waiting, then the session last-update time is fetched from the session backend. If it does not match the session last-update time of the fetched session data, then the session data is reloaded.

This will allow for timing of a situation like this (`L`ocked, `S`aved):
```
Request-1: |---------L===S-|
Request-2:     |------------------L===S-|
```

In this scenario, the first session will be able to acquire the lock without waiting and save its data. Then, the second request will be able to also acquire the lock without waiting, because it's not currently held; however, the session data was modified since it was fetched at the beginning of the request.

5a14d9c adds support for session last-update checking before issuing the lock. 0d24834 avoids touching the data in the backend if it was not locked (modified).

PR #3062, opened 2016-04-22
org: Remove users from deleted org.

protich:
Use org_id to filter users to update on organization deletion.

### DIFF
--- a/include/class.orm.php
+++ b/include/class.orm.php
@@ -229,6 +229,17 @@ class ModelMeta implements ArrayAccess {
         return $this->fields;
     }
 
+    function getByPath($path) {
+        if (is_string($path))
+            $path = explode('__', $path);
+        $root = $this;
+        foreach ($path as $P) {
+            list($root, ) = $root['joins'][$P]['fkey'];
+            $root = $root::getMeta();
+        }
+        return $root;
+    }
+
     /**
      * Create a new instance of the model, optionally hydrating it with the
      * given hash table. The constructor is not called, which leaves the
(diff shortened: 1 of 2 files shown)


include/class.organization.php
```diff
@@ -436,7 +436,7 @@ function delete() {
 
         // Remove users from this organization
         User::objects()
-            ->filter(array('org' => $this))
```
greezybacon: Does this not work? Seems like the ORM should be fixed if it doesn't

protich: It doesn't -- join to a just deleted org.

greezybacon: I see. Yeah, the join would fail. Perhaps the ORM should use the local side of the relationship if possible (automatically what your change proposes).

greezybacon:
```patch
diff --git a/include/class.orm.php b/include/class.orm.php
index 3e5d666..688017b 100644
--- a/include/class.orm.php
+++ b/include/class.orm.php
@@ -229,6 +229,15 @@ class ModelMeta implements ArrayAccess {
         return $this->fields;
     }

+    function getByPath($path) {
+        if (is_string($path))
+            $path = explode('__', $path);
+        $root = $this;
+        foreach ($path as $P)
+            $root = $root['joins'][$P];
+        return $root;
+    }
+
     /**
      * Create a new instance of the model, optionally hydrating it with the
      * given hash table. The constructor is not called, which leaves the
@@ -2246,9 +2255,21 @@ class SqlCompiler {
             // Handle relationship comparisons with model objects
             elseif ($value instanceof VerySimpleModel) {
                 $criteria = array();
-                foreach ($value->pk as $f=>$v) {
-                    $f = $field . '__' . $f;
-                    $criteria[$f] = $v;
+                // Avoid a join if possible. Use the local side of the
+                // relationship
+                if (count($value->pk) === 1) {
+                    $path = explode('__', $field);
+                    $relationship = array_pop($path);
+                    $lhs = $model::getMeta()->getByPath($path);
+                    $local = $lhs['joins'][$relationship]['local'];
+                    foreach ($value->pk as $f=>$v) {
+                        $criteria["{$field}__{$local}"] = $v;
+                    }
+                }
+                else {
+                    foreach ($value->pk as $f=>$v) {
+                        $criteria["{$field}__{$f}"] = $v;
+                    }
                 }
                 $filter[] = $this->compileQ(new Q($criteria), $model, $slot);
             }
```
Here's a possible fix for the ORM if you wanna try it too

protich: Added to the pull request

PR #3065, opened 2016-04-23
fix auto refresh rate options

manfer:
When auto refresh rate equals 1 you finish with the following sprinf code

`sprintf('Every minute', $i)`

which leads to sprintf warning when a staff member tries to edit his profile.

This fixes the problem.


greezybacon:
This will avoid translating the phrase properly. Perhaps we should just use `@` in front of the sprintf call

manfer:
I tested myself before creating the pull request to be sure the new code were translated properly. Anyway is fine if it is solved with @ in front of the sprintf.

greezybacon:
@manfer thanks for the confirmation

manfer:
I bet the issue I was facing was coming from the spanish translation where the plural translation for the string "Every minute" reads "cada %d %minutos" instead of the correct translation "Cada %d minutos".

I reported the bug in crowdin.

greezybacon:
@manfer good catch

PR #3082, opened 2016-04-28
Don't clean the config file, re-deploy rewritten files

### LINKED ISSUE #3025: deploy --clean removes the ost-config.php file
It looks like this was fixed a long time ago in #60, but I guess it's back.  I just upgraded 1.9.12 to 1.9.13 on my test server, and then did

    php setup/cli/manage.php deploy -Cv /var/www/helpdesk/

And it deleted `include/ost-config.php` and `include/.MANIFEST`

greezybacon:
With the advent of the MANIFEST file, files which are rewritten when deployed, such as those which have the GIT hash in the query string to force browser reloading, were no longer deployed unless the content of those files changed.

This patch adds a flag to the MANIFEST file to indicate that the file was rewritten when it was deployed. This allows the file to be deployed and rewritten again when the deployment is run, but also adds an indication to the console output to distinguish deploying changes as opposed to deploying rewrites.

It also improves the exclusion mechanism for cleaning so the .MANIFEST and ost-config.php files are not cleaned from the deployment target.

Maybe fixes #3025

### DIFF
--- a/include/cli/modules/deploy.php
+++ b/include/cli/modules/deploy.php
@@ -48,13 +48,14 @@ class Deployment extends Unpacker {
      * Removes files from the deployment location that no longer exist in
      * the local repository
      */
-    function clean($local, $destination, $recurse=0, $exclude=false) {
+    function clean($local, $destination, $root, $recurse=0, $exclude=false) {
         $dryrun = $this->getOption('dry-run', false);
         $verbose = $dryrun || $this->getOption('verbose');
         $destination = rtrim($destination, '/') . '/';
         $contents = glob($destination.'{,.}*', GLOB_BRACE|GLOB_NOSORT);
         foreach ($contents as $i=>$file) {
-            if ($this->exclude($exclude, $file))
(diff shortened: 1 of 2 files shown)


include/cli/modules/deploy.php
```diff
@@ -145,19 +146,28 @@ function getEditedContents($src) {
...
-    function copyFile($source, $dest, $hash=false, $mode=0644) {
-        $contents = $this->getEditedContents($source);
+    function isChanged($source, $hash=false) {
+        $local = str_replace($this->source.'/', '', $source);
+        $hash = $hash ?: $this->hashFile($source);
+        list($shash, $flag) = explode(':', $this->readManifest($local));
+        return ($flag === 'rewrite') ? $flag : $shash != $hash;
+    }
+
+    function copyFile($source, $dest, $hash=false, $mode=0644, $contents=false) {
```
greezybacon: The `$contents` parameter isn't used

PR #3090, opened 2016-05-02
orm: Delegate ResultSet calls to QuerySet Iterator

protich:
Allows for ability to access results without the need to call getIterator on the QuerySet. 

For example - ```$this->members->findFirst(array(...));```

This commit addresses regression issue brought about by commits 8ab4432f and eb0ba31 -- where ResultSet processing changed based on caching mode.

### DIFF
--- a/include/class.orm.php
+++ b/include/class.orm.php
@@ -1310,6 +1310,16 @@ class QuerySet implements IteratorAggregate, ArrayAccess, Serializable, Countabl
         unset($this->count);
     }
 
+    function __call($name, $args) {
+
+        if (!is_callable(array($this->getIterator(), $name)))
+            throw new OrmException('Call to undefined method QuerySet::'.$name);
+
+        return $args
+            ? call_user_func_array(array($this->getIterator(), $name), $args)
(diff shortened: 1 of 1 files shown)


greezybacon:
In your example, isn't `members` an `InstrumentedList`, not a `QuerySet`?

protich:
That was a bad example --  see `Dept::getPrimaryMembers()`, `Dept::getExtendedMembers()` and `Dept::getMembers()` - all return a query set that can be filtered further (`Dept::isMember()` is a good example)

trueserve:
This fixes a bug in 1.10-rc3 where agents, when viewing tickets that do not have an assignment, see no details for the ticket. The top ticket bar renders, but the page is otherwise blank. The issue is with the mentioned `Dept::isMember` when $members == this->getMembers(). I don't have any further details and don't know this codebase but can perform tests if it helps.

Relevant error log:
[Thu Jun 16 09:37:00.029879 2016] [:error] [pid 20921] [client 1.2.3.4:60793] PHP Fatal error:  Call to undefined method QuerySet::findFirst() in /webroot/osticket/include/class.dept.php on line 341

PR #3111, opened 2016-05-06
[Feature] allow template variables injection with Signal

quazardous:
Add a signal call in osTicket::replaceTemplateVariables() so plugins can inject custom template variables in emails.


quazardous:
now you can add template vars this way in your plugins

```php
...
        Signal::connect('replace.variables', function ($replacer, &$data) use ($ost, $config) {
            $data['my_custom_var'] = 'BOUYA !';
        });
...
```

PR #3116, opened 2016-05-09
Staff ticket details - Custom Form title

uziel200:
When a Ticket is created and contains a custom form, in the details is just appears the name of field with it respective value. It will be nice that show a division with the title of the custom form, and then the name and value of the field


ntozier:
Please close or delete if possible this PR (pull request)

If you are trying to open a bug report you do that here:
https://github.com/osTicket/osTicket/issues/new

PR #3120, opened 2016-05-10
Update ajax.users.php

ShaneOss:
Remove forward slash from $id variable and replace with backslash.


greezybacon:
@ShaneOss could you provide a use case for this patch so I could test it?

ShaneOss:
https://github.com/osTicket/osTicket-plugins/issues/29

PR #3153, opened 2016-05-27
List Item Preview

protich:
Add a preview tool tip to show list item properties on mouse-over.

**Admin Panel > List >> Items**
screen shot 2016-05-27 at 9 06 14 am

**Agent Panel > Ticket / Task View**
screen shot 2016-05-27 at 9 18 04 am


jdelhome3578:
:+1: I think I may have suggested that. I like it. Maybe an always show option on a property so that important properties show without a preview.

jdelhome3578:
Something here breaks the Ticket Status list.
`PHP Fatal error:  Call to undefined method TicketStatus::getListId() in C:\inetpub\osTicket\DPR\include\staff\templates\list-item-row.tmpl.php on line 18`

protich:
@jdelhome3578 - Just pushed a commit to add ticket status preview.  Thanks.

jdelhome3578:
@protich One other comment on this. I have list items that have properties that are pulling from another list which also has properties attached to the items. Any way to dig down on these layered items?

PR #3168, opened 2016-06-02
i18n: Ensure UTF-8 locale is set

### LINKED ISSUE #2910: [SOLVED] Emails with charset iso-8859-1 are cut (v1.9.8)
It's been a few months since I upgraded to 1.9.8, and while I had this issue in the beginning, it wasn't so important as it only happened when the staff sent emails with iso-8859-1 charset, so I told my partners to configure their email clients to send always in utf-8. 

However, recently this problem has come back with a few new customers, and I'd rather not tell them to do the same. Is there anything I cand do to fix it in the code? I'm attaching an email source example in case it helps. 

[emailSource_ISO_charset_problem.txt](https://github.com/osTicket/osTicket-1.8/files/117111/emailSource_ISO_charset_problem.txt)

I'm also attaching a screenshot of our system information:
osticket_system_info

Please let me know if you need more info.

greezybacon:
This is required for system requests such as email piping. Since a user is not logged in for the request, the system locale will not be set and text might get corrupted based on the PHP configuration.

Fixes #2910

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -290,6 +290,10 @@ class Bootstrap {
         }
         if (extension_loaded('iconv'))
             iconv_set_encoding('internal_encoding', 'UTF-8');
+
+        // Set the default locale to UTF-8. It will be changed by
+        // ::setLocaleForUser() later for web requests. See #2910
+        Translation::setLocale(LC_ALL, 'en_US.UTF-8');
     }
 
     function croak($message) {


balojs:
@greezybacon shows this error
Fatal error: Class 'Translation' not found in /var/www/solicitudes/bootstrap.php on line 296

protich:
@balojs - Please try PR #3184 - prior merge reverted. My apologies

balojs:
thanks @protich #3184 its works

PR #3175, opened 2016-06-06
Translation issues on homepage and kb page

rbukovansky:
Adding possibility to use translations for some texts on clients homepage and clients main knowledgebase page.


Chefkeks:
Interestingly I think I translated these phrases over on crowdin, but could be wrong and currently cannot take a look to verify.

Anyway as far as I can see the PR here looks correct. But for sure only @greezybacon knows ;)

Cheers
Michael

balojs:
+1

rbukovansky:
@Chefkeks Some of them are on crowdin, some are not. I believe we need to get crowdin source files for translations updated.

Issue #3194, opened 2016-06-12
1.10 RC3 - issue report - "Data Integrity"-Label (ticket form) is missing in field configuration window


mfelber:
When you configure a field for the ticket form and you are in the settings tab, you will see that this label is missing:

![required to close a thread](https://cloud.githubusercontent.com/assets/5539456/15993265/3ded64ae-30e0-11e6-9250-e06e04202447.jpg)

I guess this has something to do with the file dynamic-field-config.tmpl.php in include/staff/templates.

Line 102 seems to be the source of the problem:

`<div class="span4">
        <div style="margin-bottom:5px"><strong><?php __('Data Integrity');
    ?></strong>
        <i class="help-tip icon-question-sign"
            data-title="<?php echo __('Required to close a thread'); ?>"
            data-content="<?php echo __('Optionally, this field can prevent closing a thread until it has valid data.'); ?>"></i>
        </div>
    </div>`

This is the corresponding part for the Editable-Label to compare:

`<div class="span4">
        <div style="margin-bottom:5px"><strong>Editable</strong>
        <i class="help-tip icon-question-sign"
            data-content="<?php echo __('Fields marked editable allow agents and endusers to update the content of this field after the form entry has been created.'); ?>"
            data-title="<?php echo __('Editable'); ?>"></i>
        </div>
    </div>`

Normally the Label is between the strong tags, in this line there is no label, only this code "<?php __('Data Integrity'); ?>" .

By the way: I'm very excited because of the new releases from osticket.com: 1.10 RC3 and 1.9.14. 👍 
Thank you very much @greezybacon @protich !

mfelber:
Adding "echo" in line 102:

`<div style="margin-bottom:5px"><strong><?php echo __('Data Integrity');`

..seems to solving the problem:

![data integrity](https://cloud.githubusercontent.com/assets/5539456/16202953/47e9bef6-3718-11e6-9a56-c2c26a2a5813.jpg)

greezybacon:
@mfelber thanks for the hint. I fixed the issue directly

thuc960:
Thanks

PR #3195, opened 2016-06-12
Untested fix for #3194

Chefkeks:
Have not tested this fix, just watching Germany playing soccer :D but I think it shall fix https://github.com/osTicket/osTicket/issues/3194


Chefkeks:
Soccer and fixing osTicket isn't a good combo, closing this as it obviously won't fix #3194

PR #3196, opened 2016-06-13
i18n: Fix build language packs for the 1.10.x branch

greezybacon:
Crowdin requires a `branch` argument when requesting the language pack download from their system.


greezybacon:
My bad on my comments earlier. The second commit here properly changes the location of the language files to be from GitHub.

Issue #3142, opened 2016-05-19
Rewrite setup/install.php to twig


verschoof:
Hi guys,

Would you be interested if I implement twig for the install.php?
Why? I go use osTicket, since its open source I want to do something back.

This is my first step to write osTicket to Twig Templating Engine.
I will start in `/setup/`, why? becasuse is isolated so I thought I go start there, so we can do this step by step.

What do you guys think? And I would like to know from the repo maintainers if they would appreciate it.

verschoof:
@protich @greezybacon What do you guys think?

greezybacon:
@verschoof I would like to see the entire project move to Twig personally. If you're up for the work, I think starting in `/setup/` is a fairly good idea. Suggest your initial pull requests against the `develop-next` branch. That's where the new crazy features go in

philbertphotos:
How far you want to go with this ... and do you want help implementing it further in osticket...

verschoof:
I do wanna help implement this further in osTicket. I noticed that osTicket its pretty complex code while it don't has to be.

Personally, My idea for my self was to improve the setup dir, add twig, add controllers, maybe some services. and see how it ended up. Also to see how you guys would react on it, because those are pretty big changes.

My goal is to make the code more simple so it would be more "attractive" for people to help with OsTicket.

I don't have much time, but I wanna help with the project.

I'm fan of Symfony, and I love there components. There are many components we can use, HttpFoundation, Routing, ConsoleCommands etc =)

PR #3211, opened 2016-06-18
Feature/twig

verschoof:
- Implemented composer with autoload
- Implemented Twig
- Implemented Controller structure
- Fixed dryrun

It's a pretty big change with a little cleanup.
I do need some help to see if the deploy is still working, can somebody confirm that?

The easiest way to review this PR by commit

As discussed #3142 😉


include/cli/modules/deploy.php
```diff
@@ -277,8 +276,12 @@ function run($args, $options) {
                 "*/.htaccess", ".MANIFEST"));
         }
 
-        if (!$options['dry-run'])
-            $this->writeManifest($this->destination);
+        if (!$options['dry-run']) {
+            $this->writeManifest();
+
+            $this->clean($root, $this->destination, -1, array($include, "setup/"));
+            $this->clean("$root/include", $include, -1, array("ost-config.php","settings.php","plugins/", "*/.htaccess"));
```
greezybacon: There's an option for cleaning. IMO, since it could delete things an administrator might want to keep, it should be run explicitly with the option, perhaps with the `dry-run` flag as well, before permanently removing things

greezybacon:
@verschoof I'll be honest. I'm not crazy about the Symfony framework (beyond Twig). Specifically the controller part. I'm going to keep reading through your PR and read some more docs on Symfony, though.

If you're up for a continued challenge, one thing @protich has wanted for a long time is that the installer would no longer write to the config file directly. Instead, the installer would provide a link at the end of the installation wizard which would allow the admin to download the config file. The page displayed would have the instructions on how to upload the config file to the server, where to put it, and perhaps other information, like keeping a backup of the config file, making a note of the MySQL username and password used for osTicket, etc.

include/cli/modules/deploy.php
```diff
@@ -95,21 +95,19 @@ function clean($local, $destination, $root, $recurse=0, $exclude=false) {
         }
     }
 
-    function writeManifest($root) {
+    private function writeManifest() {
```
verschoof: > Why the change to `private` here
It is not used in other classes, the time you need this as extend lets see that case.

I prefer to make everything private if its not used, by doing that I know for sure its not used somewhere else (makes is less scary to edit the code!). Also, it forces you to think, maybe extend the class is not the answer but extract it from the class to reuse it.

So, if you wanna be "flexible" make sure you give your functions the right task and only that task. This way you can reuse it easily and you are flexible again! 😄 

But yea, that's my opinion and you guys are the repo leaders, so I go with what you want, but please, consider it.

include/cli/modules/deploy.php
verschoof: Hmm, I only added the dry run, maybe something went wrong with merging, I will take a look at this!

verschoof:
@greezybacon Symfony's framework is a different then its components. The framework is just the glue to put all the components together. (really awesome).

Before I will grab a next peace of code I want this PR been merged and working without any problems =)

Maybe chat sometime? I'm curious what you don't like about the controller part

verschoof:
@greezybacon is there more feedback? so I can go process it all at once =)

greezybacon:
Perhaps we should chat sometime about the controller bit. What you have is so far from where we are currently, that it makes a difficult case to merge and continue plowing ahead. Once something is merged, it makes going back very difficult. So, I had hoped to create "templates" like what we're doing in the `include/staffs and `include/client` folders for the setup, and move the template files to be twig based rather than PHP.

Currently the setup folder is recommended to be removed when the setup is completed, so that it cannot be used as an exploitable vulnerability, if there is one. But you seem to be suggesting we should move all our templates into a shared folder, which would mean they cannot be removed. Perhaps using Twig removes the vulnerabilities. Who knows.

So what you have changes the CLI deployment script, the auto load process, adds composer, the setup process has templates in a new location, the "resources/" folder, and the setup process itself is also completely changed to use this Symfony controller. And what's more, you're implying that I have to merge your code before you can add to it.

So I'm not really sure how to move forward. I assume I should spend some time with @protich and work on deciding if the direction you're suggesting is where we want the entire project to move. That is, if we want to use a consolidated templates path, to use composer for dependency resolution, to use the composer auto loader, to use a Symfony controller for the setup and perhaps other things, and Twig for a subset of the templates.

I would like to see a lot of those items. Don't get me wrong. I just have to decide exactly how we should implement them.

verschoof:
>  Symfony controller.

I didn't use a symfony controller, I just the MVC pattern concept (https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) Actually the VC, not the models 😜 

> And what's more, you're implying that I have to merge your code before you can add to it.

I'm not sure what you mean? I only asked if there is feedback so I can change that in my PR? Or do I miss understand you, Maybe a side not, I'm not a native english speaker (maybe I'm causing some confusion, let me know).

But I totally agree that you guys have to talk about the project structure, how to implement maybe I can join the conversation once?

greezybacon:
> Before I will grab a next peace of code I want this PR been merged and working without any problems =)

Maybe you meant something else. I read that you won't further the idea before this is merged.

> I didn't use a symfony controller, I just the MVC pattern concept 

I owe you an apology. I totally overlooked that. I will pull your code and read the setup controller code more closely.

On a deeper subject, I'm not a fan of the MVC pattern. The model and view parts make total sense, but the controller part usually becomes everything else that's needed to get things working together.  It's usually bloated and confusing. I'm trying to put together some concepts to migrate toward the [MOVE](https://cirw.in/blog/time-to-move-on) pattern. I think it makes more sense for software patterns. Specifically the Event and Operation pieces. I think they fit better in the software model. Your opinions are welcomed. Keep the conversation going.

verschoof:
>>Before I will grab a next peace of code I want this PR been merged and working without any problems =)

>Maybe you meant something else. I read that you won't further the idea before this is merged.

What I read was, you wanted to continue to another feature that is not related to my changes, and what I like is to work feature by feature. Or did I read it wrong?

I'm not sure what to think about the MOVE pattern, the move pattern is talking about 3 layers, and as re-usability I think that's not enough, and events? YES 👍 

Repository (model will be created here) -> service
The service is used to called from the controlers/console commands.

The controller is gathering information to give that to the view, same like the command.

BTW, i'm not sure if we should continue talking here so we go focus here about the PR it self.

Does osTicket has a channel at freenode IRC? (so I can jump in sometime). Or do you use slack?

verschoof:
@greezybacon Shall we go pick a date to talk? =)

Issue #3096, opened 2016-05-03
Unable to add FAQ after adding Category - v1.10-rc.2-376-g784184b


Taps7734:
osTicket Version v1.10-rc.2-376-g784184b (784184b) —  Up to date  
Web Server Software Apache/2.4.6 (CentOS) PHP/5.4.16 
MySQL Version 5.5.47 
PHP Version 5.4.16 

If I go to the page /scp/kb.php I can see that there are 0 (zero) categories in the drop down.  I also see 0 (zero) FAQ entries in that drop down.

Then go to /scp/categories.php and add a category.  The category is added correctly.

Then go back to /scp/kb.php to add a FAQ entry into the KB and I am no longer able to activate either of the drop downs, and I see the following in my apache error log:

`PHP Fatal error:  Call to a member function sort() on a non-object in /var/www/html/include/staff/faq-categories.inc.php on line 147`

rbukovansky:
I'm getting this as well...

rbukovansky:
BTW I have reverted from

146:     if (count($categories)) {
147:        $categories->sort(function($a) { return $a->getLocalName(); });

to this:

146:    if (count($categories)) {
147:        usort($categories, function($a, $b) {
148:            return strcmp($a->getLocalName(), $b->getLocalName());
149:        });

and it works.

balojs:
same problem :(

balojs:
thanks @rbukovansky it works

greezybacon:
I believe this is formally fixed in #3216

PR #3216, opened 2016-06-20
faq: Fix crash rendering list of categories

### LINKED ISSUE #3147: Fatal error  Call to a member function sort() on array
On **develop**, the page https://example.com/scp/kb.php gives
`Fatal error: Call to a member function sort() on array in /root_path/include/staff/faq-categories.inc.php on line 148`

I use a workaround like this:

```
- $categories->sort(function($a) { return $a->getLocalName(); });
```

```
+ sort($categories);
```

but I don't think it is the better way

Here is the config:
- osticket develop
- Debian Jessie
- Nginx

greezybacon:
Fixes #3147

### DIFF
--- a/include/staff/faq-categories.inc.php
+++ b/include/staff/faq-categories.inc.php
@@ -141,7 +141,7 @@ if($_REQUEST['q'] || $_REQUEST['cid'] || $_REQUEST['topicId']) { //Search.
 } else { //Category Listing.
     $categories = Category::objects()
         ->annotate(array('faq_count'=>SqlAggregate::COUNT('faqs')))
-        ->all();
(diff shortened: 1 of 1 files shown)


rbukovansky:
So, if my understanding is correct I can patch my osTicket installation with this and remove the changes I've made in #3096 ?

greezybacon:
@rbukovansky yes. This patch is a more direct fix

crypticmind:
FWIW 👍

jdelhome3578:
Already fixed with 624a4b0

rayfoss:
@jdelhome3578 thanks for the catch.

PR #3219, opened 2016-06-21
dates: MySQL 'GMT' includes BST

### LINKED ISSUE #2440: 1.10 Timezone issue / Time ahead of actual time (PHP 5.6.x, PHP 5.5.x related issue)
Hi @greezybacon and @protich 

not sure if either you noticed the following comments from the stabilize 1.10 PR so I thought would be best to open this separate issue report. It seems like time(zone) settings are broken/wrong with PHP 5.6.x. (using intl extension - have not tried without intl extension yet). Just look at the comments/links below:

Issue (screenshot shows 1.10-RC1, but still in 1.10-RC2)
https://github.com/osTicket/osTicket-1.8/pull/2056#issuecomment-124461437

Seems like it's related to PHP as @mfelber found out:
https://github.com/osTicket/osTicket-1.8/pull/2056#issuecomment-124475586
 
Michael

greezybacon:
The PHP 'GMT' timezone is equivalent to the 'UTC' timezone and therefore does not include summertime offsets. The 'GMT' timezone reported by MySQL, however, does use the BST summertime offsets. Therefore, the 'GMT' reported by MySQL should be explicitly translated to the 'UTC' timezone for use in PHP.

Maybe fixes #2440

### DIFF
--- a/include/class.timezone.php
+++ b/include/class.timezone.php
@@ -166,12 +166,20 @@ class DbTimezone {
         // Attempt to fetch timezone direct from the database
         $TZ = db_timezone();
 
+        // Translate ambiguous 'GMT' timezone
+        if ($TZ === 'GMT') {
+            // PHP assumes GMT == UTC, MySQL assumes GMT == Europe/London.
+            // To shore up the difference, assuming use of MySQL, use the
+            // timezone in PHP which honors BST (British Summer Time)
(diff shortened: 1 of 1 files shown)


robintoy:
Works a treat!!!
timeworks

PR #3220, opened 2016-06-21
Update cron.php

robertskiba:
On some Linux-Systems, you need to assign the interpreter to run the cron job correctly. Other systems just ignore the setting, but it may help many users with setting up the cron job correctly.


ntozier:
What would that do on windows?
Why not just follow the actual setup instructions here:
http://osticket.com/wiki/POP3/IMAP_Setting_Guide
instead of trying to fix a problem that doesn't seem to be a problem.

greezybacon:
This is a misunderstanding. Your cron should be set up like this:
```
* * * * * /usr/local/php /path/to/cron.php
```
I seriously doubt more than 2% of users have php installed in `/usr/local` (most would use `/usr/local/bin` or `/usr/bin` or `/opt/remi/php5.5/bin` or ... Hopefully you get the idea. There's no way to write the exact path for everyone at the top of this script.

PR #3237, opened 2016-06-25
forms: Fix subject default to help topic name

### LINKED ISSUE #3224: 1.10 RC3 ticket subject empty
Upgraded from 1.10 RC2 to 1.10 RC3, release afd63e1198c5c7d7f6dd6c9057f176d9b007242f

I've got a ticket type where I've unchecked many of the default form questions like "Issue Summary" which gets stored as the variable "subject".  Prior to RC3, the Subject of new tickets was set to the name of the ticket type, and now (since I do not ask the user to provide an Issue Summary) the Subject of all new tickets is empty.  I've confirmed that by re-enabling that form field, and entering a value, the Subject of the ticket is set to the user-entered value.

Would it be possible to have the Subject behave like it did in RC2, where if the user did not provide a value for Subject, it would default to the ticket type?

greezybacon:
Maybe fixes #3224

### DIFF
--- a/include/class.dynamic_forms.php
+++ b/include/class.dynamic_forms.php
@@ -954,8 +954,8 @@ class DynamicFormEntry extends VerySimpleModel {
         foreach ($this->getAnswers() as $ans) {
             $f = $ans->getField();
             if ($f->isStorable() && $f->get('name') == $name) {
-                $f->reset();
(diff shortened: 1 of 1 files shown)


protich:
See commit # 891a5a43 - addressed the same issue.

PR #3239, opened 2016-06-25
forms: Properly scrub email addresses

### LINKED ISSUE #1996: Validations on forms not permit "0" as value
the vahabior of validation on custom forms does not permit input a value of "0" (zero) , in osticket 1.9.4 or 1.9.5..

the realproblem its that i choose no validate any in the input field that i added and when goes to the ticket system and create a ticket , fill all input fields (including the custon field created without validation) , so then when click to create and save the ticket still marks that the fields that i already made custom and put "0" (zero) as value said required and cannot create any ticket

the custom input fiel has size 40 and max ling 50, validation none.. with those parameters when i input "0" raised a error, but for any other value does work..  
i thin this its incorrect due 1 or 0 are both values that i marked to do not validate

### LINKED ISSUE #2473: v1.9.5.1 - possible duplicate email address
I found this issue on v1.9.5.1

It is possible to enter the same email address in different format. E.g.

```
name.surname@domain.com
<name.surname@domain.com>
```

Both are accepted and stored to the database in this way, because email address uniqueness is done at DB level and the two strings are different, this makes possible to have two user with the same email address.

The format with bracket is also not shown on the user detail page.

greezybacon:
This reduces `<an@email.addr>` by removing the angle brackets. It also drops any other valid part of the email address which is not used by osTicket.

Maybe fixes #2473
Maybe fixes #1996

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -1314,11 +1314,7 @@ class TextboxField extends FormField {
                 }, __('Value does not match required pattern')
             ),
         );
-        // Support configuration forms, as well as GUI-based form fields
-        $valid = $this->get('validator');
-        if (!$valid) {
-            $valid = $config['validator'];
-        }
+        $valid = $this->getValidatorName();
         if (!$value || !isset($validators[$valid]))
             return;
         $func = $validators[$valid];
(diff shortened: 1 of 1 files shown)


include/class.forms.php
```diff
@@ -1330,7 +1326,29 @@ function($v) use ($config) {
...
+            $valid = $config['validator'];
+        }
+        return $valid;
+    }
+
     function parse($value) {
+        if ($this->getValidatorName() === 'email') {
+            // Strip everything but the core email address. It's already
+            // been through the validator, so this is just repeating some of
+            // the work. No need to be as cautious.
```
greezybacon: That's not true. This runs first, and the output is sent to the validator

ntozier:
@greezybacon you are officially -talking- writing to yourself. :)

include/class.forms.php
```diff
@@ -668,7 +668,7 @@ function validateEntry($value) {
 
         # Validates a user-input into an instance of this field on a dynamic
         # form
-        if ($this->get('required') && !$value && $this->hasData())
+        if ($this->get('required') && !strlen($value) && $this->hasData())
```
mckaygerhard: this fixeds the issue #1996 but i tried in a 1.9.5 version

Issue #2074, opened 2015-05-15
SQL schema utf8 -> utf8mb4


TinoDidriksen:
Any reference to MySQL's utf8 charset and collations should instead use the utf8mb4 versions. The old utf8 charset simply cannot store all Unicode code points, but the utf8mb4 versions can.

greezybacon:
@TinoDidriksen I like the way you think; however, `utf8mb4` isn't available until MySql 5.5. Currently osTicket requires 5.0. We'll have to bump the requirements first

PR #3244, opened 2016-06-25
thread: Add support for emoji characters

### LINKED ISSUE #1475: [1.9.4] Email is truncated due to special chars
sticket version 1.9.4

Problem: An incoming email is truncated since it contains special chars.

I have verfied this issue with this test email:

Original email with header:

```
From - Tue Nov 04 08:48:05 2014
X-Account-Key: account7
X-UIDL: UI
X-Mozilla-Status: 0000
X-Mozilla-Status2: 00000000
X-Mozilla-Keys:                                                                                 
Return-Path: <originalemail@example.org>
X-Original-To: ouremail@example.org
Delivered-To: ouremail@example.org
Received: [...]
Subject: Verifiy osticket issue
From: =?utf-8?Q?Tester?= <myemail@example.org>
Content-Type: text/plain;
    charset=utf-8
X-Mailer: iPhone Mail (12A405)
Message-Id: <id@example.org>
Date: Mon, 3 Nov 2014 21:43:18 +0100
To: "department@example.org" <department@example.org>
Content-Transfer-Encoding: quoted-printable
Mime-Version: 1.0 (1.0)
X-Provags-ID:  [...]
X-UI-Out-Filterresults: notjunk:1;

Hello,
this is just a test mail with a special char.=20=

Do you see a smile at the end of "this"=F0=9F=98=83 string after
```

Osticket truncate the email text after "this". So I guess
=F0=9F=98=83
cause this trouble.

greezybacon:
This adds emoji support to the ticket thread. This is especially useful for platforms which do not have built in emoji symbols. It also fixes the issue where `real_escape_string` converts emoji characters to querstion marks (`?`).

This requires MySQL 5.6 and manual conversion of the %thread_entry.body field to be collation utf8mb4.

Fixes #1475

### DIFF
--- a/README.md
+++ b/README.md
@@ -138,3 +138,4 @@ osTicket is supported by several magical open source projects including:
   * [php-gettext](https://launchpad.net/php-gettext/)
   * [phpseclib](http://phpseclib.sourceforge.net/)
   * [Spyc](http://github.com/mustangostang/spyc)
+  * [Twitter Emoji](https://github.com/twitter/twemoji)
--- a/include/mysqli.php
+++ b/include/mysqli.php
@@ -75,7 +75,7 @@ function db_connect($host, $user, $passwd, $options = array()) {
         'COLLATION_CONNECTION'  => 'utf8_general_ci',
         'SQL_MODE'              => '',
     ), 'session');
-    $__db->set_charset('utf8');
+    $__db->set_charset('utf8mb4');
 
     $__db->autocommit(true);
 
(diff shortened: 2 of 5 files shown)


mfelber:
Why does this require MySQL 5.6? As far as I know utf8mb4 was introduced with MySQL 5.5.3. Is there another reason? In my test environment with MySQL 5.5.47 I can't see any problems/errrors..

greezybacon:
My bad. I'm an idiot. 5.5.3 is sufficient. I didn't look that up before I
posted.

mfelber:
At least you knew it last year ;-)
https://github.com/osTicket/osTicket/issues/2074

A couple of days I made an issue report and noticed some time later that I did the same issue report a year before. Typical memory overflow..

rayfoss:
5.5 is eol as of earlier this month anyhow. What happens if someone tries to upgrade to this while still on 5.4? I'm guessing there is a minimum version variable somewhere that we need to specify...

greezybacon:
> This requires MySQL 5.5.3 and manual conversion of the %thread_entry.body field to be collation utf8mb4.

As stated, there is no _automatic_ migration. Because osTicket does not require MySQL 5.5 (or newer), there is no associated upgrade patch. That is, because this change requires a database migration, an upgrade patch would need to be created, and would need to be added to the install script. So, it will have to be part of a major version bump with database patches.

Should I chock up a migration patch against v1.10?

PR #3249, opened 2016-06-30
Make `cron` signals conform to documentation.

d--j:
The [signal documentation](https://github.com/osTicket/osTicket/blob/develop/setup/doc/signals.md) for the `cron` signal states that the context is `null`. But it was missing until now (and the `$data` parameter was wrongly forwarded as`$object`)

### DIFF
--- a/include/class.cron.php
+++ b/include/class.cron.php
@@ -107,7 +107,7 @@ class Cron {
         self::MaybeOptimizeTables();
 
         $data = array('autocron'=>false);
-        Signal::send('cron', $data);
+        Signal::send('cron', null, $data);
     }
 }
 ?>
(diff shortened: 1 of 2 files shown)


d--j:
I stumbled upon this inconsistency by accident. I did check the [implementation of `Signal`](https://github.com/osTicket/osTicket/blob/develop/include/class.signal.php#L89) and [other invocations of `Signal::send`](https://github.com/osTicket/osTicket/search?l=php&q=Signal%3A%3Asend&utf8=%E2%9C%93). They are consistent with the documentation.

The `Signal::send` method has three parameters with the third one being the `$data` array. Every other invocation of `Signal::send` does provide either  (a) three parameters or (b) two parameters where the second parameter is the context object or `null`. The only semantically wrong invocation of `Signal::send` are those two.

I did not test this change on an actual osTicket installation – if that's what you're asking.

Issue #3064, opened 2016-04-22
Suggestion - Add a custom.css file


Taps7734:
In my header.inc.php file I add a line each time for a custom.css file.

```
<link type="text/css" rel="stylesheet" href="<?php echo ROOT_PATH ?>scp/css/translatable.css?956ad7b"/>

<link type="text/css" rel="stylesheet" href="<?php echo ROOT_PATH ?>scp/css/custom.css?956ad7b"/>
```

Then I keep all of my customized CSS in this file so that I only have to update this one inc file each time I update.

Is it possible to add this line in the repo?  And possible a toggle somewhere under system for using it?  So that you can turn it on and off at will for testing?

Taps7734:
In addition to this, I would like to add that all hard-coded table width be removed from the HTML and allow it to be customized via CSS.

ntozier:
> In addition to this, I would like to add that all hard-coded table width be removed from the HTML and allow it to be customized via CSS.

The plan is to make the UI themeable.

Taps7734:
Even better.  I am getting ahead of myself then.

ntozier:
@Taps7734 I just made a PR (pull request) for a possible solution that I have just implemented on my test box (see above).  It doesn't add a toggle, but it allows you to add a custom css file.

netbrothers-tr:
I'm up for it. What's the status on this one?

PR #3259, opened 2016-07-06
New Feature: custom css include

ntozier:
This PR adds the ability for site admins to specify a custom css /path/to/css/file.css to override or supplement existing css in the Admin panel -> Setting -> System panel.

nst-pr-1

My hope is that this will at least temporarily make plugin devs have an easier time with being able to control what their plugin interface looks like. Originally I didn't include the client side, but decided that someone might want it there also. This should also be a solution for #3064


Chefkeks:
Even though I don't have tested this PR nor intend to use it, I have an idea/suggestion on how to improve it ;)

My suggestion is to define (hardcode) a custom css folder, like the plugin folder, where folks can put the css files and then select via a dropdown (with all css files of that css folder) instead of a text input field.

In case you want to do it really fancy, you could implement a multi-select dropdown, so the users/admins can select several files which will then be combined, so they'd be flexible and could keep css in several files instead of one file in case they like/want to do so.

Btw. also an Disabled or None option / value would be possible in the dropdown which nice too I think :)

Cheers,
Michael

ntozier:
I was initially going to do that (re: hardcode dir).... but decided that I wanted people to have the freedom to not have to move their .css files out of their plugin folders.  I'm all about plugins not needing to have to edit the core files to work and look the way their devs want them too.

I figured the first step was getting the concept into core, and then if people want to make it better-er (multi select drop downs, toggles to enable/disable, fancy AJAX popups, etc) they can do so.

ArthurBorsboom:
Hi,

Since there has been no commits or responses since July 2016 on the original osTicket project, I have decided to create a fork.

https://github.com/ArthurBorsboom/osTicket

There are many developers who have created awesome improvements by pull requests, which are eagerly waiting for merging.

This is an attempt to incorporate most pull requests to get create an improved version of osTicket. Hopefully this can later be reintegrated in the original project, or if the original project is dead as a rebranded project.

Would you mind to offer your pull request again to this fork?

Best regards,
Arthur Borsboom.

ntozier:
@rayfoss @JediKev @protich 
Any chance this might be accepted?

devosnw:
Bump. This feature would be extremely helpful for us.

PR #3262, opened 2016-07-08
export: Fix crash exporting users

greezybacon:


sistemasad:
Correct!, with those lines, the report is generated without issues.

Also -don't know if this have always be this way- only  NAME,ORG,MAIL,PHONE,NOTES and USERNAME are exported. Creation and modification dates, status, and others user information are not included.

I think that on this CSVs should be included all the columns, to be filtered later (while column selector is not implemented ;-))

Thank you.

Issue #3328, opened 2016-08-30
Showing FAQ results in fatal error


martin-rueegg:
When navigating to `/scp/kb.php` a fatal error is caused (probably depending on the data):

`Fatal error: Call to a member function sort() on array in /include/staff/faq-categories.inc.php on line 147`

ntozier:
Version of osTicket?

robintoy:
I would guess he means v1.10-rc.3 as we experience this and have jumped back to rc.2

martin-rueegg:
Sorry @ntozier, I forgot to post that information. Actually, I was running off the development branch (which might be a bad idea, of course). ;-)

patch will follow.

martin-rueegg:
btw: https://github.com/protich/osTicket/commit/0a06ab7810e76accd4745dda03680bd351b45c70 is a similar fix like mine #3331

martin-rueegg:
any comments on the fix? #3331 
ping @protich, @greezybacon

jdelhome3578:
Fixed #3216

PR #3329, opened 2016-08-30
Temp fix faq sorting #3328

martin-rueegg:
Disable sorting call avoids the error of #3328.


martin-rueegg:
Does include additional local commit. Needs amendment.

PR #3408, opened 2016-10-19
Adjust Subject Truncation in Ticket list

jdelhome3578:
Cell is actually 271 after padding is applied. Reduced space for paperclip to 15.


protich (commented):
Rationale for the adjustment?

jdelhome3578:
@protich I had some rows extending to two lines and the cell is actually 271 px not 279 px.

Issue #3029, opened 2016-04-12
.msg attachment behavior wonky (1.9.7 and 1.9.12)


ntozier:
Forum user fgunsili reported:

> We are using 0s-ticket ver.1.9.12 and it works fine. Most
> of the attachment of tickets works fine but only in one case it does not work.
> When we receive  attachment  as (*.msg ) it doesnt appears. We
> couldn’t solve it. What should we do ? Do you have any suggestions ?

source: http://www.osticket.com/forum/discussion/87672/can-not-receive-msg-attachments-in-osticket

I just tried this on my system and it doesn't work right for me either.

**Test 1**
add .msg to the extensions allowed.
Went to Outlook and Save As an email (with out an attachment)
Email ticket system (drag and dropped .msg into my email)
Ticket opens, with the .msg content added after my email text (and signature).
[note: maybe the system should add a separator before the content?]
**Result: Works.**

**Test 2**
add .msg to the extensions allowed.
Went to Outlook and Save As an email (with an attachment [.pdf])
Email ticket system (drag and dropped .msg into my email)
Ticket opens, with **NO** .msg content added.
I get a system message that reads: 

DATE time am          File Import Error
FILENAME:Unable to save file
**Result: Fails.**

Note 1: I was also under the impression that .msg was being phased out in favor of the .eml format.
Note 2: I tested on version 1.9.7 since that's the version my live sites run currently.

kest874:
This is still an issue in 1.10rc2 and rc3 with a .msg that has an attachment inside however it's throwing an error: 

>  Call to a member function logNote() on null line 917

![image](https://cloud.githubusercontent.com/assets/12860950/16819198/f79b3366-4910-11e6-9257-a868fcce0f0e.png)

then stops processing email, so every time cron runs it fetches the same mail throwing the error again.  

I've work around it for now with:

![image](https://cloud.githubusercontent.com/assets/12860950/16819284/52d54f00-4911-11e6-899e-d88501405d53.png)

the actual error is `string(33) "image001.png: Unable to save file"`

greezybacon:
If the `.msg` attachment MIME type is `application/tnef`, then the system will assume that Outlook has included it as a better format for the self-same email. This is because Outlook will send a text message body and a TNEF attachment which will include an HTML or rich text body, along with all the other attachments.

If, when attaching a .msg attachment, it will become a `application/tnef` attachment, then the system will not process it as a "regular" attachment, it will process it as **the entire email**, in a better format.

[...]

PR #3418, opened 2016-10-25
Save RFC822 Attachments

protich:
This pull request addresses cases where  .msg email attachments are not saved due to missing filenames (e.g issue #3029). 

The subject line of the attached message is used in such cases and attached messaged saved as .eml type file. 

- [x] Force file download when type is `message/rfc822` - necessary to open attachment in a mail client like OutLook
- [ ] Transcode encoding content stream from `UTF8` to `ascii` on download. Will require database patch to store target charset.


include/class.file.php
```diff
@@ -776,9 +791,13 @@ function read($amount=0, $offset=0) {
     /**
      * Convenience method to send all the file to standard output
      */
-    function passthru() {
```
Kuzuto: In PHP7 you need to change the original function with same parameters to prevent Warnings.
Line 967 needs to be changed to : passthru($filter=false) also.
Else, thank you for the fix, it works for me.

irsl:
Why is this not merged if it solves a known issue?

JediKev:
@irsl 

This pull is over two years old, also, we have a different way of dealing with this that should make it as it's own pull request soon.

Cheers.

PR #3426, opened 2016-10-26
Loading overlay

JediKev:
Use class instead of id on form element that triggers overlay on submit.


jdelhome3578:
@JediKev Shouldn't there be an overlay when posting a response or note on a ticket? I am not seeing that. I am seeing the overlay on the admin pages when saving settings.

PR #3430, opened 2016-10-27
Task duedate format

JediKev:
Fixes missing due date on task create caused by invalid datetime format sent to db.

**Pending**
- [ ] Invalid timezone conversion


jdelhome3578:
What about #3100

JediKev:
@jdelhome3578 that one looks similar but I see they have added some extra stuff that wasn't necessary so this adds just what we need to fix the problem.

protich:
Commit now part of PR #3438 - which addresses the pending item on this PR.

edukondalugorla:
Task Due date is displaying but can't edit due date. How to edit task due date

Issue #2414, opened 2015-08-14
Allowed Memory Size Exhausted in class.filter.php v.1.9.11


abssoft:
I have Ticket Filter that should prevent reply for ticket, come from no-reply@tld.com
![2015-08-14 08-58-33](https://cloud.githubusercontent.com/assets/946560/9265816/b02f880c-4262-11e5-9cab-41b465d4a084.png)

And after upgrading to 1.9.11 cron started to raise an Exeption "Allowed Memory Size Exhausted in class.filter.php:325"

Exeprion is raised at /include/class.filter.php:325

```
throw new FilterDataChanged($ticket);
```

catched by /include/class.ticket.php:2481 (function filterTicketData)

```
catch (FilterDataChanged $ex) {
        // Don't pass user recursively, assume the user has changed
        return self::filterTicketData($origin, $ex->getData(), $forms);
}
```

and started over again

seems that there is infinite recursion

justpusher:
I faced the same problem and found one of the reasons. 
In my case there was a saved user with lowercase email (abc@domain.com), and the received email has some of the letters uppercase (ABc@domain.com). So ticket email was first overwritten in    
 ```
Ticket::filterTicketData 
$vars['email'] = $user->getEmail(), 
```

then again overwritten in 

```
Filter::apply 
$ticket['email'] = $info['reply-to']
```

, and ticket considered changed, and filter applied again, and so on without the end. 
I suggest using case-insensitive comparison (strcasecmp) in Filter::apply

`$changed = 0 != strcasecmp($info['reply-to'], $ticket['email'])
                || ($info['reply-to-name'] && 0 != strcasecmp($ticket['name'], $info['reply-to-name']))`

JensEB:
Hello,

It isn´t a bug in PHP.
If a reply-to adress exists, the function Ticket::filterTicketData() produces a endless loop, because it´s a recursive funktion.
The function Filter::apply() checks only if $info['reply-to'] exists and it exist in every loop...

so I changed the checkup from: (osTicket 1.9.14 - ./include/class.filter.php line 318 function apply)
if ($this->useReplyToEmail()  && $info['reply-to']) {
to:
if ($this->useReplyToEmail()  && (strcasecmp($info['reply-to'], $ticket['email']) != 0)) {

Now it checks if reply-to is not to-adress. This fails in second loop, because to-adress is changed in reply-to-adress....

It seems to work. I hope, this helps you...

tmp2000:
[...]

PR #3437, opened 2016-11-01
Ticket Filter Memory Leak

protich:
Fix case where filter applying 'Use Reply-To Email' filter action triggers a reprocessing loop that exhausts php memory.


protich:
Addresses #3437 #2414

PR #3441, opened 2016-11-02
Add Remote User

jdelhome3578:
This will allow a remote user to be added from the lookup typehead.

### DIFF
--- a/include/ajax.thread.php
+++ b/include/ajax.thread.php
@@ -67,6 +67,8 @@ class ThreadAjaxAPI extends AjaxController {
     function addRemoteCollaborator($tid, $bk, $id) {
         global $thisstaff;
 
+        $id = str_replace('/','\\',$id);
+
         if (!($thread=Thread::lookup($tid))
                 || !($object=$thread->getObject())
                 || !$object->checkStaffPerm($thisstaff))
--- a/include/ajax.users.php
+++ b/include/ajax.users.php
@@ -304,6 +304,8 @@ class UsersAjaxAPI extends AjaxController {
     function addRemoteUser($bk, $id) {
         global $thisstaff;
 
+        $id = str_replace('/','\\',$id);
+
         if (!$thisstaff)
             Http::response(403, 'Login Required');
         elseif (!$thisstaff->hasPerm(User::PERM_CREATE))


protich:
@jdelhome3578  - Thanks for the PR. Can you please provide some context as to why the escaping is necessary?

jdelhome3578:
@protich osTicket/osTicket-plugins#29 might help. There might be some other solution but this works.

jdelhome3578:
@protich This fix is still needed

jdelhome3578:
@JediKev This still is an issue.

JediKev:
@jdelhome3578 

Just did some testing with v1.14.1 and OpenLDAP and cannot replicate this issue. I have a comma in the CN for the user in question and it populates the Add New User popup correctly, even on Ticket Create. If you could provide me with _exact_ steps to replicate that would be nice.

Cheers.

jdelhome3578:
@JediKev All I did was create a new ticket and looked up a user that has not been registered this is what I get in Chrome console as the request url and response without the fix:
Request URL: https://domain.name/scp/ajax.php/users/select/auth%3Aldap%3ACN%3Dlastname%5C%2C%20firstname%2COU%3DUsers%2COU%3DDepartment%20User%20Group%2CDC%3Ddomain%2CDC%3Dlocal
Request Method: GET
Status Code: 404 Not Found

This is an MS AD install.

JediKev:
@jdelhome3578 

I get the same request URL as well but mine works flawlessly:
- https://domain.name/scp/ajax.php/users/select/auth%3Aldap%3Acn%3DLast%5C2C%20First%2Cou%3DUsers%2Cdc%3DIP.ADDRESS.HERE

This is my User in OpenLDAP:
Screen Shot 2020-04-14 at 10 22 10

This is the lookup on Ticket Create:
Screen Shot 2020-04-14 at 10 22 28

After selecting the User the fields are populated:
Screen Shot 2020-04-14 at 10 22 37

After clicking Add User they are added properly:
Screen Shot 2020-04-14 at 10 22 47

Cheers.

JediKev:
@jdelhome3578 

Also, your URL params translate to `auth:ldap:CN=lastname\, firstname,OU=Users,OU=Department User Group,DC=domain,DC=local` which has no `/` to replace. So I don't see how these changes help in your case.

```php
echo str_replace(
    '/',
    '\\',
    'auth:ldap:CN=lastname\, firstname,OU=Users,OU=Department User Group,DC=domain,DC=local');

// result: auth:ldap:CN=lastname\, firstname,OU=Users,OU=Department User Group,DC=domain,DC=local
```

Cheers.

jdelhome3578:
@JediKev Not sure why but it fixes the issue. Here is some more info osTicket/osTicket-plugins#29 might help

JediKev:
@jdelhome3578

I looked at that but it gave me virtually no info to replicate the issue..only that the comma breaks it which I confirmed is not the case. Can you retest with latest `develop-next`?

Cheers.

jdelhome3578:
@JediKev So I changed the 404 response to include the $id and it shows that the slash is being changed to `/`
```
        elseif (!($backend = AuthenticationBackend::getSearchDirectoryBackend($bk))
                || !($user_info = $backend->lookup($id)))
            Http::response(404, $id);
```

JediKev:
@jdelhom3578

Hmmm...when I `var_dump($id);` in the `addRemoteUser()` method I see the normal backslash slash on my install. I will have to look at what could be changing the slash and go from there. Thanks for this helpful info btw.

Cheers.

jdelhome3578:
@JediKev I am using Chrome and IIS.

PR #3450, opened 2016-11-02
Feature :: Auto Close for v1.10

jdelhome3578:
This is a rebase PR for the Auto Close feature #2917. This PR will work with v1.10 and also allows for admin to select the status from which will be used to automatically close tickets.

Need help:
- [ ] set status using status name instead of id in AutoClose function
    `class.ticket.php` line 3641 `$ticket->setStatus('3', 'Ticket Closed by the SYSTEM after '.$grace.' '.$plural.' of no activity.',$errors, false);`

image

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -3615,5 +3615,29 @@ implements RestrictedAccess, Threadable {
 
         require STAFFINC_DIR.'templates/tickets-actions.tmpl.php';
     }
+   function AutoClose() {
+        global $cfg;
+        
+        $grace = $cfg->getAutoCloseDuration();
+        $graceStatusId = $cfg->getAutoCloseStatusId();
+        
+        if($graceStatusId != 0){ 
+            if($grace == 1){
+                $plural = 'hour';
+                } else {
+                $plural = 'hours';
+            }
+            // select all tickets marked as selected status from config page where updated is older than ($grace) hours ago
+            $sql  = 'SELECT ticket_id FROM '.TICKET_TABLE.' ticket '
+                    .' INNER JOIN '.TICKET_STATUS_TABLE.' status ON (status.id=ticket.status_id AND status_id = '. $graceStatusId .') '
+                    .' AND TIME_TO_SEC(TIMEDIFF(NOW(),ticket.updated))>='.$grace.'*3600';
+            if(($res=db_query($sql)) && db_num_rows($res)) {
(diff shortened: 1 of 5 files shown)


clonemeagain:
No way! This would have saved some time, had no idea. Actually wrote a plugin to do it:
https://github.com/clonemeagain/plugin-autocloser

evx:
I have never seen an Auto Close function in version 1.9 or 1.10 release. Anyone know if Auto Close will be a part of future version?

Thanks **clonemeagain**, I will try your plugin instead.

ntozier:
I believe that the answer is no.  Because this is the type of thing that is better off being a plugin.

include/class.ticket.php
```diff
@@ -3615,5 +3615,33 @@ static function agentActions($agent, $options=array()) {
...
+        
+        if($graceStatusId != 0){ 
+            if($grace == 1){
+                $plural = 'hour';
+                } else {
+                $plural = 'hours';
+            }
+            
+            // select all tickets marked as selected status from config page where updated is older than ($grace) hours ago
+            $sql  = 'SELECT ticket_id FROM '.TICKET_TABLE.' ticket '
```
jdelhome3578, line 3634: @protich could you help convert this to ORM?
This is what I think it might be:
```php 
       $autoclose = static::objects()
            ->filter(array(
                'status_id' => $graceStatusId,
                    Q::all(array(
                        'updated__gt' => $grace*3600)
                        )
                ))
            ->limit(100);

        foreach ($autoclose as $ticket)
            $ticket->setStatus('3', 'Ticket Closed by the SYSTEM after '.$grace.' '.$plural.' of no activity.',$errors, false);
```

protich: Sorry for a late response -  it would be something like below - assuming grace period is in hours.

```php
      $tickets = static::objects()
            ->filter(array(
                'status_id' => $graceStatusId,
                'updated__lt' => SqlFunction::NOW()->minus(SqlInterval::HOUR($grace))
             ))
             ->limit(100);
```

Totally unrested - ;)

jdelhome3578: @protich Thanks for the input. I will test. One other question setstatus() does not accept "closed" as input, it only accepts the ID number for the status. Is this right?

jdelhome3578:
Closed in favor of #5108

PR #3453, opened 2016-11-03
ui: improve action buttons

jdelhome3578:
This PR will make the whole action button clickable as opposed to only the icon.

When viewing the queue
image

When viewing a Ticket
image


jdelhome3578:
Re-based for the updated code. PR #5454

PR #3469, opened 2016-11-05
Make FAQ & Pages Attachments Viewable

protich:
Commit c4579277 introduced an extra administrative security feature to restrict files access to signed in users only, even  if a user has a valid & signed download URL. The feature, however, did not take into account public images & files associated with FAQs and pages such as landing/thank-you pages.

This PR addresses the shortcoming by adding a reference ID (attachment ID) to the download/access URL, that can be used to deduce the model/object type that the file request is associated with. The technique will allow us in the future to enforce ACL at the file level depending on privacy settings and
the security clearance of the user (agent).

### DIFF
--- a/file.php
+++ b/file.php
@@ -26,20 +26,33 @@ if (!$_GET['key']
     Http::response(404, __('Unknown or invalid file'));
 }
 
-// Enforce security settings
-if ($cfg->isAuthRequiredForFiles() && !$thisclient) {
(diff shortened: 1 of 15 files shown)


Bram0194:
I replaced all the 15 files, restarted apache and re-added the image to the offline page but it still doesn't show up correclty.

Issue #2207, opened 2015-07-02
advanced ticket search problem osticket 1.10


zizackov:
Hi
when trying to do a search for ticket between 2 dates, i get no results although there are tickets in this date range. other advanced searches work fine.  
thank you, 
![osticket search](https://cloud.githubusercontent.com/assets/13097295/8470259/fbdeb30c-2090-11e5-8654-22606a03b8d2.jpg)

greezybacon:
Thanks for the bulliten. I've drafted a fix for this, but haven't put a pull request together. I'll reference your issue when I post it

SpudR:
Just to add to this - the searches for "in the last n days" and "in the next n days" don't work properly.
If I enter 365 days I just get 0 results though there are tickets in both ranges.
Also, what is the difference between these searches and the "After [date]" search?

PR #3470, opened 2016-11-06
Fix date range search (issue #2207)

jamesangi:


t-oster:
+1. Works for me.

jamesangi:
@rayfoss I agree, updated accordingly

jamesangi:
I just pushed another commit into my copy of osTicket intending to do a separate pull request but I see that it has automatically put it here on this pull request.  Please excuse my git ignorance, I'm not quite sure how to keep it separate.  Anyway, as the commit line states, it's a hack, but at least ndays searching works now.  I would appreciate any suggestions for cleaner code.

PR #3478, opened 2016-11-07
Fixes typos and lack of localization function calls

emansije:


emansije:
You're welcome. :-)
Glad to contribute.

Issue #3443, opened 2016-11-02
[Feature Request] Manage -> Plugins should display plugin version (1.10)


ntozier:
When you install a new plugin in osTicket it displays the version of the plugin that you are trying to install.
(aka when you go do Admin panel -> Manage -> Plugins -> Add New Plugin)

However after its installed when you go to Admin panel -> Manage -> Plugins there is no column for nor a display of the version of the plugin installed.   Preferably this should be displayed here (but I would settle for when you click on the installed plugin)

Rational: this will allow people to quickly and easily see what version of the plugin they have installed so that when they upgrade they can check and see if a new version of the plugin is available.

Chefkeks:
I second that idea :+1:

PR #3486, opened 2016-11-08
Update plugins.inc.php

ntozier:
Add plugin version column and version# to /scp/plugins.php

This takes care of feature request https://github.com/osTicket/osTicket/issues/3443

### DIFF
--- a/include/staff/plugins.inc.php
+++ b/include/staff/plugins.inc.php
@@ -56,7 +56,7 @@ $showing=$pageNav->showing().' '._N('plugin', 'plugins', $count);
         <tr>
             <th width="4%">&nbsp;</th>
             <th width="56%"><?php echo __('Plugin Name'); ?></th>
-			<th width="10%"><?php echo __('Version'); ?></th>
+	    <th width="10%"><?php echo __('Version'); ?></th>
             <th width="10%"><?php echo __('Status'); ?></th>
             <th width="20%"><?php echo __('Date Installed'); ?></th>
         </tr>
@@ -65,12 +65,12 @@ $showing=$pageNav->showing().' '._N('plugin', 'plugins', $count);
 <?php
 foreach ($ost->plugins->allInstalled() as $p) {
     if ($p instanceof Plugin) {
-			$ppath = $p->getInstallPath();
-			if($p->isPhar()) {
-				$info = include 'phar://'.INCLUDE_DIR.'/'.$ppath.'/plugin.php';
-			} else {
-				$info = include($ppath.'/plugin.php');
-			}
+	$ppath = $p->getInstallPath();
+	if($p->isPhar()) {
+		$info = include 'phar://'.INCLUDE_DIR.'/'.$ppath.'/plugin.php';
+	} else {
+		$info = include($ppath.'/plugin.php');
+	}
 	?>
     <tr>
         <td align="center"><input type="checkbox" class="ckb" name="ids[]" value="<?php echo $p->getId(); ?>"
--- a/include/staff/plugins.inc.php
+++ b/include/staff/plugins.inc.php
@@ -74,8 +74,7 @@ foreach ($ost->plugins->allInstalled() as $p) {
     <tr>
         <td align="center"><input type="checkbox" class="ckb" name="ids[]" value="<?php echo $p->getId(); ?>"
                 <?php echo $sel?'checked="checked"':''; ?>></td>
-        <td><a href="plugins.php?id=<?php echo $p->getId(); ?>"
-            >
+        <td><a href="plugins.php?id=<?php echo $p->getId(); ?>">
 	<?php echo $p->getName(); ?></a></td>
 	<td><?php echo $info['version']; ?></a></td>
         <td><?php echo ($p->isActive())
--- a/include/staff/plugins.inc.php
+++ b/include/staff/plugins.inc.php
@@ -55,7 +55,8 @@ $showing=$pageNav->showing().' '._N('plugin', 'plugins', $count);
     <thead>
         <tr>
             <th width="4%">&nbsp;</th>
-            <th width="66%"><?php echo __('Plugin Name'); ?></th>
+            <th width="56%"><?php echo __('Plugin Name'); ?></th>
+			<th width="10%"><?php echo __('Version'); ?></th>
             <th width="10%"><?php echo __('Status'); ?></th>
             <th width="20%"><?php echo __('Date Installed'); ?></th>
         </tr>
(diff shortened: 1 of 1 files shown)


ntozier:
This is what it looks like:

mockup

jdelhome3578 (changes requested):
Could you clean up the indenting of the code.

jdelhome3578 (changes requested):
Line 91 needs colspan="5"

JediKev (approved):
This looks nice! Great job! 👍

FossPrime (changes requested):
push a commit if you figure out the date not showing/ missing td issue

ntozier:
1. The display code wasn't touched in this PR so the issue has nothing to do with this pull request. 
2. reverting the code to stock does not fix the issue so the issue has nothing to do with this pull request.
3. Removing "Format::datetime()" from around the date makes the date re-appear so I would say the problem probably lies with that function, which is well outside of the scope of this PR.  This PR was authored against 1.10rc3 tree where the dates worked.  It does not work in 1.10 final.

include/staff/plugins.inc.php
```diff
@@ -55,20 +55,29 @@
...
+		$info = include($ppath.'/plugin.php');
+	} ?>
     <tr>
         <td align="center"><input type="checkbox" class="ckb" name="ids[]" value="<?php echo $p->getId(); ?>"
                 <?php echo $sel?'checked="checked"':''; ?>></td>
         <td><a href="plugins.php?id=<?php echo $p->getId(); ?>"
-            ><?php echo $p->getName(); ?></a></td>
+            >
+	<?php echo $p->getName(); ?></a></td>
+	<td><?php echo $info['version']; ?></a></td>
```
rayfoss: dangling anchor tag

rayfoss:
We can't reproduce the issue with date being missing at all. I tried with and without this on develop and 1.10. It must be something completely unrelated. Looks good aside from the minor detail above.

ntozier:
maybe its a PHP version thing?
I see nothing in my PHP error logs.

rayfoss:
@ntozier Shouldn't be... Kevin runs 5.6, I run 7.0

ntozier:
BTW the date thing ended up being an PHP IIS thing.  Changing the Date and Time Format: settings in Admin panel -> Settings -> System fixed the issue.

ntozier:
change requested by ray done.

JediKev:
@ntozier 

Reviewing this again, I would suggest the following changes:
```
diff --git a/include/class.plugin.php b/include/class.plugin.php
index aedc065..1b75e48 100644
--- a/include/class.plugin.php
+++ b/include/class.plugin.php
@@ -334,6 +334,8 @@ class PluginManager {
             .', install_path='.db_input($path)
             .', name='.db_input($info['name'])
             .', isphar='.db_input($is_phar);
+        if ($info['version'])
+            $sql.=', version='.db_input($info['version']);
         if (!db_query($sql) || !db_affected_rows())
             return false;
         static::clearCache();
@@ -388,6 +390,7 @@ abstract class Plugin {
     function getName() { return $this->__($this->info['name']); }
     function isActive() { return $this->ht['isactive']; }
     function isPhar() { return $this->ht['isphar']; }
+    function getVersion() { return $this->info['version'] ?: $this->ht['version']; }
     function getInstallDate() { return $this->ht['installed']; }
     function getInstallPath() { return $this->ht['install_path']; }
 
diff --git a/include/staff/plugins.inc.php b/include/staff/plugins.inc.php
index 7f21024..b182f6e 100644
--- a/include/staff/plugins.inc.php
+++ b/include/staff/plugins.inc.php
@@ -55,7 +55,8 @@ $showing=$pageNav->showing().' '._N('plugin', 'plugins', $count);
     <thead>
         <tr>
             <th width="4%">&nbsp;</th>
-            <th width="66%"><?php echo __('Plugin Name'); ?></th>
+            <th width="56%"><?php echo __('Plugin Name'); ?></th>
+            <th width="10%"><?php echo __('Version'); ?></th>
             <th width="10%"><?php echo __('Status'); ?></th>
             <th width="20%"><?php echo __('Date Installed'); ?></th>
         </tr>
@@ -67,8 +68,9 @@ foreach ($ost->plugins->allInstalled() as $p) {
     <tr>
         <td align="center"><input type="checkbox" class="ckb" name="ids[]" value="<?php echo $p->getId(); ?>"
                 <?php echo $sel?'checked="checked"':''; ?>></td>
-        <td><a href="plugins.php?id=<?php echo $p->getId(); ?>"
-            ><?php echo $p->getName(); ?></a></td>
+        <td><a href="plugins.php?id=<?php echo $p->getId(); ?>">
+        <?php echo $p->getName(); ?></a></td>
+        <td><?php echo $p->getVersion(); ?></a></td>
         <td><?php echo ($p->isActive())
             ? 'Enabled' : '<strong>Disabled</strong>'; ?></td>
         <td><?php echo Format::datetime($p->getInstallDate()); ?></td>
@@ -78,7 +80,7 @@ foreach ($ost->plugins->allInstalled() as $p) {
     </tbody>
     <tfoot>
      <tr>
-        <td colspan="4">
+        <td colspan="5">
             <?php if($count){ ?>
             <?php echo __('Select'); ?>:&nbsp;
             <a id="selectAll" href="#ckb"><?php echo __('All'); ?></a>&nbsp;&nbsp;
```

Note that instead of using this logic:
```
$ppath = $p->getInstallPath();
if($p->isPhar()) {
    $info = include 'phar://'.INCLUDE_DIR.'/'.$ppath.'/plugin.php';
} else {
    $info = include($ppath.'/plugin.php');
}
```
I'm using a new method `$p->getVersion()`. The reason for this is so we don't have to open/load each PHAR Archive just to get the version number which can be very expensive. Instead, `getVersion()` will use the version value provided by the `Plugin::load()` function or it will use the `version` value from the database. Previously installed plugins will not have a `version` value in the db (as we never inserted one, oops) so this is why I'm using `$this->info['version'] ?: $this->ht['version']`. My suggested changes also includes an insert statement for `version` when installing plugins so we insert a version from now on.

With this being said, you can either:
- Rebase your branch, apply these changes, and force push to update this pull
- Delete the branch, close this pull, create a new branch, apply the changes, and make a new pull
- Or I can submit a pull with the changes and with you being the author of the pull

You can do whatever, it's really up to you.

Cheers.

ntozier:
Yeah with out the version in the db i didnt see any other way to get that info. :)

Option 3 please. :)

JediKev:
**Note:**
Rewrite is located [here](https://github.com/osTicket/osTicket/pull/5357).

PR #3495, opened 2016-11-10
ui: Redirect advanced search to new page

jdelhome3578:


t-oster:
Makes it much better. Search is one of the main problems in my opinion.

PR #3522, opened 2016-11-21
Client Side Long FAQ Title

JediKev:
This addresses issue where long FAQ titles as links don't line up with icons.


ntozier:
I can confirm that this fixes the issue.

PR #3545, opened 2016-12-03
assets insecure logo fixed

meshr-net:
cloudflare https causes issues here: mixed http and https content


rayfoss:
@meshr-net could you describe the issue a little more this doesn't seem like it should be a problem, the redirect is relative and resolves to https if you're using that already

PR #3557, opened 2016-12-14
Knowledgebase Category language

gcasas84:
If you have (Knowledgebase) and select a category, always see category in default language, not in selected language, with this change now it works fine


JediKev:
Good catch.

PR #3575, opened 2016-12-23
Update for compatibility w/ PHP 7, E_STRICT

heisian:
Many method declarations were not matching their parents & triggered warnings. Added matching variables defaulting to `null`.

PHP 7 also introduces a built-in `Error` class which triggered a Fatal Error. Renaming osTicket's Error class to ErrorOs seems to fix, but haven't yet tested everything.

### DIFF
--- a/WHATSNEW.md
+++ b/WHATSNEW.md
@@ -1,3 +1,29 @@
+osTicket v1.9.13
+================
+### Enhancements
+  * Help topic can be specified by the ID number in the URL for client new
+    ticket page (#2735)
+
+### Improvements
+  * Fix crash requesting registration email as a guest, thanks @bailey86
+    (#2552)
+  * Fix attachment filename encoding (#2586)
+  * Fix bounce message loop for message alert to a bad agent email address
+    (#2639)
+  * Sort help topic names case insensitively (#2350)
+  * Fix redactor toolbar appearing over the overlay (#2697)
+  * Add help tip for primary role, thanks @colonelpopcorn (#2680)
+  * Add icons to assigned-to column, thanks @antriver (#2695)
+  * Upgrade to htmLawed 1.20 (#2935)
+  * Fix stripping of `src` attribute in `iframe` elements (#2940)
+
+### Performance and Security
+  * Reduce memory usage processing attachments, thanks @ericLemanissier (#2491,
+    #2492)
+  * Protect access to files shown in the FileUpload field (#2618)
+  * Always perform validation server-side for ajax uploads (e3c9e0f)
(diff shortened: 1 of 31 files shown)


greezybacon:
@heisian @rayfoss I think there might be some confusion here. The `develop` branch already has the E_STRICT errors resolved. Because addressing the PHP7 issues required significant work to the ORM and other major system components, it was decided not to backport the changes to the 1.9.x branch. Therefore, either this request should remain on the `1.9.x` branch, or should attempt to fix one of the remaining PHP7 E_WARNING emits from the `develop` branch.

4nanook:
How do I pull a copy of the development branch?  1.10.x still is not
working right with PHP7.  When I create a new ticket, there is no box to type
text into.

-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-
  Eskimo North Linux Friendly Internet Access, Shell Accounts, and Hosting.
    Knowledgeable human assistance, not telephone trees or script readers.
  See our web site: http://www.eskimo.com/ (206) 812-0051 or (800) 246-6874.

heisian:
@greezybacon Thanks you are right - I attempting pulling this branch for use instead of looking at the latest stable version. It was a bit hard to find..

Will close this request as its not relevant any longer.

@4nanook it sounds like the issue you're having might be related to NGINX .conf? I had a similar issue of no form being loaded on a new ticket & using NGINX.

4nanook:
No, I am using Apache 2.4.25 and I am sure it is related to PHP7 as
OS Ticket worked fine under PHP 5.6 but failed when I upgraded PHP to PHP7.

      I am not going to downgrade PHP because it resulted in about a 40%
speed improvement of my Wordpress sites.  And it's been out a year so really
there is no excuse for anything not to be compliant at this point.

-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-_-
  Eskimo North Linux Friendly Internet Access, Shell Accounts, and Hosting.
    Knowledgeable human assistance, not telephone trees or script readers.
  See our web site: http://www.eskimo.com/ (206) 812-0051 or (800) 246-6874.

ntozier:
You can run more then one version of PHP at the same time on the server.
Maybe you should install 5,6 and use that for the osTicket site until this gets sorted.

heisian:
it does seem to work guys, at least for me. php 7 using this branch: https://github.com/osTicket/osTicket/tree/v1.10

PR #3586, opened 2017-01-02
Add Custom Forms to Ticket Filter Data

protich:
Include custom fields data associated with custom forms attached to a user or an organization to filter data available on ticket creation.


RipperHead:
This exist now?

JediKev:
As you see this has not been merged into the codebase, so no.

PR #3596, opened 2017-01-05
Feature/travis

FossPrime:
This is using a dockerfile to allow running on other CI's and better mimicking osTicket environments. In the future using a docker-compose.yml to run on several different php versions would be ideal.

The syntax check and undefined check take 30 minutes+ each with imap, so I disabled it's pecl ext and xdebug.

Note everything should fail until we address the 8 issues that currently do not pass. Will not affect anything technical.


rayfoss:
@protich please see. Now passes without anything fatal 
screenshot from 2017-03-17 15-09-35

Issue #3620, opened 2017-01-17
Initial form field flags saved to database incorrectly when installing under PHP 7.


dellsala:
### Description

After a fresh installation, the client open ticket form appears to be missing fields. Under the admin panel, under Manage -> Forms, the visibility of all fields is listed as "Disabled". In the database, all `ost_form_field`.`flags` values are set to 0.

### Steps to Reproduce

1. Run a fresh install of OSTicket 1.10 under PHP 7.
2. Browse to client "Open Ticket" form.

**Expected behavior:** Expect to see a working open ticket form with contact fields.

**Actual behavior:** No contact fields appear in the form. The ticket fields do not appear after selecting a help topic. Form field visibilities are listed as "Disabled" in the form manager under the admin panel.

![screen shot 2017-01-17 at 8 48 25 am](https://cloud.githubusercontent.com/assets/977797/22023367/db73f436-dc93-11e6-9a57-1da8671ffc80.png)

### Versions

osTicket: v1.10 (901e5ea)
PHP: 7.0.8
OS: Ubuntu 16.04
MySQL: 5.7.16

dellsala:
I'm pretty sure this is related to a change in string handling in PHP 7.

[Hexadecimal strings are no longer considered numeric](http://php.net/manual/en/migration70.incompatible.php#migration70.incompatible.strings.hex)

The initial field flags are entered as hex values in `form.yaml`. Under earlier PHP versions, I'm guessing those values got converted to INTs by the yaml parsing library. Under PHP 7, it appears they come in as strings. When MySQL tries to write a string of a hex number (ex: "0x777A3") to an INT column, it saves it as 0.

rayfoss:
I can reproduce this... it also caused problems loging in with internet exploder 11. We should really catch this is testing.

See pull request https://github.com/osTicket/osTicket/pull/3621

-- Sent from my iPhone

PR #3621, opened 2017-01-17
Fixes issue #3620.

dellsala:
Change to the Spyc library to correctly convert hex strings to INTs under PHP 7. Should fix issue #3620

### DIFF
--- a/include/Spyc.php
+++ b/include/Spyc.php
@@ -615,10 +615,12 @@ class Spyc {
       return false;
     }
 
+    if (is_string($value) && preg_match('/^0x[0-9a-fA-F]+$/', $value)) {
+      return hexdec($value);
+    }
+
     if (is_numeric($value)) {
       if ($value === '0') return 0;
-      if (stripos($value, '0x') === 0)
-        $value = hexdec($value);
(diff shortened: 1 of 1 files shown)


rayfoss:
there is a new version of Spyc with a regex for hex that supports capital X... this is an upstream problem.
We should probably just update spyc once it's fixed. https://github.com/mustangostang/spyc/pull/60

but this fix works, it's kind of urgent. 

http://php.net/manual/en/function.is-numeric.php
https://github.com/mustangostang/spyc

rayfoss:
the upstream pulled this change, but we had also modified Spyc to support multi line scalars... pushed a request to merge the change. Waiting on them.

rayfoss:
Still waiting on Spicy to review my request to merge @greezybacon 's multiline scalar support. The documentation for `is_numeric` is also confusing as their example shows output from 5.6, which is wrong on 7.0.0, I've submitted an update to their doc.

As of `7.0.0	Strings in hexadecimal (e.g. 0xf4c3b00c) notation are no longer regarded as numeric strings, i.e. is_numeric() returns FALSE now.`

PR #3661, opened 2017-02-07
CDN Features for files uploaded to system.

iHunt101:
Hi,
Can you included a php functions that will uload the files in the system and trow it to another server?
Thanks!


ntozier:
Wouldn't that behavior fall under an Attachment::Remote plugin?
(There are a couple plugins that do things like filesystem, AWS, etc.)  
you would add this at the plugin repo.

PR #3662, opened 2017-02-07
CDN Features for files uploaded to system

iHunt101:
Can you add this functionality o a script of php, just like the CDN rules that transfer or store all files on another server? Currently base on the plugin available, it oly store data in a folder. This will be a big help for a small storage server.


ntozier:
@iHunt101 Wouldn't that behavior fall under an Attachment::Remote plugin?
(There are a couple plugins that do things like filesystem, AWS, etc.)
you would add this at the plugin repo.

iHunt101:
Yes i got you. But i have no AWS and were trying to build our own CDN server to connect to out support system. This system is intranet base and we only use this for the company.

ntozier:
Which doesn't change that this would still belong in the plugin repo.

Please close this and open a request there.
https://github.com/osTicket/osTicket-plugins

Issue #3654, opened 2017-02-04
ticket time is showing wrong


lijothelegend:
Ticket time is showing wrong . I have checked in my server time and my sql server time is correct there.

![os info](https://cloud.githubusercontent.com/assets/22472451/22619818/eae33d12-eb22-11e6-811b-a77fccd75172.JPG)
![os ticket my sql time](https://cloud.githubusercontent.com/assets/22472451/22619819/eae41f7a-eb22-11e6-8eef-7a941972a68d.JPG)
![os ticket](https://cloud.githubusercontent.com/assets/22472451/22619821/eae54a9e-eb22-11e6-9939-7f9337524f64.JPG)
![os timezone](https://cloud.githubusercontent.com/assets/22472451/22619820/eae4e770-eb22-11e6-9dfa-c5bb404a9b56.JPG)
![osdate](https://cloud.githubusercontent.com/assets/22472451/22619822/eae7f5a0-eb22-11e6-8c92-bbf8b2bba111.JPG)

Ticket time is showing wrong . I have checked in my server time and my sql server time is correct there.

iHunt101:
Try to check your php.ini conf and set your time zone..

lijothelegend:
time zone in my php.ini is correct 



Thanks&Regards
Lijo Joseph George
Technical Team|| ManoramaOnline
Mobile: +91 9895336448 | Office : +91 0481 2563646 | Ext :670
Visit us:www.manoramaonline.com | LinkedIn | Facebook | Twitter

fdellwing:
If the time is wrong as you state, what should the time be? By how much is it off?

lijothelegend:
Seem that its showing time of  Timezone IST (Interpreted as 
Asia/Jerusalem) not of Asia/Kolkata  I need the time of Asia/Kolkata for 
the tickets.

Thanks&Regards
Lijo Joseph George
Technical Team|| ManoramaOnline
Mobile: +91 9895336448 | Office : +91 0481 2563646 | Ext :670
Visit us:www.manoramaonline.com | LinkedIn | Facebook | Twitter

fdellwing:
Well, this is already know: #3583

fdellwing:
Looked into it and are completely confused...

[...]

PR #3663, opened 2017-02-07
trying to fix timezone problem #3654

fdellwing:
This will solve the problem in #3654 but I am not sure what the `$fromDb` is for. Is this needed? Why was it where in the first place?

What this change will do, is converting the timestamp we get from DB (queried in tickets.inc.php) in an UTC DateTime object instead of converting it to an unix timestamp and giving that to the IntlDateFormatter.

Please test this with other timezones!

### DIFF
--- a/include/class.format.php
+++ b/include/class.format.php
@@ -516,11 +516,12 @@ class Format {
             return '';
 
         if ($fromDb)
-            $timestamp = Misc::db2gmtime($timestamp);
+            $timestamp = new DateTime($timestamp);
 
         if (class_exists('IntlDateFormatter')) {
             $locale = Internationalization::getCurrentLocale($user);
             $key = "{$locale}:{$dayType}:{$timeType}:{$timezone}:{$format}";
+
             if (!isset($cache[$key])) {
                 // Setting up the IntlDateFormatter is pretty expensive, so
                 // cache it since there aren't many variations of the


fdellwing:
Second commit because I dont need an UTC DateTime object.

But I am quite sure, it is not correct now either. How to get the timezone of the database? Because the DateTime object needs to be in that timezone to work. This will not be visible as long as the user and the database have the same timezone, but if the differ, the time will be wrong.

fdellwing:
Should work now. Please test.

fdellwing:
Not happy with this. We should use `date_default_timezone_get()` to determinate the timezone of the database. Because db_timezone() will give only the abbr. But in osTicket this will always give `UTC`...

fdellwing:
Ok, I am happy with this one now. But we should make some things clear:

Setting date.timezone should be forced not just recommended.
There are some files where db_timezone() is used. The function should be altered to use the ini_get() function instead of reading it from `@@system_time_zone`.
We need to check if the abbr of the ini_get() function ist the same as `@@system_time_zone` (NOT the other way around!).

include/mysqli.php
```diff
@@ -115,7 +115,14 @@ function db_version() {
...
 function db_timezone() {
-    return db_get_variable('system_time_zone', 'global');
+	$php_tz = ini_get('date.timezone');
+	$php_abbr = new DateTime('now', new DateTimeZone($php_tz));
+	$php_abbr = $php_abbr->format('T');
+	$sql_tz = db_get_variable('system_time_zone', 'global');
+	if ($php_abbr === $sql_tz)
+		return $php_tz;
+
+	return false;
```
fdellwing: Here my first version of the new `db_timezone()` function. It works fine if the `$php_tz` is returned, but I am not quite sure what to return if not. Best would be showing a warning and using `$php_tz` non the less

fdellwing: Here my first version of the new `db_timezone()` function. It works fine if the `$php_tz` is returned, but I am not quite sure what to return if not. Best would be showing a warning and using `$php_tz` non the less

include/mysqli.php
```diff
@@ -115,7 +115,15 @@ function db_version() {
...
 function db_timezone() {
-    return db_get_variable('system_time_zone', 'global');
+	$php_tz = ini_get('date.timezone');
+	$php_abbr = new DateTime('now', new DateTimeZone($php_tz));
+	$php_abbr = $php_abbr->format('T');
+	$sql_tz = db_get_variable('system_time_zone', 'global');
+	if ($php_abbr === $sql_tz)
+		return $php_tz;
+
+	echo '<i class="icon icon-warning-sign" title="Mismatch between \'date.timezone\' and \'mysql timezone\'"></i> ';
```
fdellwing: This seems to work now as intented. But it seems quite strange ` Europe/London (Interpretiert als Europe/Berlin) `...

It checks there for a config key named `db_timezone`, but I cant find that config in database and dont know how and when this will change.

rayfoss: mismatches are normal... you don't always have authority of the mysql server timezone. That echo might break some things... best to log in dashboard.

fdellwing: As I know, the function gets only called by system.php. But if you explain me, how to log that somethere else, I will do that.

include/class.format.php
```diff
@@ -514,13 +514,21 @@ function __formatDate($timestamp, $format, $fromDb, $dayType, $timeType,
 
         if (!$timestamp)
             return '';
+	
+	// We need this codepiece to make sure we have our timestamp in 'Y-m-d H:i:s' format, as sometimes it is given to this function as unixtime
+	//TODO fix that!
+	if (is_numeric($timestamp)) {
+		$tmp = new DateTime();
+		$timestamp = $tmp->setTimestamp($timestamp)->format('Y-m-d H:i:s');
```
scslogin: fdellwing  - This is very helpful. In my case, system_time_zone in MySQL picks up "AST" from the host, and with your fix, the time is correct on our tickets, but bottom of the page in /scp/system.php now says "Timezone: America/Santo_Domingo (Interpreted as Asia/Riyadh)".

fdellwing: Yes, I know that. But I have not yet found proper solution for that.

scslogin: It seems like this needs to be added to the next release, even if it's not perfect yet. It's a whole lot closer than the completely broken timezone assignment it was doing before. How does that happen (that this fix gets merged into the next release)?

GodOfWarWeed:
this broke my time entirely. Please check if you are able to see the system timezone on the ticket queue or new ticket form

Issue #3674, opened 2017-02-09
Error in pagination: Users->Organizations


adi76:
### Prerequisites

* [yes ] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [no ] Do you have any errors in the PHP error log, or javascript console?
* [yes] Did you check the [osTicket forums](http://osticket.com/forum/categories)?
* [no] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/osTicket/osTicket/wiki/Github-Issues-Guidelines)

### Description

When you go to User->Organizations (for example: url: ..../scp/orgs.php?id=1) and try to go to next page (for example: 2) the destination url should looks like: ..../scp/orgs.php?id=1&p=2 but I got .../scp/users.php?sort=&order=&p=2

The problem is in file: include/staff/templates/users.tmpl.php line 39:

Is:
$pageNav->setURL('users.php', $qs);

Should be:
$pageNav->setURL('orgs.php?id=',$org->getId(), $qs);

### Steps to Reproduce

1. Login to SCP
2. Go to Users
3. Go to Organizations of (for example: id=1)
4. Try to go to page number (for example: 2)

**Expected behavior:** [What you expected to happen]
..../scp/orgs.php?id=1&p=2

**Actual behavior:** [What actually happened]
.../scp/users.php?sort=&order=&p=2

### Versions
1.9.14
1.10
Admin panel -> Dashboard -> Information which also additionally gives you information about you server.

Also, please include the OS and what version of the OS you're running. As well as your browser and browser version.

fdellwing:
I dont have enough users to test that, but I am quite sure you can not change the $pageNav to the URL you want.

The userlist in orgs.php is the same  as in users.php. So you would end up with a really messy URL there...

See incluce/staff/org-view.inc.php:
```html
<div class="tab_content" id="users">
<?php
include STAFFINC_DIR . 'templates/users.tmpl.php';
?>
</div>
```

adi76:
I do that change in my $pageNav (as I described earlier), and it works fine for me. I wrote this issue, to let you know that the error exists in osTicket 1.9.14 and 1.10. Maybe you should make this change to osTicket, to have this working just fine for other users in the future.

ps. sorry for my english :-)

fdellwing:
[...]

PR #3677, opened 2017-02-10
fixxes navbar while pagination #3674

fdellwing:
Fixxes Issue #3674

### DIFF
--- a/include/staff/templates/users.tmpl.php
+++ b/include/staff/templates/users.tmpl.php
@@ -34,9 +34,9 @@ $total=db_count('SELECT count(DISTINCT user.id) '.$from.' '.$where);
 $page=($_GET['p'] && is_numeric($_GET['p']))?$_GET['p']:1;
 $pageNav=new Pagenate($total,$page,PAGE_LIMIT);
 $qstr = '&amp;'. Http::build_query($qs);
-$qs += array('sort' => $_REQUEST['sort'], 'order' => $_REQUEST['order']);
+$qs += array('id' => $org->getId(), 'sort' => $_REQUEST['sort'], 'order' => $_REQUEST['order']);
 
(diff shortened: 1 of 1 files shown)


include/staff/templates/users.tmpl.php
```diff
@@ -34,9 +34,9 @@
 $page=($_GET['p'] && is_numeric($_GET['p']))?$_GET['p']:1;
 $pageNav=new Pagenate($total,$page,PAGE_LIMIT);
 $qstr = '&amp;'. Http::build_query($qs);
-$qs += array('sort' => $_REQUEST['sort'], 'order' => $_REQUEST['order']);
+$qs += array('id' => $org->getId(), 'sort' => $_REQUEST['sort'], 'order' => $_REQUEST['order']);
 
-$pageNav->setURL('users.php', $qs);
+$pageNav->setURL(basename($_SERVER['PHP_SELF']), $qs);
```
fdellwing: You could hardcode that, because this template is only used in `incluce/staff/org-view.inc.php`. But this looks cleaner.

balojs:
@balojs

include/staff/templates/users.tmpl.php
adi76: I made changes you wrote, but I think it doesn't work fine. The missing part in new URL is "id=<number>", where number is id of organization. 

With your changes the url looks like: ...../scp/orgs.php?sort=&order=&p=2

but it should:

....../scp/orgs.php?id=5&p=2

fdellwing: Where is a changed line above that that you see here ;)

Please look at the whole change in "Files Changed" section before commenting.

adi76:
Sorry, my mistake

PR #3734, opened 2017-03-09
Fix lookup of database time zone

### LINKED ISSUE #3583: Timezone IST (and others) Ambiguity
The timezone returned by `timezone_name_from_abbr` can not be trusted as it just selects the first result sorting the tz name alphabetically. This messes up times displayed in the site even though timezone in OSTicket System Settings, OSTicket User Prefs, php.ini, MySQL, Server OS are all set to the same value.

Eg. IST refers to Israel/Indian/Irish Standard Times (Ref: [List_of_time_zone_abbreviations](https://en.wikipedia.org/wiki/List_of_time_zone_abbreviations)) 
One may intend to use Indian/Irish timezone but "Asia/Jerusalem" comes first alphabetically.
Few other TZs which I found to have duplicates are ACT, AMT, BST, etc.,

Either function `get_from_database` in `class.timezone.php:165` should be modified to translate ambiguous timezones with some other reference value or a better method must be implemented.

Related StackOverflow answer: http://stackoverflow.com/a/30958892

antriver:
Fixes https://github.com/osTicket/osTicket/issues/3583 and https://github.com/osTicket/osTicket/issues/3654

**The problem:**

Times on ticket entries, and calculations of response times are broken.

In our case we are using Asia/Shanghai, so CST (China Standard Time) is the timezone on the MySQL sever. However `timezone_name_from_abbr` returns America/Chicago as it interprets CST as Central Standard Time. This can be seen on the Dashboard > Information page as:

> Timezone	CST (Interpreted as America/Chicago)

**The fix:**

`timezone_name_from_abbr` supports a second parameter which is a timezone offset in seconds, to aid selecting the correct timezone. This PR selects the current timezone offset in seconds to pass to  `timezone_name_from_abbr`. This results in the correct Asia/Shanghai timezone being returned.

More info: http://php.net/manual/en/function.timezone-name-from-abbr.php

The System information page now shows:

> Timezone	CST (Interpreted as Asia/Chongqing)

Asia/Chongqing is equivalent to Asia/Shanghai, so it's all good.

**Notes:**

This will be okay with DST, as by default (omitting the third parameter to `timezone_name_from_abbr)` it will take into account wether DST is currently in effect or not. So passing an offset of -6 hours for MST will work during daylight savings, and -7 will work at other times.

### DIFF
--- a/include/class.timezone.php
+++ b/include/class.timezone.php
@@ -165,6 +165,7 @@ class DbTimezone {
     function get_from_database() {
         // Attempt to fetch timezone direct from the database
         $TZ = db_timezone();
+        $utc_offset_seconds = db_timezone_offset();
 
         // Translate ambiguous 'GMT' timezone
         if ($TZ === 'GMT') {
@@ -176,7 +177,7 @@ class DbTimezone {
         // Forbid timezone abbreviations like 'CDT'
         elseif ($TZ !== 'UTC' && strpos($TZ, '/') === false) {
             // Attempt to lookup based on the abbreviation
-            if (!($TZ = timezone_name_from_abbr($TZ)))
+            if (!($TZ = timezone_name_from_abbr($TZ, $utc_offset_seconds)))
                 // Abbreviation doesn't point to anything valid
                 return false;
         }
--- a/include/mysqli.php
+++ b/include/mysqli.php
@@ -118,6 +118,20 @@ function db_timezone() {
     return db_get_variable('system_time_zone', 'global');
 }
 
+/**
+ * Returns the database server's offset from UTC in seconds.
+ *
+ * @return int|null
+ */
+function db_timezone_offset() {
+    $res = db_query('SELECT TIMESTAMPDIFF(SECOND, UTC_TIMESTAMP, NOW());');
+    if ($row = db_fetch_row($res)) {
+        return (int)$row[0];
+    }
+
+    return null;
+}
+
 function db_get_variable($variable, $type='session') {
     $sql =sprintf('SELECT @@%s.%s', $type, $variable);
     return db_result(db_query($sql));


scslogin:
I have AST, which in my case is America/Santo_Domingo. However, the system incorrectly interprets AST as Asia/Riyadh. Tix unfortunately did not resolve my problem.  https://github.com/osTicket/osTicket/pull/3663 fixed my time issue, but it still reads incorrectly in the system settings (see my comment at the bottom of that page).

roman-1983 (approved):

scslogin:
Thank you, roman-1983

antriver:
@scslogin Could you please show the output of this query on your database server:

```sql
SELECT @@system_time_zone, @@session.time_zone, NOW(), CURRENT_TIMESTAMP, UTC_TIMESTAMP, TIMESTAMPDIFF(SECOND, UTC_TIMESTAMP, NOW());
```

scslogin:
AST	SYSTEM	2017-03-16 21:25:02	2017-03-16 21:25:02	2017-03-17 01:25:02	-14400

scslogin:
I did the changes in both files again, just to make sure I hadn't made a mistake, and still saw no change from the shipipng version. I had to go back to https://github.com/osTicket/osTicket/pull/3663.

antriver:
@scslogin That sounds like you were manually applying these changes to a different version of the code, right?

Could you test with the actual branch in this PR and report back if that works or not.

Thanks!

GodOfWarWeed:
This worked for me . 
It was IST interpreted as Asia/Jerusulum
It should be Asia/Kolkata
Now its Asia/Calcutta
Basically sameplace different spellings.

I gotta ask, was using db_timezone absolutely necessary since 1.10?

scslogin:
@antriver - I'd like to, but, honestly, I'm not sure how to do that and am worried that I'll break my deployment by switching from something other than an unzipped version of the current stable release. What do most people do?

atanu2016:
> This worked for me .
> It was IST interpreted as Asia/Jerusulum
> It should be Asia/Kolkata
> Now its Asia/Calcutta
> Basically sameplace different spellings.
> 
> I gotta ask, was using db_timezone absolutely necessary since 1.10?

Please let me know how you fixed the issue !! i am also facing the same issue & need your help. Where & what changes have to made to get that working

Issue #3679, opened 2017-02-10
Dashboard Response Time


fdellwing:
So, we have this query for the dashboard:
```sql
SELECT A2.id AS dept__id, AVG(DATEDIFF(A1.closed, A1.created)) AS ServiceTime, AVG(DATEDIFF(A4.created, A5.created)) AS ResponseTime FROM ost_ticket A1 JOIN ost_department A2 ON (A1.dept_id = A2.id) LEFT JOIN ost_thread A3 ON (A3.object_type = 'T' AND A1.ticket_id = A3.object_id) LEFT JOIN ost_thread_entry A4 ON (A3.id = A4.thread_id AND A4.type = 'R') LEFT JOIN ost_thread_entry A5 ON (A4.pid = A5.id) WHERE A1.dept_id IN ('1', '3', '4', '5', '6') GROUP BY A2.id;
```
The ResponseTime gets calulated by this `AVG(DATEDIFF(A4.created, A5.created)) AS ResponseTime` where A4 is `ost_thread_entry A4 ON (A3.id = A4.thread_id AND A4.type = 'R')` and `A5 is ost_thread_entry A5 ON (A4.pid = A5.id)`

But as you can see here, only responses posted by filters even have a `pid`, agent ones not:
```
id, pid, thread_id, staff_id, user_id, type, flags, poster, editor, editor_type, source, title, body, format, ip_address, created, updated
24, 0, 7, 1, 0, R, 64, Fabian Dellwing, , , , , aa, html, 172.16.3.104, 2017-02-06 13:28:02, 0000-00-00 00:00:00
25, 0, 7, 1, 0, R, 64, Fabian Dellwing, , , , , test, html, 172.16.3.104, 2017-02-06 13:36:28, 0000-00-00 00:00:00
26, 0, 7, 1, 0, R, 64, Fabian Dellwing, , , , , test, html, 172.16.3.104, 2017-02-07 10:36:41, 0000-00-00 00:00:00
27, 0, 7, 1, 0, R, 64, Fabian Dellwing, , , , , test, html, 172.16.3.104, 2017-02-07 10:46:17, 0000-00-00 00:00:00
28, 0, 7, 1, 0, R, 64, Fabian Dellwing, , , , , test, html, 172.16.3.104, 2017-02-07 11:00:49, 0000-00-00 00:00:00
[...]

PR #3738, opened 2017-03-09
Fix service & reponse times on dashboard

antriver:
Currently the agent dashboard statistics are showing the same service and response time for every department.

**Before:**
screen shot 2017-03-09 at 13 54 05

This is happening because in `class.report.php` it is trying to look up the times for the department by the `dept_id` key. But in the query the id is actually `dept__id` (2 underscores). I changed dept__id back to dept_id.

I also cleaned up the code a bit as it was initially hard to understand due to how compact it was.

**After:**
screen shot 2017-03-09 at 14 09 45

### DIFF
--- a/include/class.report.php
+++ b/include/class.report.php
@@ -186,48 +186,48 @@ class OverviewReport {
             ));
 
         switch ($group) {
-        case 'dept':
-            $headers = array(__('Department'));
-            $header = function($row) { return Dept::getLocalNameById($row['dept_id'], $row['dept__name']); };
-            $pk = 'dept_id';
-            $stats = $stats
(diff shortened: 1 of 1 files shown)


balojs:
@antriver thanks a lot

fdellwing:
Take a look at: #3679 for the Response Time and #3678 for the Service Time

compuvin:
Appreciate this. Department statistics are now in line with agent statistics.

gtulga:
I implemented fix patch to 1.10.1 and it works fine. But if I change timeframe, service time and response time are not updated. They are always showing one value.

PR #3761, opened 2017-03-23
Feature: Show urls on custom field texts.

TheTonon:
Using regex, show urls inside custom field contents. Pretty useful for when you have a lot of locations and must use maps to redirect your agents.


aydreeihn:
Thank you for the pull request. We actually have a function you can call (Format::clickableurls) to achieve this.

PR #3763, opened 2017-03-24
Special Answer on Closing Ticket

encendedor:
Corresponding to
https://github.com/osTicket/osTicket-1.7/pull/392
I wanted to ask if anybody is still working on a way to define a special e-mail answer for closing a ticket? Ther was a workaround till 1.9 adding a few lines into class.ticket.php but that doesn't work anymore. 

The thing is that often customers answer with a "thank you" to a closed ticket which reopens it. I want to add the information to the e-mail, that answering reopens it.

### DIFF
--- a/css/redactor.css
+++ b/css/redactor.css
@@ -327,7 +327,7 @@ body .redactor_box_fullscreen {
   background: #fff;
   border: none;
   box-shadow: 0 1px 2px rgba(0, 0, 0, 0.2);
-  z-index: 3;
+  z-index: 1;
 }
 .redactor_toolbar:after {
   content: "";


ntozier:
The devs have never indicated to me that this was something that they were planning on implementing. 

I also do not feel that it is required or needed.  This is a Work flow: update ticket, select closed.  You just send a Closed ticket notification.

encendedor:
I think it would be a nice option for many users, because for example you can also add a link to a poll tool or something like this to ask the customer how satisfied he was with the solution of his problem. Your workflow is possible, but therefore I always have to add the same line ("please do not reply...") in every final message manually.

ntozier:
No actually you do not.  You can already do that with Canned Responses...  using what I said.  We've been doing it for years this way.

PR #3783, opened 2017-04-07
Address edge cases where timezone mixups happens on DateTimeField

protich:
Allow datetime field to be timezone agnostic (not timezone aware) to display the timezone used to set the field. The timezone of the last user or agent that edited the field is used.

### DIFF
--- a/include/class.format.php
+++ b/include/class.format.php
@@ -512,74 +512,128 @@ class Format {
         global $cfg;
         static $cache;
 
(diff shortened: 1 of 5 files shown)


blueyeguy:
Thank you @protich, this will make a lot of people in the forums happy.

protich:
With addition of TimezoneField, it's now possible to set a desired timezone on a DateTimeField. See PR #3786 for details.

PR #3794, opened 2017-04-14
subject max lenght changed from 30 to 70

davidegiunchidiennea:
The maximum lenght of the new email's subject was not specified, so it was default to 30 characters, that's too little. Changed to 70, as of recommended by rfc http://www.faqs.org/rfcs/rfc2822.html

### DIFF
--- a/include/class.filter_action.php
+++ b/include/class.filter_action.php
@@ -545,6 +545,7 @@ class FA_SendEmail extends TriggerAction {
                 'required' => true,
                 'configuration' => array(
                     'size' => 80,
+                    'length' => 70,
                     'placeholder' => __('Subject')
                 ),
             )),


ntozier:
I'd like to see this changed to be admin controllable (perhaps either by parameter in the ost-config file, or some place in the UI.)

davidegiunchidiennea:
@ntozier i don't think that this setting will be useful: if my patch would limit to 70 character the subject of every email sent by osticket, an admin-controllable-setting would be useful.
But my patch raise the limit (that actually is 30 character) only of the new email on the new email filter!
you can find it on "/scp/, then "managed administrative panel" -> manage -> ticket filters -> new filter -> filter operation -> select "send an email", then you will see the subject, it's only THAT subject.

PR #3795, opened 2017-04-14
Use <a> tag for links

jdelhome3578:
Changed the `<div>` tag to an `<a>` tag to properly display links in some browsers.


clonemeagain (commented):
Does this fix the table view of tickets, allowing middle-click for new-tab opening of multiple tickets simultaneously? 
Because I'm on board with that!

Issue #3800, opened 2017-04-19
Client ticket thread broken


Xm4Tr1X:
osTicket Version 	v1.10 (901e5ea) — Up to date
Web Server Software 	Apache/2.4.12 (Win32) OpenSSL/1.0.1l PHP/5.6.8
MySQL Version 	5.6.24
PHP Version 	5.6.8

### Description
On the client portal, if richtext is enabled and client types a `<div>` tag and posts the message without closing it, then the alignment of the following thread entries are broken until another thread entry follows up with </div>. This does not affect thread events(such as assignment and transfers)
However the staff panel does not have this issue. 

![ticket thread](https://cloud.githubusercontent.com/assets/19608968/25167912/1b1518f0-24ff-11e7-926d-ab57169ae812.png)

### Steps to Reproduce

1. Login to client portal
2. open a ticket and in the issue description write `<div> `and post the message.
3. post more messages after the <div> message. 

PS: richtext must be enabled through admin panel -> system -> enable richtext. 
PPS: the message must contain only `<div>` open tag and nothing else for this to take effect, Else it gets stripped if theres other content

GodOfWarWeed:
Ok so i found a temp fix. thanks to http://stackoverflow.com/a/3810341/4908928

In ticket.php  `function closetags($html) {
    if(!extension_loaded(tidy)){
    preg_match_all('#<(?!meta|img|br|hr|input\b)\b([a-z]+)(?: .*)?(?<![/|/ ])>#iU', $html, $result);
    $openedtags = $result[1];
    preg_match_all('#</([a-z]+)>#iU', $html, $result);
    $closedtags = $result[1];
    $len_opened = count($openedtags);
    if (count($closedtags) == $len_opened) {
        return $html;
    }
    $openedtags = array_reverse($openedtags);
    for ($i=0; $i < $len_opened; $i++) {
        if (!in_array($openedtags[$i], $closedtags)) {
            $html .= '</'.$openedtags[$i].'>';
        } else {
            unset($closedtags[array_search($openedtags[$i], $closedtags)]);
        }
    }
    return $html;
    }
 else {
        $tidy = new tidy();
        $clean = $tidy->repairString($html,[
            'output-xml' => true,
            'input-xml' => true
            ]);
        return $clean;
    }
} `

then at line 112  `'message' => closetags($_POST['message'])`

PR #3801, opened 2017-04-19
Update tickets.php

GodOfWarWeed:
solves issue #3800


tickets.php
```diff
@@ -19,6 +19,35 @@
 
 if ($thisclient->isGuest())
     $_REQUEST['id'] = $thisclient->getTicketId();
+function closetags($html) {
```
GodOfWarWeed: basically a function to close open tags in messages. such as divs etc. 
use tidy if its available, else use the regex

PR #3832, opened 2017-05-17
Add GTID compatibility

kler:
# Adding GTID compatibility to osTicket.

Because GTID-based replication is dependent on transactions, some features otherwise available in MySQL are not supported when using it.

`CREATE TABLE ... SELECT` is not safe for statement-based replication. When using row-based replication, this statement is actually logged as two separate events — one for the creation of the table, and another for the insertion of rows from the source table into the new table just created. When this statement is executed within a transaction, it is possible in some cases for these two events to receive the same transaction identifier, which means that the transaction containing the inserts is skipped by the slave. Therefore, `CREATE TABLE ... SELECT` is not supported when using GTID-based replication.

GTID compatibility is unconditionally enforced by for instance Google Cloud SQL Second Generation instances.

The error produced when running osTicket in a MySQL instance with GTID-based replication was rather obscure, since the original error was in `buildDynamicDataForms()` in `include/class.dynamic_forms.php`. This error was however hidden, and the next error that did show up in the log was when `include/class.orm.php` `MySqlExecutor->execute()` threw an error after trying to read from a table which didn't exist (`ost_ticket__cdata`).

The origin of the error is that `buildDynamicDataForms()` uses a `CREATE TABLE ... SELECT` style SQL call, and this is not supported when using GTID-based replication.
The error was resolved by splitting the query into several statements, using temporary tables.

# How to reproduce the error

1. Setup osTicket and start MySQL using the parameters: `--gtid-mode=ON --enforce-gtid-consistency=ON --log-bin --server-id=0` to enable GTID-based replication.
2. Log in as the agent created in the setup
3. Default ticket exists *(osTicket installed!)*, it's counted as **Open** but not listed

gtid_incompatible

# References

- http://stackoverflow.com/questions/5859391/create-a-temporary-table-in-a-select-statement-without-a-separate-create-table
- https://www.drupal.org/node/2209853
- https://dev.mysql.com/doc/refman/5.7/en/replication-gtids-restrictions.html 

# Stacktrace of the rather obscure error

    PHP Fatal error:  Uncaught exception 'InconsistentModelException' with message 'Unable to prepare query:  SELECT A5.`staff_id` AS `lock__staff_id`, A1.`staff_id` AS `staff_id`, A1.`isoverdue` AS `isoverdue`, A1.`team_id` AS `team_id`, A1.`ticket_id` AS `ticket_id`, A1.`number` AS `number`, A3.`subject` AS `cdata__subject`, A7.`address` AS `user__default_email__address`, A1.`source` AS `source`, A4.`priority_color` AS `cdata__:priority__priority_color`, A4.`priority_desc` AS `cdata__:priority__priority_desc`, A1.`status_id` AS `status_id`, A8.`name` AS `status__name`, A8.`state` AS `status__state`, A1.`dept_id` AS `dept_id`, B0.`name` AS `dept__name`, A6.`name` AS `user__name`, A1.`lastupdate` AS `lastupdate`, A1.`isanswered` AS `isanswered`, B1.`firstname` AS `staff__firstname`, B1.`lastname` AS `staff__lastname`, B2.`name` AS `team__name`, (SELECT COUNT(R0.`id`) AS `count` FROM `ost_thread` Q7 JOIN `ost_ticket` Q8 ON (Q7.`object_type` = 'T' AND Q7.`object_id` = Q8.`ticket_id`) LEFT JOIN `ost_thread_collaborator` R0 ON (Q7.`id in /var/www/localhost/htdocs/	 on line 3134, referer: http://127.0.0.1/scp/login.php

### DIFF
--- a/include/class.dynamic_forms.php
+++ b/include/class.dynamic_forms.php
@@ -286,10 +286,31 @@ class DynamicForm extends VerySimpleModel {
     }
 
     static function buildDynamicDataView($cdata) {
-        $sql = 'CREATE TABLE IF NOT EXISTS `'.$cdata['table'].'` (PRIMARY KEY
-                ('.$cdata['object_id'].')) DEFAULT CHARSET=utf8 AS '
-             .  static::getCrossTabQuery( $cdata['object_type'], $cdata['object_id']);
-        db_query($sql);
+        $table = $cdata['table'];
+        // To make temp table compatible with mysqli using persistent connections
+        $rand = substr(md5(microtime()), rand(0,26),5);
+        $tmptable = $table.'_'.$rand;
+
+        $sql = 'SHOW TABLES LIKE "'.$table.'"';
+        $ret = db_query($sql);
+        $exists = (mysqli_num_rows($ret) > 0);
+        if ($exists) {  // Nothing to do
+            return;
+        }
+
+        // This rather complicated structure using temporary table is to make the code GTID compatible.
+        // GTID-based replication does not allow CREATE TABLE ... SELECT
+        // See https://dev.mysql.com/doc/refman/5.7/en/replication-gtids-restrictions.html
+        $ret = true;
+        $ret = $ret && db_query('CREATE TEMPORARY TABLE `'.$tmptable.'`'
+            .'(PRIMARY KEY ('.$cdata['object_id'].')) DEFAULT CHARSET=utf8 AS '
+            .static::getCrossTabQuery($cdata['object_type'], $cdata['object_id']));  // create temp table with data and structure
+        $ret = $ret && db_query('CREATE TABLE IF NOT EXISTS `'.$table.'` LIKE `'.$tmptable.'`');  // create structure
+        $ret = $ret && db_query('INSERT `'.$table.'` SELECT * FROM `'.$tmptable.'`');  // copy data
+        $ret = $ret && db_query('DROP TEMPORARY TABLE IF EXISTS `'.$tmptable.'`');  // remove temp table
+        if (!$ret) {
+            throw new Exception(__('buildDynamicDataView: failed to create temporary table'));
+        }
     }
 
     static function dropDynamicDataView($table) {


rlKoekie:
Thank you for this patch! I've patched my 1.9 version to use it, which required some changes.
The following buildDynamicView() function seems to work for me, and it might also work on the 1.10 branch:
```
    static function buildDynamicDataView() {
        $table = '\''.TABLE_PREFIX.'ticket__cdata\'';
        // To make temp table compatible with mysqli using persistent connections
        $rand = substr(md5(microtime()), rand(0,26),5);
        $tmptable = $table.'_'.$rand;
        $sql = 'SHOW TABLES LIKE "'.$table.'"';
        $ret = db_query($sql);
        $exists = (mysqli_num_rows($ret) > 0);
        if ($exists) {  // Nothing to do
            return;
        }
        // This rather complicated structure using temporary table is to make the code GTID compatible.
        // GTID-based replication does not allow CREATE TABLE ... SELECT
        // See https://dev.mysql.com/doc/refman/5.7/en/replication-gtids-restrictions.html
        $ret = true;
        $ret = $ret && db_query('CREATE TEMPORARY TABLE `'.$tmptable.'`'
            .'(PRIMARY KEY (ticket_id)) DEFAULT CHARSET=utf8 AS '
            .static::getCrossTabQuery('T', 'ticket_id'));  // create temp table with data and structure
        $ret = $ret && db_query('CREATE TABLE IF NOT EXISTS `'.$table.'` LIKE `'.$tmptable.'`');  // create structure
        $ret = $ret && db_query('INSERT `'.$table.'` SELECT * FROM `'.$tmptable.'`');  // copy data
        $ret = $ret && db_query('DROP TEMPORARY TABLE IF EXISTS `'.$tmptable.'`');  // remove temp table
        if (!$ret) {
            throw new Exception(__('buildDynamicDataView: failed to create temporary table'));
        }
    }
```

joeraven0 (commented):
Solves the issue with hidden tickets

mdanka:
@joeraven0 @kler  Any chance this could be merged?

kler:
I can't do that, but it certainly should.

PR #3844, opened 2017-06-07
Feature/dept access export

aydreeihn:
This feature allows an administrator to export a department's members and their roles in the department in csv format.


include/class.export.php
```diff
@@ -258,6 +258,67 @@ function ($f) { return $f->getLocal('label'); }, $fields)));
...
+                        }
+                    }
+                    return $record;
+                    })
+                );
+        exit;
+
+    }
+
+//adriane
```
JediKev: lol

Issue #3842, opened 2017-06-07
Special characters, accents, umlauts in page names generate an error 403


DMTGMBH:
If you are using special characters, accents or umlauts in your pages names osticket will create an error 403.

affected file: include/class.format.php

Remove or escape line 683 to 696:

    // Thanks, http://stackoverflow.com/a/2955878/1025836
    /* static */
    function slugify($text) {
        // replace non letter or digits by -
        $text = preg_replace('~[^\p{L}\p{N}]+~u', '-', $text);

        // trim
        $text = trim($text, '-');

        // lowercase
        $text = strtolower($text);

        return (empty($text)) ? 'n-a' : $text;
    }

and replace it with:

    /* DMT Slugify to support special characters Start */
   
        function slugify($text)
    {
          // replace non letter or digits by -
          $text = preg_replace('~[^\\pL\d]+~u', '-', $text);
        // trim
          $text = trim($text, '-');
        // transliterate
          if (function_exists('iconv'))
              {
            $text = iconv('utf-8', 'us-ascii//TRANSLIT', $text);
              }
        // lowercase
          $text = strtolower($text);
        // remove unwanted characters
       
          $text = preg_replace('~[^-\w]+~', '', $text);
        $text = preg_replace('#&([A-za-z])(?:acute|cedil|caron|circ|grave|orn|ring|slash|th|tilde|uml);#', '\1', $text);
        $text = preg_replace('#&([A-za-z]{2})(?:lig);#', '\1', $text);
        $text = preg_replace('#&[^;]+;#', '', $text);
          if (empty($text))
           {
            return 'n-a';
           }
          return $text;
    }
   
    /* DMT Slugify Ende */

Best regards,
Jürgen

JediKev:
Thanks for the find and suggested fix! Unfortunately, `iconv()` will not convert all normal special characters (e.g. 'ø'). I have made a [pull request](https://github.com/osTicket/osTicket/pull/3874) that will convert all normal special characters.

Thanks again!

DMTGMBH:
With the mod above, characters like  "ø" or "ã" will be converted too.  All special characters, unwanted characters and things like acute, grave etc. will be replaced with preg_replace. Have a look at "//remove unwanted characters", you can adjust this section to your needs. But the mod should replace all unwanted characters for URLs.

JediKev:
[...]

PR #3874, opened 2017-06-28
pages: Translate Special Characters

JediKev:
This addresses issue #3842 where special characters in a Page name causes a
403 or 404 error. This adds a method to convert the special characters to
html entities and then removes the entity suffixes leaving only
un-accented characters behind. (e.g. 'ã' => '\&atilde;' => 'a')

### DIFF
--- a/include/class.format.php
+++ b/include/class.format.php
@@ -737,6 +737,13 @@ class Format {
     // Thanks, http://stackoverflow.com/a/2955878/1025836
     /* static */
     function slugify($text) {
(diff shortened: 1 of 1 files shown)


DMTGMBH:
With the following mod, posted in osticket forum, characters like  "ø" or "ã" will be converted too.

http://osticket.com/forum/discussion/90555/bugfix-class-format-php-error-when-using-special-characters-accents-or-umlauts-in-pages-names#latest

PR #3877, opened 2017-06-29
feature: add ticket closed notification

davidegiunchidiennea:
feature: add ticket closed notification. taken from: https://gist.github.com/clonemeagain/8089055#file-list-of-changes-php-L28 and adapted to the new osticket's version. Please add even the SQL query written in the same url:

INSERT INTO ost_email_template (`id`, `tpl_id`, `code_name`, `subject`, `body`, `created`, `updated`) VALUES (NULL, 1, 'ticket.close', '[#%{ticket.number}] %{ticket.subject}', '%{ticket.name},
Our customer care team has closed the ticket, #%{ticket.number} on your behalf.
If you wish to provide additional comments or information regarding this issue, please don''t open a new ticket. You can update or view this ticket''s progress online here: %{ticket.client_link}.
%{signature}', NOW(), NOW());

### DIFF
--- a/include/class.filter_action.php
+++ b/include/class.filter_action.php
@@ -545,6 +545,7 @@ class FA_SendEmail extends TriggerAction {
                 'required' => true,
                 'configuration' => array(
                     'size' => 80,
+                    'length' => 70,
                     'placeholder' => __('Subject')
                 ),
             )),
--- a/include/class.template.php
+++ b/include/class.template.php
@@ -84,6 +84,14 @@ class EmailTemplateGroup {
                 'ticket', 'signature', 'message', 'poster', 'recipient',
             ),
         ),
+        'ticket.close'=>array(
+            'group'=>'a.ticket.user',
+            'name'=>/* @trans */ 'Ticket Closed Notification',
+            'desc'=>/* @trans */ 'Notice sent to client on the closure of a ticket.',
+            'context' => array(
+                'ticket', 'signature', 'response', 'recipient',
+            ),
+        ),
         'ticket.alert'=>array(
             'group'=>'b.ticket.staff',
             'name'=>/* @trans */ 'New Ticket Alert',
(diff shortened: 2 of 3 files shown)


ntozier:
So to be clear: When an Agent goes into the UI and types a response, and selects Change Status: Closed, will this change generate two emails to the User? (one for the update, and now one for the Close Notification)

Because if it does your chances of this Pull Request being accepted just sank faster than the Titanic.

davidegiunchidiennea:
If an agent select Change Status: Closed and then write a notification to the user (not an internal one): yes, it will generate two emails to the users.
But if an agent doesn't write any notice (it's common), the user will receive one email.

I think that this patch could be more useful if somebody write a "switch" option in the panel to make this function on or off.
In our case this two "close email" is extremely useful because it embed a "satisfaction survey", and i think that this feature will be useful for a lot of people.

ntozier:
This will not get accepted because it generates multiple emails.  If you want it to get accepted then you will need the means to toggle it on/off as part of the PR.

You can achieve this same behavior using a canned response (with the satisfaction survey) and selecting Closed.

JediKev:
You can add the Ticket Status variable in the Reply/Response template so the User sees that the ticket was Closed.

**The template with variable:**
screen shot 2017-06-29 at 10 11 55

**The Email the User receives:**
screen shot 2017-06-29 at 10 16 08

As for the satisfaction survey, you can create a canned response containing the survey. Once you are about to close the ticket add the survey canned response, change the status to closed, and send the response. Now they will get your last response, see the ticket is closed, and get the survey.

davidegiunchidiennea:
I know about the "canned response" feature, the problem is that the agent must select the right canned response EVERY time, if the agent doesn't select the right canned response, the survey will not be send.
I needed the survey to be send every time, independently to the agent's choice (it's a feature present on other ticket's softwares).

ntozier:
I get why you want it.  I don't disagree with your logic.  However if you want this change to have a chance of being accepted you need to also code a way to toggle it on/off.

ManZosh:
@davidegiunchidiennea Hey David, I was wondering what version of osTicket you made these code changes on? I really need some kind of _send email of ticket closing to department manager + end user_ functionality and was wondering if your work is ripe to adapt for this purpose? I've implemented osTicket 1.14.1 with PHP 7.2 for Ubuntu 18.04 into my company's env. Lemme know!

Issue #3849, opened 2017-06-09
Editor formatting issue in Chrome


peterson-dane:
### Prerequisites

* [YES] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [NO] Do you have any errors in the PHP error log, or javascript console?
* [YES] Did you check the [osTicket forums](http://osticket.com/forum/categories)?
* [YES] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

### Description

When attempting to change formatting of text in any input or response field while using the past several current versions of Chrome, the text formatting is not applied. Italics, bold, strikethrough, and underline have absolutely  no effect, while attempting to change the font face or font colors results in `<span [style|color|etc…]=XXXX></span>` being inserted immediately before the highlighted text. I'm including a screenshot of what the HTML looks like, but I'm attempting to format while using the normal text editor mode.

![helpdesk issue](https://user-images.githubusercontent.com/6052087/26987581-1599bb7c-4d1a-11e7-8271-d2ddff239687.PNG)

This issue is present not just on our production server, but is present on a brand new clean server I created to test using the current commit on the `develop` branch of the project.

This issue only presents itself while using Chrome. I've also tested it in Microsoft Edge (v. 40.15063.0.0), and everything seems to work correctly there.

### Steps to Reproduce

1. Open any field where you can type formatted text.
2. Highlight text.
3. Attempt to apply formatting. Formatting is not applied.
4. Check HTML editor to see that tags are either not present or have not been created properly.

**Expected behavior:** Text should be formatted correctly when performing this task within Chrome.

**Actual behavior:** Formatting wasn't applied or was not applied correctly, depending on which type of formatting (see description).

### Versions
**osTicket Version on test server:** 1.8-git (d2ef3b1f) *(current `develop` branch commit)*
**osTicket Version on production server:** v1.10 (901e5ea)

**Remaining software versions are all the same:**
**Web Server Software:** Apache/2.4.18 (Ubuntu)
**MySQL Version**	5.7.18
**PHP Version**: 7.0.18-0ubuntu0.16.04.1
**Server OS:** Ubuntu 16.04.2 LTS
**Client OS:** Windows 10 1703
**Browser:** Chrome 58.0.3029.110

peterson-dane:
Formatting is correctly applied if you select a whole line and begin the selection on the line after the one you're trying to format:

[...]

PR #3885, opened 2017-07-06
redactor: Chrome Formatting Issues

JediKev:
This addresses issue #3849 where Redactor bold, italics, etc. does not work
properly on Chrome versions 58+.


eltreno:
I tried this, but it didn't fix the issue, still won't allow highlighting, then formatting of single or multiple words.

Using Chrome Version 59.0.3071.115 (Official Build) (64-bit) 

Firefox as reported works fine.

Also on Chrome, it will do very random things like:
1. Highlight single word
2. Click Bold --> Nothing ( Inspecting dom shows new blank link inserted under sentence. )
3. Highlight whole sentence - single line only.
4. Click Bold --> (at random few words are bolded but not whole sentence) ( Inspecting dom shows new blank link inserted under sentence with strong tags added areour the few words and another blank new line inserted.)

eltreno:
This fix does work!!!

Apologies for my previous POST, Cloudflare didn't remove old cached file without the version parameter on it - of course!

From my Testing in Chrome, this does in fact fix the wsywyg formatting issue in Chome where you cannot format individual or multiple words in osticket V1.10

telx85:
This fix works for me as well. Thank you very much.

telx85:
Can we please get this added to 1.10.1 (or 1.10.2 ??). Easy enough for me to just replace the file, but for others it can be mighty annoying.

JediKev:
It can't be added to 1.10.1 because it's already packaged and released but it will be added to the next release coming out soon (v1.10.2, I think is what it will be called). Stay tuned!

JediKev:
#4113 is more robust and replaces this. ☹️

Issue #3724, opened 2017-03-06
[Bug]contact information: Editable Settings 


mwells1702:
[ No] Do you have any errors in the PHP error log, or javascript console?
[ Yes] Did you check the osTicket forums?
[ Yes] Did you perform a cursory search to see if your bug or enhancement is already reported?

How to reproduce:

Go to forms -> users informations
Create a field and make it immutable (not editable) for users and agents
Login as user and open the profile
Change the field just as you like. It will be saved.

osTicket Version: v1.10 (901e5ea) —  Up to date
Web Server Software: Microsoft-IIS/8.5
MySQL Version:5.6.33
PHP Version:7.0.16
OS: Windows Server 2012 R2 DC
-Getting no errors. Just does not seem to work.
![upload](https://cloud.githubusercontent.com/assets/26208878/23593198/7af4d6bc-0203-11e7-98d7-293f358796d5.png)

Ok so the field (tick box) that allows you to set "read-only" in the "contact information" section is not working for me. I want users to be able to sign up with name then not be able to change it once logged-in. However when i uptick the box my test user account is still able to edit the full name box?

"Fields marked editable allow agents and endusers to update the content of this field after the form entry has been created."

full information here (http://osticket.com/forum/discussion/89994/field-config-full-name#latest

blueyeguy:
The thread you linked suggested trying PHP 5.6, did you do that? Your information still shows PHP 7

mwells1702:
I have PHP 5.6.30 and have tried that.?

mwells1702:
[06-Mar-2017 19:08:57 UTC] PHP Warning:  array_reduce() expects parameter 1 to be array, null given in C:\Inetpub\vhosts\<domain>\<sub-domain>\include\ajax.forms.php on line 76

Dont know if this helps

dogatemycomputer:
+1 

Marking a field READ ONLY allows the customer to edit it by accessing the customer portal and clicking EDIT.   I dug around in the PHP code and could not find anyplace where the CSS marks the field READ ONLY when generated.  

Server Information
osTicket Version	v1.10 (901e5ea) —  Up to date
Web Server Software	Apache/2.2.15 (CentOS)
MySQL Version	5.5.54
PHP Version	5.6.30

![editable](https://cloud.githubusercontent.com/assets/7565020/24068404/06ee14d0-0b4b-11e7-9113-4e9a0972993e.jpg)

JediKev:
[...]

PR #3888, opened 2017-07-07
forms: User Form Field Editability

JediKev:
This addresses issue #3724 where setting a form field to be non-editable by
a Users does not work. The User will still be able to edit the field when
they edit a ticket. This corrects the check on form field permissions to not
allow a User to edit the field if it’s not enabled.

### DIFF
--- a/include/client/edit.inc.php
+++ b/include/client/edit.inc.php
@@ -16,7 +16,7 @@ if(!defined('OSTCLIENTINC') || !$thisclient || !$ticket || !$ticket->checkUserAc
     <tbody id="dynamic-form">
     <?php if ($forms)
         foreach ($forms as $form) {
-            $form->render(false);
+            $form->render(false, false, $options);
     } ?>
     </tbody>
 </table>
--- a/include/client/templates/dynamic-form.tmpl.php
+++ b/include/client/templates/dynamic-form.tmpl.php
@@ -18,6 +18,10 @@
             if (!$field->isVisibleToUsers() && !$field->isRequiredForUsers())
                 continue;
         }
+        elseif (isset($options['mode']) && $options['mode'] == 'edit') {
+            if (!$field->isEditableToUsers())
+                continue;
+        }
         elseif (!$field->isVisibleToUsers() && !$field->isEditableToUsers()) {
             continue;
         }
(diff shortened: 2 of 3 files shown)


fdellwing:
This does not makes them readonly, it does not display them at all.

JediKev:
@fdellwing not understanding what you mean...can you explain more?

JediKev:
Do you mean that you made a field only Visible and when the user goes to edit the ticket they don't appear? If so then that's by design. **Visible** means at the time of ticket creation they can see and add to the field, **Editable** means _after_ ticket creation they are allowed to see/edit the field, and **Required** means anytime they edit the ticket or create a ticket they have to insert data into the field.

fdellwing:
I am talking about the form "Contact Information" that is used on the users profil page. osTicket does not gives a shit about if that field is "readonly" or not, you can change your name/mail as you wish.

With this PR this gets not fixxed, but the fields for name/email are not there at all.

JediKev:
>I am talking about the form "Contact Information" that is used on the users profil page. osTicket does not gives a shit about if that field is "readonly" or not, you can change your name/mail as you wish.

That is what this pull request fixes...? If you want the User to not update their email address simply go to the form field and disable the Settings option "Editable For EndUsers". This will then hide that field for the Users. 👍 

>With this PR this gets not fixxed, but the fields for name/email are not there at all.

That's because if the field is readonly then the User will not be able to see/edit the field in edit mode. When you click Profile as a User it sends you to the User Contact Information form in edit mode.

I'll explain the field Settings options for you again but this time a little more clear:
**Enabled** - Means that the field can be seen and used.
**Visible** - Means that the User can see/edit the field upon initial ticket creation or account creation.
**Required** - Means that the User must insert data before clicking save. (initially and in edit mode)
**Editable** - Means that the User can see/edit the field in edit mode. (when editing a ticket or editing their profile)

This comment did point me in the direction of a bug with this pull request though. When the email field is hidden the profile will not update because the field is referenced in the update function. So I have to look into this part now.

fdellwing:
> That is what this pull request fixes...? If you want the User to not update their email address simply go to the form field and disable the Settings option "Editable For EndUsers". This will then hide that field for the Users. :+1:

But this is not that "not editable" means for me. Not editable means, the field is there and you can not change it. Without this PR, the field is a normal editable field, with this PR the field is not there at all. That is not what "not editable" means!!

I want my users to see there data, but not be able to change them (because they are managed from outside of osTicket through the database). That is not possible at the current moment and this PR does not fix that!

ntozier:
So your saying that this is a Feature Request not a Bug Report?

JediKev:
I'm sorry but that's how we invision it.

If you want to have the field non editable but still viewable in edit mode on the profile then you'll have to write your own custom code 🙂

Cheers.

JediKev:
addressed in a different way in subsequent versions

PR #3893, opened 2017-07-13
Set Session Cookie to HttpOnly

compuvin:
Proposing that the OsTicket session cookie be set to HttpOnly to mediate a cross-site scripting vulnerability. The details are outlined here: https://msdn.microsoft.com/en-us/library/ms533046.aspx
I have been testing this change locally and it have not noticed any issues.


fdellwing:
Please merge this!

PR #3896, opened 2017-07-14
Sort users feature

clonemeagain:
Forum user Buczko requested ability to sort tickets when viewing a user, mostly copied code from the tickets.inc.php page, chopped it around a bit, got it working, likely needs work, but presented here for possible cherry-pick inclusion. 

http://osticket.com/forum/discussion/90786/sorting-tickets


clonemeagain:
Hmm, didn't intend for that other commit to get lumped in, oops!

clonemeagain:
- [ ] add clickable headings to sort like normal tables.

Issue #3670, opened 2017-02-08
Reset button does not work: v1.10


balojs:
### Prerequisites

* [X] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [X] Do you have any errors in the PHP error log, or javascript console?
* [X] Did you check the [osTicket forums](http://osticket.com/forum/categories)?
http://osticket.com/forum/discussion/89444/delete-draft-and-reset-button-seems-slow-or-unresponsive
* [x] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/osTicket/osTicket/wiki/Github-Issues-Guidelines)

### Description

When typing in the box you text response and click on the reset button, it does not clear the contents of the text field

### Steps to Reproduce

1. Open a new Ticket
2. Write a some text on Response:
3. click on buttom "Reset"

**Expected behavior:**
Clean the content on Response field

**Actual behavior:** 
The content on Response field not reset

### Versions
v1.10-rc.3-1-gafd63e1

osTicket Version	v1.10-rc.3-1-gafd63e1 (afd63e1)  Upgrade — v1.10 is available
Web Server Software	Apache/2.4.6 (Red Hat Enterprise Linux) PHP/5.6.22
MySQL Version	5.5.44
PHP Version	5.6.22
PHP Extensions
gdlib	 Used for image manipulation and PDF printing
imap	 Used for email fetching
xml	 XML API
xml-dom	 Used for HTML email processing
json	 Improves performance creating and processing JSON
mbstring	 Highly recommended for non western european language content
phar	 Highly recommended for plugins and language packs
intl	 Highly recommended for non western european language content
fileinfo	 Used to detect file types for uploads
APCu	 Improves overall performance
Zend Opcache	 Improves overall performance
PHP Settings
cgi.fix_pathinfo	 "1" is recommended if AJAX is not working

browser version : Google Chrome Versión 56.0.2924.87

* [yes ] Do you have any errors in the PHP error log, or javascript console?

POST http://xxxx.xxxx.com/scp/ajax.php/lock/ticket/106 418 (I'm a teapot)
send @ jquery-1.11.2.min.js?afd63e1:4
ajax @ jquery-1.11.2.min.js?afd63e1:4
acquire @ ticket.js?afd63e1:58
(anonymous) @ ticket.js?afd63e1:247
each @ jquery-1.11.2.min.js?afd63e1:2
[...]

PR #3898, opened 2017-07-18
oops: Fix Reset Button(s)

JediKev:
This addresses issue #3670 where the reset button does nothing but return
an error. This adds a check for `data-draft-id` to the reset button. If
there is one it will delete the draft and reset the text box. If there is
not one then it will just reset the text box.

Also includes:
- A line to add the `data-draft-id` to the text box automatically, when a
draft is saved, through AJAX.


balojs:
thanks @JediKev

PR #3901, opened 2017-07-21
Fix internal notes not displayed on template

CarlosMoraRoa:
This is because the variable $info[] is overwrite when osticket do the foreach to show the templates, a simple fix is copy $info[] in another array to keep the tpl info.


JediKev:
@CarlosMoraRoa 

Thank you for this pull request. I have made a few changes with the following pull request:
- https://github.com/osTicket/osTicket/pull/5580

In your pull request, you are setting the original `$info` variable to `$infoTemplate` and allow the original `$info` variable to be overwritten by the `foreach(){}`. Instead, my PR proposes to prevent the original `$info` variable from being overwritten in the first place, so no need for a third variable or clone of the original `$info`.

Thanks again for the PR, we really appreciate it!

Cheers.

Issue #3903, opened 2017-07-25
Bug, osticket 1.10: "primary contacts" cannot be removed via gui


mfelber:
### Prerequisites

* [y] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [n] Do you have any errors in the PHP error log, or javascript console?
* [y] Did you check the [osTicket forums](http://osticket.com/forum/categories)?
* [y] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

### Description

When you add an primary contact to an organization it is saved correctly. When you try to remove it the change will not be written to the table in the database.

### Steps to Reproduce

1. Add an primary contact to an organization and click on update.
2. Remove the primary contact and update the organization:
![image](https://user-images.githubusercontent.com/5539456/28583379-3706e58c-7169-11e7-87f9-aac0855ca41b.png)
3. Navigate to the setting tab again and you will see that the change was not saved.

**Expected behavior:** The user should not be longer a primary contact for the organization.

**Actual behavior:** The user was not removed as a primary contact for the organization. The field "status" in the ost_user-table for the corresponding user is still set to the value "1" (which apparently means he is a primary contact).

### Versions

Server Information
--
osTicket Version | 1.8-git (?)  Upgrade— 901e5ea is available
Web Server Software | Apache
MySQL Version | 5.6.26
PHP Version | 5.6.12

gdlib | Used for image manipulation and PDF printing
imap | Used for email fetching
xml | XML API
xml-dom | Used for HTML email processing
json | Improves performance creating and processing JSON
mbstring | Highly recommended for non western european language content
phar | Highly recommended for plugins and language packs
intl | Highly recommended for non western european language content
fileinfo | Used to detect file types for uploads
APCu | Improves overall performance
Zend Opcache | Improves overall performance

cgi.fix_pathinfo | "1" is recommended if AJAX is not working
date.timezone | Europe/Berlin

Schema | osticket110 (localhost)
Schema Signature | 98ad7d550c26ac44340350912296e673
Space Used | 2.45 MiB
Space for Attachments | 0.01 MiB
Timezone | CEST (Interpreted as Europe/Berlin)

OS: ubuntu 14.04
Browser:  Chrome 59.0.3071.115, Safari 10.1.2 (12603.3.8)

JediKev:
I have made a pull request (#3904) that should fix your issue. Please test it and get back to me. Thanks!

[...]

PR #3904, opened 2017-07-25
orgs: Remove Primary Contacts

JediKev:
This addresses issue #3903 where you cannot remove Primary Contacts from
an Organization through the UI. This adds functionality to remove
collaborators from the select box.

### DIFF
--- a/include/class.organization.php
+++ b/include/class.organization.php
@@ -425,6 +425,12 @@ implements TemplateVariable {
                 $u->setPrimaryContact(array_search($u->id, $vars['contacts']) !== false);
                 $u->save();
             }
+        } else {
+            $members = $this->allMembers();
+            $members->update(array(
+                'status' => SqlExpression::bitand(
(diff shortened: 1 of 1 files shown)


include/class.organization.php
```diff
@@ -425,6 +425,11 @@ function update($vars, &$errors) {
                 $u->setPrimaryContact(array_search($u->id, $vars['contacts']) !== false);
                 $u->save();
             }
+        } else {
+            foreach ($this->allMembers() as $u) {
```
protich: Use `ORM` mass update instead of `foreach` loop.
```php
$members = $this->allMembers();
return $members->update(array(
   'flags' => SqlExpression::bitand(
        new SqlField('flags'), ~User::PRIMARY_ORG_CONTACT)
   ));
```

protich (changes requested):

include/class.organization.php
JediKev: Thank you! Updated now!

PR #3917, opened 2017-08-10
Fixes Datetime variable output in email templates.

clonemeagain:
For some reason when VariableReplacer is called on Ticket Form Template Tariables that are DateTime's, the FormattedDate constructor was converting the timestamp from "2017 Mar 17 00:10" format, through an `(int)` cast into just "2017", which wasn't bad enough to fail, so it happily worked, however the FormattedDate constructor expects a timestamp (GMT?) integeter, so the cast removes an error, it breaks the Variable.

The proposed patch simply converts the string into a timestamp via strtotime, which, in my tests, seems to fix the variable replacement issue.

Phew. that took some digging.

See original thread here: http://osticket.com/forum/discussion/90854/custom-form-with-date-time-field-gives-a-problem-when-using-as-variable-in-email#latest

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -1774,7 +1774,7 @@ class DatetimeField extends FormField {
 
     function asVar($value, $id=false) {
         if (!$value) return null;
-        return new FormattedDate((int) $value, 'UTC', false, false);
+        return new FormattedDate(strtotime($value), 'UTC', false, false);
     }
     function asVarType() {
         return 'FormattedDate';


clonemeagain:
Likely fixed better by storing the timestamp in the database instead.

egorrishe:
I think it is safer to save backward compatibility:
```
$v = (strtotime($value) === false) ? (int)$value : strtotime($value);
return new FormattedDate($v, 'UTC', false, false);
```

clonemeagain:
@egorrishe, I added that without critically examining it, reading it again it still has the original problem because:
A formatted date string won't return false from strtotime.
Eg: `strtotime("2017 May 2 12:34")` will return an integer timestamp, not false, so the unary true branch will then return `(int) "2017 May 2 12:34"` , which will return `2017`, which is a valid timestamp.. but obviously the wrong one. So we're back to the original bug. Doh.

clonemeagain:
Ok, new version works nicely and should retain any preexisting `(timestamp) $value` functionality.

It will be good when we can properly TypeHint methods!..

PR #3920, opened 2017-08-10
Only allow image attachments to be opened in the browser window

aydreeihn:
When opening attachments, we should only allow image files to be opened in the browser window. This prevents possible xss attacks where malicious code could be executed if an attachment were opened in the browser window. This was tested in Chrome, Safari, and Firefox.

### DIFF
--- a/include/class.http.php
+++ b/include/class.http.php
@@ -106,7 +106,7 @@ class Http {
     }
 
     function download($filename, $type, $data=null, $disposition='attachment') {
-        if (strpos($type, 'image/') !== 0)
+        if (strpos($type, 'image/') !== 0 || preg_match('/image\/.*\+.*/', $type))
           $disposition='attachment';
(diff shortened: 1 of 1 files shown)


perryk:
Apologies, I understand this relates to a third party plugin and not this PR, however, I have noticed that this change will break this plugin:  https://github.com/clonemeagain/attachment_preview .

JediKev:
That's a third party plugin. You'll have to take that up with the plugin developer.

perryk:
Sure, no troubles at all, just giving some details. I understand this PR is security related too.
Cheers.

JediKev:
@perryk No worries, just letting you know! Yes it is security related 😊

clonemeagain:
Cheers for the heads-up @perryk!
The plugin has now been patched to work around it.

PR #3922, opened 2017-08-13
trying to add ticket status column

idana0410:
Hello, I am trying to add ticket status column to my Osticket window where open tickets are. I got this link on a discussions forum and also there I got the instructions what to change in this function. Since it is my first time using github, I hope I am doing this the right way. I apologize for the inconvenience if I caused one.
Regards, Ida

### DIFF
--- a/include/staff/tickets.inc.php
+++ b/include/staff/tickets.inc.php
@@ -30,6 +30,8 @@ $sort_options = array(
     'closed' =>             __('Most Recently Closed'),
     'hot' =>                __('Longest Thread'),
     'relevance' =>          __('Relevance'),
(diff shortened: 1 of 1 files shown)


ntozier:
This does not appear to be an issue report... This is however probably the most asked question on the forums.

This thread has directions for most contemporary versions of osTicket.  
http://osticket.com/forum/discussion/75457/a-better-ticket-list

Alternatively there is the forthcoming "Custom Queues" feature here:
https://github.com/osTicket/osTicket/pull/2577

Personally I would just wait for the Custom Queues feature due out in 1.11, or use the feature from github in it's current alpha state.

Please close this issue report.

PR #3924, opened 2017-08-14
lock: Require Lock For Ticket Actions

JediKev:
This adds locks to ticket actions so that Agents have to posses a lock in
order to do any action on the ticket. This prevents separate Agents from
doing two different actions on the same ticket. Example: Closing a Ticket
and Editing a Ticket

**Lock Process:**
- If Agent A does not have a lock on the ticket, the system will
generate one and will allow Agent A to perform the action.
- If Agent B already has a lock on the ticket, the system will not generate
a lock for Agent A and will prevent Agent A from doing any actions.
- If Agent A had a lock on the ticket previously but it expired the system
will try to renew it.


JediKev:
>Could you extract the repeating part out into a method?

@clonemeagain which part are you talking about?

>Maybe I'm just lazy.. I do like the disabling of the buttons while it's locked, however, you would have to refresh to get the button back.

Thank you so much for this comment (you'll see why below).

Yes and No. 

**Yes** - As it stands now you would have to refresh the page in order to get the buttons back. 

**No** - This got me thinking. In theory you shouldn't have to refresh because every time you click the action button it runs the lock process. In that lock process it tries to make a lock for you and if there is no _valid_ lock on a ticket the system should create one. So I tested this and, in fact, even though the other agent's lock expired, the system didn't create a new one for me. It still said "Locked by someone else". This shouldn't happen..? So I looked at what determined the expiration of the lock and it led me to a function called `isExpired()`. This function compares two things: 
1.) `$this->getExpireTime()`
2.) `Misc::dbtime()`

1.) This returns the value of the `expire` column for the lock. (the expiration date/time)
2.) This function should return the database time right? Well it returned some whack ass time that was 5-6 hours ahead of me...wtf?? So I looked at this function (include/class.misc.php line 101) and it turns out the incorrect time is due to this line here:
```php
$D = DateTime::createFromFormat('U', $time);
```
This in theory should return the time in Unix format in your timezone but it always returned the time in the timezone of `+00:00`...ALWAYS. Even though my timezone is America/Chicago (-06:00 or -05:00 in DST). So I looked through PHP's documentation and people have pointed out that the `'U'` part (that specifies Unix format) is sometimes buggy when you pass an already unix timestamp in. When you do, it will return the `+00:00` timezone...awesome 🙄. So I then pondered the idea of do we even need that line at all since `time()` returns unix? In fact we do because of the `user2gmtime()` function returns a date string instead of unix. So I moved the `createFromFormat(..` line under the `user2gmtime()` function and changed `$D` to `$time`. Then I changed `$dbtz->getOffset($D)` to `$dbtz->getOffset($time)`. This way we should get the correct time no matter what we throw at the function (unix or regular date/time string). I then tested the locks again by creating a lock for myself (clicking an action), changing the lock to point at a different agent and tried the action again. This gave me "Locked by someone else." Which is good. So then I changed the expire date to be expired, tried the action again (without refreshing) and VOILA it works! 

Needless to say I made a pull request for this too! 😄
So, if you use this pull along with this one #3933 then you'll be good! 👍 

**Working Example:**
locks

clonemeagain:
The repeated bit is here: https://github.com/osTicket/osTicket/pull/3924/files
Scroll down, the same section of code appears in seven places, making it ideal for refactoring into a method. 

Is the expire time not a timestamp? Strewth. Good hunting buddy!

JediKev:
@clonemeagain It is now updated! Thanks mate! 😁

PR #3931, opened 2017-08-22
format: Mailto TLD Length

JediKev:
This addresses an issue where the text formatter only allows 2-4
characters in email domain TLDs. This causes an issue if you have a TLD
with more than 4 characters, where the full email address will not be
converted into the link. This increases the allowed characters to 2-63
characters allowing all modern TLDs to be converted correctly. Why 63
chars? Have a look at the 2.3.4 Size Limts section here:
http://www.ietf.org/rfc/rfc1035.txt

**Before**
`<a href="mailto:user@domain.tech">user@domain.tech</a>nology`
<a href="mailto:user@domain.tech">user@domain.tech</a>nology

**After**
`<a href="mailto:user@domain.technology">user@domain.technology</a>`
<a href="mailto:user@domain.technology">user@domain.technology</a>


clonemeagain (approved):
Seems good.

PR #3933, opened 2017-08-25
oops: Fix Incorrect dbtime() Conversion

JediKev:
This addresses an issue where the `dbtime()` function returns the
incorrect time when calling it without a timestamp. When you call it
without a timestamp it creates the time with the `time()` function. The
`time()` function returns the time in Unix format. At the end of
`dbtime()` it converts the time into Unix format with
`DateTime::createFromFormat('U', $time)`. This is buggy when you pass an
already unix timestamp and will return the time in the timezone of +00:00
. This moves the lines around so that if the time is already Unix it will
not convert to Unix, resulting in the correct time.

### DIFF
--- a/include/class.misc.php
+++ b/include/class.misc.php
@@ -108,14 +108,14 @@ class Misc {
         } else {
             // User time to UTC
             $time = self::user2gmtime($var);
+            // UTC to db time
+            $time = DateTime::createFromFormat('U', $time);
         }
 
         if (!isset($dbtz)) {
             $dbtz = new DateTimeZone($cfg->getDbTimezone());
         }
-        // UTC to db time
-        $D = DateTime::createFromFormat('U', $time);
-        return $time + $dbtz->getOffset($D);
+        return $time + $dbtz->getOffset($time);
     }
 
     /*Helper get GM time based on timezone offset*/


perryk:
Hello JediKev,
Would you possibly be able to explain a little further why this PR is closed ?
I've read your PR 3924 also, however, I've found that this closed PR sometimes helps with some timezone/ticket locks troubles I am currently having.
I'm seeing a range of troubles however so it is hard to pinpoint the exact problem or explain it well.
Thanks

JediKev:
@perryk 

Yes, I closed this because at the time of making another feature I was using older code, and I was having timezone issues. So, I fixed the time issue for the old code and pushed to here. After I pushed it I realized that I was using old code, I then updated my code to the latest on git and all the time issues went away.

>Whilst I am still trying to figure out this timezone problem it is currently looking like a javascript implementation issue and Chrome is giving me an incorrect timezone whilst Firefox and Chromium are returning the correct timezone. Talk about weird!!

That is _very_ weird...what time is off? Everything in the system or the ticket times, the task times, etc.? It could also be an issue with browser caching, have you tested your issue using Chrome's Incognito Window? Because I've noticed that even when I clear all my cache for the site it still has some cached items on the site...strange I know.

perryk:
Ok, now this is no longer looking to me like a browser specific bug, instead, it looks like the different browsers were just storing different session details in the ost_session table. Phew.

I'm still no further on obtaining a way to always ensure Australia/Brisbane is returned however I have a little more to go on now so hopefully will figure it out soon.

Issue #3942, opened 2017-08-29
osTicket Doesn't Accept OneNote Links


arthurlt:
### Description

When adding a OneNote link prefaced with `onenote:` there's a `denied:` added to the front. These OneNote links link straight to a OneNote Notebook. This looks like an easy fix, just the `safe_html` function updated in the file `include/class.format.php`

#2950
#3691

PR #3943, opened 2017-09-01
added support for OneNote links

arthurlt:
#3942 

Consider this a fix for that issue.


arthurlt:
I actually haven't had a chance to test it and this fix may not work as OneNote links go something like this:
`onenote:https://website.com/notebook/section.one`

PR #3955, opened 2017-09-12
login: Fix CSRF fail, add shake effect on authentication fail

greezybacon:
It's possible I'm the only user annoyed by this issue. Everyday when I come into work and start working on tickets, I click somewhere in the interface and am presented with the login screen. After authenticating, I'm presented with the CSRF failure message and receive a database error email for DB Error 1062 (Duplicate entry '[blah]' for key 'PRIMARY').

This fixes an issue where the current session data is not retrieved from the database, because it is expired. However, the session-id is not reset either. Therefore, the session data with the new CSRF token is not updated in the database and the user may have trouble logging in. (This is where the db error email comes from).

This problem manifests itself as a session expires. If a user clicks somewhere in the software and their session is now expired, a redirect to the login page is triggered. However, the CSRF token sent in the login page is not saved in the database. So when the user logs in, they are greeted with the CSRF failure message.

This issue is addressed by retrieving the session data from the database, but clearing the content if it is expired. Therefore, the session data appears to the software as invalid and is properly reset and saved to the database, thereby avoiding the errors.

This also changes the staff authentication process to use AJAX to authenticate (with a failsafe of using traditional page requests). If the authentication fails on the server because of CSRF problems, the page is refreshed in the browser fetching a new CSRF token. If the authentication fails because of password issues, the login box shakes and the user can retry. As before, the `forgot password` link shows up after the first failure. Upon success, a redirect is issued into the staff control panel.

This should not affect external authentication plugins such as LDAP or OAUTH, but I have not yet tested them.

### DIFF
--- a/include/class.ostsession.php
+++ b/include/class.ostsession.php
@@ -178,10 +178,15 @@ extends SessionBackend {
 
     function read($id) {
         try {
-            $this->data = SessionData::objects()->filter([
-                'session_id' => $id,
-                'session_expire__gt' => SqlFunction::NOW(),
-            ])->one();
+            $this->data = SessionData::objects()
+                ->filter(['session_id' => $id])
+                ->annotate(['age' => SqlFunction::NOW()->minus(new SqlField('session_expire'))])
+                ->one();
+            if ($this->data->age > 0) {
+                // session_expire is in the past. Pretend it is expired and
+                // reset the data. This will assist with CSRF issues
+                $this->data->session_data='';
+            }
             $this->id = $id;
         }
         catch (DoesNotExist $e) {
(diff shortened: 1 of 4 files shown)


kevinoconnor7:
This patch appears to have introduced a bug to the session code which breaks my CAS Authentication plugin.

Your change to the `read` method in `DbSessionBackend` adds an annotated field, which in turn gives us back an `AnnotatedModel`.

The problem is that in the `update` method reads the field `__new__` on the model. However, this now goes through the `AnnotatedModel` magic getter which tries to read the field from the database result, rather the field on VerySimpleModel.

My quick-and-dirty fix was to update AnnotatedModel to first check if the field exists on the underlying object by updating the [getter](https://github.com/osTicket/osTicket/blob/22cb8611c280e4e71528a8018ff0e8e67bd30413/include/class.orm.php#L688) to as:

```php
    function get($what) {
        if (isset($this->annotations[$what]))
            return $this->annotations[$what];
        if (property_exists($this->model, $what))
            return $this->model->{$what};
        return $this->model->get($what, null);
    }
```

This seems a bit messy though. Instead `VerySimpleModel` should probably have a isNew method and it shouold be used in the DbSessionBackend.

PR #3962, opened 2017-09-14
xss: Prevent Shared XSS Attachments

JediKev:
This addresses an issue where a User can upload an XSS infected
attachment to a ticket, share the file link with another person, and the
person will be able to view/download the attachment. This adds a check to
see if the attachment is apart of a ticket and if it is then only Users
associated with the ticket can view the attachment. You have to enable
the setting ‘Login Required to view attachments’ for this to work.

### DIFF
--- a/file.php
+++ b/file.php
@@ -39,6 +39,18 @@ if ($cfg->isAuthRequiredForFiles() && !$thisclient) {
             require 'secure.inc.php';
         }
     }
+} elseif ($cfg->isAuthRequiredForFiles() && $thisclient) {
+    // If the file is a ticket attachment, only let Users associated with the
+    // ticket view the file. Helps prevent shared XSS attachments.
+    $attachment = $file->attachments->findFirst();
+    if (($thread = $attachment->getObject()->getThread()) && $thread->getObjectType() == 'T') {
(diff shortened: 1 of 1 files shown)


file.php
```diff
@@ -39,6 +39,18 @@
             require 'secure.inc.php';
         }
     }
+} elseif ($cfg->isAuthRequiredForFiles() && $thisclient) {
```
JediKev, line 42: Combine into:
```
if ($cfg->isAuthRequiredForFiles()) { 
    if ($thisclient) {
        xxx
    } else {
        xxx
    } 
}
```

file.php
```diff
@@ -39,6 +39,18 @@
...
+    // If the file is a ticket attachment, only let Users associated with the
+    // ticket view the file. Helps prevent shared XSS attachments.
+    $attachment = $file->attachments->findFirst();
+    if (($thread = $attachment->getObject()->getThread()) && $thread->getObjectType() == 'T') {
+        $recipients = $thread->getObject()->getRecipients();
+        foreach($recipients as $r) {
```
JediKev, line 48: ORM lookup is faster

PR #3968, opened 2017-09-19
use i18n to show open/closed status

ropaha:
The tasks assigned to a ticket doesn't use i18n to show there status in selected language pack.
osticket-issue


ntozier:
Technically this is not a bug.  The translation effort is on like phase 3 of 4.  They haven't done that part yet.

PR #3977, opened 2017-09-27
feature: Create Task Ticket Thread Action

JediKev:
This adds a Ticket Thread Action for creating a Task. This action will
popup the New Task modal and will pre-fill the body of the New Task with
the body of the Thread Entry from which the action was made from.

**Example:**
Thread Entry - “I will need to check on the status of your account.”
Click Create Task from that Thread’s Actions
System will show New Task popup with the body as “I will need to check on
the status of your account.”


include/class.thread_actions.php
```diff
@@ -366,3 +366,42 @@ protected function trigger__post() {
...
+
+    function getJsStub() {
+        return sprintf(<<<JS
+var url = '%s';
+var redirect = $(this).data('redirect');
+$.dialog(url, [201], function(xhr, resp) {
+    if (!!redirect)
+        $.pjax({url: redirect, container:'#pjax-container'});
+    else
+        $.pjax({url: 'ajax.php/tickets/378/tasks', container: '#tasks_content', push: false});
```
greezybacon: oops. Should the ticket-id be replaced here?

JediKev: Doh. Yessir. Thank ya 😉

PR #3996, opened 2017-10-11
issue: Check Missing Required Fields

JediKev:
This addresses an issue where an integrity field that’s disabled by a Help
Topic will still prevent the closing of a ticket. This is due to the check
for integrity fields on a ticket that have no data. The check includes the
fields that are disabled by Help Topic. This updates the check to
correctly exclude the disabled fields so that the ticket can be closed.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -310,7 +310,7 @@ implements RestrictedAccess, Threadable {
             return true;
 
         $warning = null;
-        if ($this->getMissingRequiredFields()) {
+        if (self::getMissingRequiredFields($this)) {
             $warning = sprintf(
                     __( '%1$s is missing data on %2$s one or more required fields %3$s and cannot be closed'),
                     __('This ticket'),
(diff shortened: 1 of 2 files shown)


perryk:
I'm receiving some PHP warnings when I have this applied.

array_filter() expects parameter 1 to be array, null given in /var/www/html/osticket/include/class.ticket.php on line 987, referer: https://domain.com/scp/tickets.php?status=open

JediKev:
@perryk is the conflicting line `$extra = array_filter($extra);`? If so I can see where the issue could pop up and I know how to fix it.

perryk:
Yes, that is line 987 for me.

JediKev:
@perryk Alright. I'll push the change tomorrow and let you know! Cheers.

perryk:
Great, thanks JediKev.

JediKev:
@perryk I just pushed the change, could you re-test this for me?

Basically, the issue was I was using `array_filter()` which filters out any empty values in an array. This has to be passed an array otherwise it'll freak out. I wasn't checking before the filter if the value was an array or not. So I updated the pull to remove `array_filter()` and just check the entry's `disable` key for emptiness. Much cleaner and should not throw an error lol 👍

perryk:
Ok tested and all good, no more line 987 errors, thanks!

Issue #3979, opened 2017-09-28
Choose correct SLA for the choosen category if agent opens ticket


fdellwing:
![osticket1](https://user-images.githubusercontent.com/22813377/30969132-c2d45272-a461-11e7-862e-eaa53d72a850.png)

Expected behaviour:

1. Click "New ticket"
2. Choose any category that has a not default SLA
3. The SLA should be automatically changed

Actual behaviour:

3. The SLA stays on system default

iHunt101:
I think you can set this to your form.

fdellwing:
Please read the issue (point 2.) ;)

I did this, and it is working fine if a user opens the ticket. But it doesnt work if an agent opens the ticket.

JediKev:
@fdellwing Thanks for the report. I fixed it with [this pull request](https://github.com/osTicket/osTicket/pull/4012). Please pull the changes, test, and let me know how it goes! Cheers.

fdellwing:
Works :+1:

JediKev:
@fdellwing Nice! Again, thanks for the report! Cheers.

akash65:
SLA plan is not changing when i assign ticket to other agent..why?

ntozier:
Agents do not have SLAs.  Why would it change if you assigned it to another agent?

PR #4012, opened 2017-10-18
issue: Help Topic SLA

JediKev:
This addresses issue #3979 where the system uses the default SLA plan if
you don’t choose one in the creation process, even though the selected
Help Topic has one explicitly set. This is due to the method that checks
for an SLA plan selection upon ticket creation. The method uses `isset()`
to determine whether or not the agent selected an SLA Plan. If you don’t
select an SLA plan the `slaId` will return `0` which, even though equals
nothing, is still set. If the `slaId` is set then it will use that SLA
plan or the default one. Since `0` is not an SLA plan, it will use the
default SLA. So this adds a check beforehand to see if `slaId` is `0`. If
so it will unset `slaId` and use either the Help Topic SLA or the default
SLA.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -3321,6 +3321,10 @@ implements RestrictedAccess, Threadable {
             elseif (!isset($vars['teamId']) && $topic->getTeamId())
                 $vars['teamId'] = $topic->getTeamId();
 
+            // Unset slaId if 0 to use the Help Topic SLA or Default SLA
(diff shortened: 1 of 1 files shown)


perryk:
Sorry, I receive a merge conflict with this if I have your #3915 PR already applied.

JediKev:
Oo yea I need to delete that one. Use this one instead it's much better.

perryk:
Thanks, will do.

PR #3456, opened 2016-11-03
Fix for LDAP/AD auth plugin

ntozier:
removed:
87: ->annotate(array('__relevance__' => new SqlCode(1)))

This fixes Lookup Create new User Search on open ticket.

PR #4013, opened 2017-10-18
Fixes & Updates

### LINKED ISSUE #3994: Object of class could not be converted to string
### Description
osTicket v1.10.1 throws php error "Object of class could not be converted to string" in include/class.auth.php line 316 
if there is more then one plugin enabled that supports AuthDirectorySearch.

function getSearchDirectories() has 2 loops, one searching for plugins that handle staff authentication, and the other searching for plugins that handle user authentication.  If you have 1 plugin handling each (one staff, and a different one for users) you get the class convert to string error.   However if you have only one or the other enabled everything works fine.

switching the order of the loops  changes what class get's reported in the error.

### Steps to Reproduce

1. install osticket
2. install and enable 2 auth plugins that support  AuthDirectorySearch
3. try and create a new ticket.
4.  catchable fatal error

**Expected behavior:** [What you expected to happen]
give me the search dialog and allow me to search for users

**Actual behavior:** [What actually happened]
blank search dialog and error in apache errorlog

### Versions
osTicket v1.10.1
linux centos 7 php56, browser doesn't matter

Also, please include the OS and what version of the OS you're running. As well as your browser and browser version.

wnoisephx:
Closes #3994 & #3456

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,7 @@ stage
 
 # Ignore packaged plugins and language packs
 *.phar
+include/plugins
 
 # Ignore mPDF temp files
 include/mpdf/ttfontdata
--- a/include/ajax.users.php
+++ b/include/ajax.users.php
@@ -77,7 +77,6 @@ class UsersAjaxAPI extends AjaxController {
                     'emails__address__contains' => $q,
                     'name__contains' => $q,
                     'org__name__contains' => $q,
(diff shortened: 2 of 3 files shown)


ntozier:
"adds backed_id to user cdata which will allow posting back of data to the backend"

Are you saying that you are saving information to a cdata (temporary) table?

PR #4014, opened 2017-10-18
issue: Clone Method Called On NonObject

JediKev:
This addresses an issue reported on the Forums where searching for
something by date will throw a PHP fatal error. (eg. Closed Date -> on ->
10/11/2017) This is due to the date value that was chosen is not a
DateTime object. The code that determines the method to search by (has a
value, before, on, etc.) for DateTime fields will sometimes clone the date
value. In order to clone the date value it has to be an object. This
updates the method that sets the date value to always return a DateTime
object for the date value.

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -2029,7 +2029,7 @@ class DatetimeField extends FormField {
         $config = $this->getConfiguration();
         $value = is_int($value)
             ? DateTime::createFromFormat('U', !$config['gmt'] ? Misc::gmtime($value) : $value) ?: $value
-            : $value;
+            : (is_object($value) ? $value : new DateTime($value));
         switch ($method) {
         case 'equal':
             $l = clone $value;


nicoletta-maia:
Thanks, this works for me!

micjes:
Thanks, this solved the problem for me :-)

ramrajone:
Thank you, this worked for now
Line 2032 in include/class.forms.php:
replacing: 
: $value; 
with: 
:(is_object($value) ? $value : new DateTime($value));

RikkiMasters:
Thank you this worked.

leenukes:
Worked for me too, thanks for the code @ramrajone

leenukes:
An update, I just updated to 1.10.7 and this is still an issue so I'm going to use this fix again although this change is now on line 2061 but it works the same.

JediKev:
addressed in subsequent versions

PR #4018, opened 2017-10-23
Support for mime encoding in Reply-To field

dukei:
Otherwise using reply-to field for creating ticket results in spoiled username


JediKev:
Do you have an example with "before and after" results? I would love to see what would become "spoiled".

dukei:
Yes. When Reply-to contained cyrillic, for example, and there was a rule to use Reply-To header then the user name looked like `=?utf-8?Q?=D0=BF=D1=80=D0=B8=D0=B2=D0=B5=D1=82?=`
And with this patch it looks like it should: `привет`

JediKev:
Ah, superb. Thanks for the example and pull request!

PR #4033, opened 2017-10-31
issue: Disable FAQ

### LINKED ISSUE #4032: Disabled FAQ problems
v1.10.1 (9ae093d)
v1.10-78-gacac370 (acac370)

When I disable FAQ or FAQ is for registered only (and accessing as a guest), I can see Featured Questions on the landpage.
Expected: This section shall be hidden too.
wtf.

Btw. in the Settings, this options may be replaced by radiobutton or select (Off/registered only/public), or 2nd option shall be unavailable when FAQ is disabled:
faq

jdelhome3578:
This fixes #4032 where disabling the FAQs does not hide the featured questions on the client landing page.

### DIFF
--- a/include/client/templates/sidebar.tmpl.php
+++ b/include/client/templates/sidebar.tmpl.php
@@ -17,7 +17,8 @@ $BUTTONS = isset($BUTTONS) ? $BUTTONS : true;
                 echo __('Check Ticket Status');?></a>
 </p>
         </div>
-<?php } ?>
+<?php } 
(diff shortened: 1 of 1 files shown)


perryk:
This gave me a merge error when I already had #3526 applied.

It can be resolved but then the code is a little odd in that it checks if the FAQs are disabled twice.

Not really a big problem, the older one has another option changed that people may not want, just adding a note about this as something to be aware of.

Cheers.

jdelhome3578:
PR modified by @jedikev #4416

Issue #4058, opened 2017-11-08
Staff Ticket Sorting by Last Updated


jdelhome3578:
![image](https://user-images.githubusercontent.com/5774055/32551192-b5345a74-c45d-11e7-96e0-9bc6274f0e79.png)

Last Updated Column URL with default sort applied: `https://helpdesk/scp/tickets.php?sort=priority,updated&dir=1&status=open`

**Expected behavior:** I would expect the URL to have `sort=lastupdated` so that the column could be sorted

**Actual behavior:** Table is sorted by reversed Priority and last updated

### Versions

Server Information
--
osTicket Version | v1.10.1 (9ae093d) —  Up to date
Web Server Software | Microsoft-IIS/8.5
MySQL Version | 5.7.20
PHP Version | 5.6.31

jdelhome3578:
PR #4060 fixes this issue.

PR #4060, opened 2017-11-08
issue: Last Updated Sorting

JediKev:
This addresses issue #4058 where the Last Updated column header's sorting
does not function properly. It adds `sort=priority,updated` to the query
which breaks the sorting for the column. This updates the column's sorting
option to `updated` which should give you the proper sorting for the
header.

### DIFF
--- a/include/staff/tickets.inc.php
+++ b/include/staff/tickets.inc.php
@@ -345,7 +345,7 @@ case 'priority,updated':
     // Fall through for columns defined for `updated`
 case 'updated':
     $queue_columns['date']['heading'] = __('Last Updated');
-    $queue_columns['date']['sort'] = $sort_cols;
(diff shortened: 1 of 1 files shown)


jdelhome3578 (approved):

jdelhome3578:
@JediKev I thought that was the fix but wasn't sure.

esamo:
It doesn't work for me - the 'Last Updated' header link has the same param 'priority,updated' instead of 'updated'. This is on Open tickets list.
`/scp/tickets.php?sort=priority,updated&dir=1&status=open`

nicoletta-maia (approved):

JediKev:
addressed in subsequent versions (custom columns/custom queues)

Issue #3741, opened 2017-03-13
manage.php deploy path problems


dmnc-net:
### Description

Errors during CLI deployment.

### Steps to Reproduce
Follow README.md Deployment section:

> And deploy the code into somewhere in your server's www root folder, for instance
> ```
> cd OsTicket
> php manage.php deploy --setup /var/www/htdocs/osticket/
> ```

**Expected behavior:**
No error

**Actual behavior:**
```
# php manage.php deploy --setup /var/www/osticket-latest/
PHP Warning:  include(/var/www/osticket-latest//bootstrap.php): failed to open stream: No such file or directory in - on line 2
PHP Stack trace:
PHP   1. {main}() -:0
PHP Warning:  include(): Failed opening '/var/www/osticket-latest//bootstrap.php' for inclusion (include_path='.:/usr/share/php:/usr/share/pear') in - on line 2
PHP Stack trace:
PHP   1. {main}() -:0
PHP Notice:  Use of undefined constant INCLUDE_DIR - assumed 'INCLUDE_DIR' in - on line 3
PHP Stack trace:
PHP   1. {main}() -:0
```

Second run of the command didn't produce an error message at all.

**Running on:**

PHP 5.6.30-0+deb8u1 (cli) (built: Feb  8 2017 08:50:21) 
Copyright (c) 1997-2016 The PHP Group
Zend Engine v2.6.0, Copyright (c) 1998-2016 Zend Technologies
    with Zend OPcache v7.0.6-dev, Copyright (c) 1999-2016, by Zend Technologies
    with Xdebug v2.2.5, Copyright (c) 2002-2014, by Derick Rethans

webunraveling:
Here is, what appears to be, the same issue but on a server running PHP7.

Output of install script:

    $ php setup/cli/manage.php deploy --setup /home/userName/public_html/my/install/dir/osTicket/
    #!/usr/bin/env php

Resulting `error_log` file

    [14-Aug-2017 16:16:46 UTC] PHP Warning:  include(/home/userName/public_html/my/install/dir/osTicket//bootstrap.php): failed to open stream: No such file or directory in - on line 2
    [14-Aug-2017 16:16:46 UTC] PHP Warning:  include(): Failed opening '/home/userName/public_html/my/install/dir/osTicket//bootstrap.php' for inclusion (include_path='.:/opt/alt/php70/usr/share/pear') in - on line 2
[...]

PR #4101, opened 2017-11-29
deploy: Failed Bootstrap Inclusion

JediKev:
This addresses issue #3741 where deploying osTicket from the command line
will throw a fatal error the first time ran. The error is caused by a
failed inclusion of the `bootstrap.php` file. This is because it’s trying
to include the file from the destination folder which doesn’t exist there
yet. This updates the deploy method to use the source bootstrap file
instead of the destination bootstrap file which already exists.

### DIFF
--- a/include/cli/modules/unpack.php
+++ b/include/cli/modules/unpack.php
@@ -210,7 +210,7 @@ class Unpacker extends Module {
         ), $pipes);
 
         fwrite($pipes[0], "<?php
-        include '{$this->destination}/bootstrap.php';
+        include '{$this->source}/bootstrap.php';
         print INCLUDE_DIR;
         ");
         fclose($pipes[0]);


perryk:
Sorry, this generates an error for me.

I receive the following when I try to use `./manage deploy <blah>` after this is applied.

~~~
PHP Warning:  include(/bootstrap.php): failed to open stream: No such file or directory in - on line 2
PHP Warning:  include(): Failed opening '/bootstrap.php' for inclusion (include_path='.:/usr/share/pear:/usr/share/php') in - on line 2
PHP Notice:  Use of undefined constant INCLUDE_DIR - assumed 'INCLUDE_DIR' in - on line 3
~~~

I'm unsure which folder exactly $this->source is pointing to. I've not really looked at it in great detail.

As an aside, I've never been able to do any ./manage.php **file** <blah, even with help for example> commands. They always return `<html></html>` and never work. I understand this may have nothing to do with this PR, I've just been meaning to mention it, perhaps it is a Centos thing ?

JediKev:
@perryk Run the command again except run as `sudo`. Make sure you haven't already tried to deploy to the directory you deploy to as it will always work the second time. We need to test the first try.

PR #4106, opened 2017-12-01
update to remove setup and ost-config.php

deltadan:


deltadan:
bad request ignore - wrong branch.

PR #4109, opened 2017-12-07
Fix image pasting and issue with repeated pasting too fast

### LINKED ISSUE #3481: v1.10 - Image PASTE not working - from filesystem
I took a screenshot from an app and I wanted to paste the screenshot to the ticket.
There is a small "progress bar" rather than a spinning wheel - which means data is being uploaded....
but the image does not appear on the edited ticket's content :(

### LINKED ISSUE #3669: Paste screenshot into redactor from clipboard breaks
### Description

We cannot paste a screen grab/ screenshot directly into the redactor wysiwyg html editor.

We can upload images through the "Upload Image" icon and also drag and drop existing ones from i.e. my desktop, but a lot of our clients are trying to paste a screenshot they have copied to their clipboard directly into the editor using (ALT+PRINT SCREEN) and then (CTRL + V) on windows.

This works if it is the first thing you do, once you start with a fresh ticket. But as soon as there are other text in the editor, it only brings up a loading image and then time-out.

I have tried to track the error down  and think it has to do with the compatibility between osTicket's code and redactor. Becasue I tested the redactor wysiwyg and it works fine on their website.

I even tried to purchase the latest redactor js files, but was not able to merge it into the latest version of osTicket. There were a lot of new errors popping up.

Does anyone know what the solutions is here and when we can expect the devs to release a bug fix?

### Steps to Reproduce

1. Open a new ticket
2. Type something in the box and press enter.
3. Take a screenshot (ALT+PRINT SCREEN) and then paste it in the editor (CTRL + V) on windows

**Expected behavior:**
It is suppose to paste the image because redactor allows for this functionality.

**Actual behavior:**
But it does not paste it. Sometimes it might, but then as you type anything else, it will stop pasting images in.

### Versions
Issue on both v1.10-rc.3 & v1.10
Web Server Software	Apache/2.4.7 (Ubuntu)
MySQL Version	5.7.16
PHP Version	5.5.9-1ubuntu4.20

### Browsers
We are experiencing this accross all browsers.

SaraivaDaniel:
Fix #3669 Fix #3481 

I've managed to fix two Redactor issues regarding:

- pasting of images
- issue when pasting repeatedly too fast where one could not break a line inside the pasted text (type some string, select it, paste it again multiple times fast - put the cursor somewhere in the middle of the pasted text - if you press Enter, the line doesn't break)

**Limitation**
The pasted image is saved inline as a data URI

The Redactor was unpacked using Chrome pretty-print (to preserve some unicode chars that were being discarded using js-beautify).

### DIFF
--- a/include/client/header.inc.php
+++ b/include/client/header.inc.php
@@ -47,9 +47,9 @@ if ($lang) {
     <script src="<?php echo ROOT_PATH; ?>js/osticket.js"></script>
     <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/filedrop.field.js"></script>
     <script src="<?php echo ROOT_PATH; ?>scp/js/bootstrap-typeahead.js"></script>
-    <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/redactor.min.js"></script>
-    <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/redactor-plugins.js"></script>
-    <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/redactor-osticket.js"></script>
+    <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/redactor.unpacked.js?1"></script>
+    <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/redactor-plugins.js?1"></script>
+    <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/redactor-osticket.js?1"></script>
     <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/select2.min.js"></script>
     <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/fabric.min.js"></script>
     <?php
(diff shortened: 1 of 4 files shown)


JediKev:
Woooowwww. So basically the minified file had some screwed up Unicode chars and replacing the minified file with the original file worked?

SaraivaDaniel:
@JediKev actually no, those characters doesn't seem to affect it or have any relation with the issues, but for a chance I've noticed the difference of Chrome generated JS and js-beautify code.

I guess the problem is related to Chrome (and Edge) - they don't paste an image to a div with `editableContent="true"` (see this [fiddle](https://jsfiddle.net/6uj93zuc/)). IE and Firefox works. The Redactor plugin pasted the content to and dynamically generated *pasteBox*  - but in Chrome and Edge, the content failed to actually paste in this box...

The other problem was fixed just by iterating the replacement of span.redactor-invisible-space, that redactor would insert in the HTML when pasting text (but it counted in timeouts to remove them, and one user can be faster than the timeout)

SaraivaDaniel:
I edited the initial comment to reference the issues regarding this and to make note of a limitation (that the images are saved inline).

gitbelstar:
@SaraivaDaniel Thanks - this fixes the issue for us. We can consistently paste images successfully now.

mlipok:
#3481 

> For now I can confirm that this problem is not occuring any more from time when we install v1.14.x

Issue #3703, opened 2017-02-21
Advanced Search Broken for Custom Form Fields of Type Choice and Custom List


jesse-thomas:
When using advanced search in the Agent console to search tickets with info from custom forms of the type Choice or Custom List no results are returned. 

### Steps to Reproduce all done in the Agent Console

1. Create a custom form, add a field of type choice, add some choices (Yes, No)
2. Add the form to a help topic
3. Create a ticket with that topic
4. Edit the ticket and select a value for the field (Yes or No), save.
5. Perform an advanced search for the Custom Form / Field  / includes value (yes or no)

Same thing when you use a custom list instead of the Choice field

I would expect to get back all tickets that had the matching value for the field

Instead I get back no results

osTicket Version	v1.10 (901e5ea) —  Up to date
Web Server Software	Apache
MySQL Version	5.6.33
PHP Version	5.5.9-1ubuntu4.21

Sever: Ubuntu 14.04
Browser: Google Chrome.

Additionally if I search for does not include i get tickets back that acutall have the value set.

parijathakumar:
I am also facing the same issue. My system details are ...

Debian -8 Jessie
PHP-5.6.30
MySQL-5.5.55
Apache-2.4.10

JediKev:
I have created this pull request that should solve this issue: #4124

**Note:** If you're using a Choices field, every choice HAS to have a key associated with them (mentioned in the helptext).
<img width="320" alt="screen shot 2017-12-14 at 13 38 51" src="https://user-images.githubusercontent.com/11823401/34011133-9c9485be-e0d4-11e7-8b2e-c2ac35c88e4e.png">

jesse-thomas:
@JediKev Thanks for letting me know. I just added the "fix" I used several months ago, please let me know if you think its any good. Also I am not using keys with my options. #4127

JediKev:
@jesse-thomas So I just tested your pull request and it gave me about 11 records that had nothing to do with the search. When I searched with my method it gave me 13 (correct) results and when I tested yours it gave me 24 (13 - correct || 11 - incorrect) results. I'm not sure what yours is doing but it's including results that shouldn't be there. Cheers.

Issue #3493, opened 2016-11-10
Advenced search not  working


edukondalugorla:
1.10 RC3 advanced search not matching "includes" on "choices" form fields

jdelhome3578:
Does PR #3495 fix your issue?

GoodbyeWorld:
Same issue, upgraded from 1.10 RC1, to RC2, to RC3, to stable release.  Still having this problem.

GoodbyeWorld:
This is a duplicate for #2625

sieveni91:
I would like to use advanced search, but the problem is still existing. Is there a chance for a fix?

JediKev:
I have created this pull request that should solve this issue: #4124

**Note:** If you're using a Choices field, every choice HAS to have a key associated with them (mentioned in the helptext).
<img width="320" alt="screen shot 2017-12-14 at 13 38 51" src="https://user-images.githubusercontent.com/11823401/34011117-91f4e75c-e0d4-11e7-9a81-0b2601e359c6.png">

Issue #2625, opened 2015-10-06
Advanced search with custom choices field not working (1.10rc2)


jamesangi:
I have a custom form with a field of type "Choices."  I cannot use Advanced Search on this field.  For example, I have two closed tickets with "Bill" selected in this field but my search returns no results.

Here are some screenshots:
![billing 1](https://cloud.githubusercontent.com/assets/14003656/10310995/c893fde0-6c12-11e5-9b41-2b8b4136ad85.PNG)
![billing 2](https://cloud.githubusercontent.com/assets/14003656/10310998/cc8d9e56-6c12-11e5-91a1-8d1f0259d97b.PNG)
![billing 3](https://cloud.githubusercontent.com/assets/14003656/10311000/cd91e0d2-6c12-11e5-999d-d0c2abf66119.PNG)

jamesangi:
No one else seeing this?  I tried a clean install and it still happens, nothing interesting in the logs.  I've noticed a couple more advanced search oddities as well:
1) When loading a saved search it "shakes" the window back and forth like there is an error, except there is none.
2) Sometimes after clicking "search" you then have to click on the "search" tab to get the results to display.

Animeonmymind:
Same situation here, custom field searches yield no results.

greezybacon:
I was able to duplicate this yesterday. I think the problem is that the search system builds a query with
```sql
<field> IN ('BILL')
```

however, the storage model for the Choices field looks like `{"Bill": "Bill"}`, because it stores the key and the value. In the event the field is changed to reflect different keys/values, the previously set value can be retained. Therefore, the SQL query needs to be built like
```
<field> LIKE '%"Bill"%'
```

mbaghert:
custom field search not result

ntozier:
@mbaghert Huh?

bpalme:
I would appreciate any hints on how to patch this until a real fix is developed.  If you can point me in the right direction, I might be able to come up with something.

ntozier:
@bpalme on Oct 22nd Jared posted what he thinks the problem is.  I would think that would be the answer to your question as to where to start.

bpalme:
Thanks.  I guess I was hoping he would throw me a little more meaty bone.  I didn't see any good way for a quick fix from class.search.  So I'm assuming I have to get into class.orm?  There is a lot there to dive into...

bpalme:
[...]

PR #4124, opened 2017-12-14
issue: List & Choice Searching

JediKev:
This addresses issues #3703, #3493, #2625, and others where searching for
List or Choice values in a Custom Field will yield no results. This was
due to us using `FIND_IN_SET()` and `IN()` to search, which will not work
with the way the values are stored in the database. This updates the
search methods to use `REGEXP()` to search for the IDs of the List and
Choice values which will give the correct results whether it's multiple
choice or not.

### DIFF
--- a/include/class.dynamic_forms.php
+++ b/include/class.dynamic_forms.php
@@ -1809,11 +1809,12 @@ class SelectionField extends FormField {
 
     function getSearchQ($method, $value, $name=false) {
         $name = $name ?: $this->get('name');
+        $val = '"?'.implode('("|,|$)|"?', array_keys($value)).'("|,|$)';
         switch ($method) {
         case '!includes':
-            return Q::not(array("{$name}__intersect" => array_keys($value)));
+            return Q::not(array("{$name}__regex" => $val));
         case 'includes':
-            return new Q(array("{$name}__intersect" => array_keys($value)));
+            return new Q(array("{$name}__regex" => $val));
         default:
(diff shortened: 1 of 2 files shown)


JediKev:
Now there is an issue with `__cdata` table results...will update this when I figure it out.

krishbang88:
@JediKev  Actually the search is not working when we select multiple custom fields as search criteria.

For example 
My search criteria was
one custom filed (free text field) with "equals" option
second custom field which is a list item with "includes" option. 
The search is giving empty results.

But if I select one custom field at a time as search criteria, search works even if the custom field is a list.

JediKev:
@krishbang88 thanks, will look into this when I have time. Cheers.

JediKev:
@krishbang88 I just got a little free time so I tried to replicate your issue and could not do so. Below is an example of it working for me:

**Advanced Search:**
advanced search

**1 Result:**
1 result

**Both Fields Set:**
both fields set

I would suggest that you check your tickets and make sure they have BOTH fields set. If a ticket has just one of the two fields set then it will not show up in the results.

Cheers.

nicoletta-maia:
@JediKev thanks

Just a note, doing a search of type "has a value" or "does not have a value" we get this warning:
> PHP Warning:  array_keys() expects parameter 1 to be array, null given in include/class.dynamic_forms.php on line 1812

JediKev:
@nicoletta-maia 

Addressed in `v1.11.0rc-1`. Thanks though.

Cheers.

PR #4127, opened 2017-12-27
Issue #3703

jesse-thomas:
This is what I put together when debugging the issue I had, #3703. I did some testing and it appeared to work. I'm posting it in the hopes that it can be tested farther or at least help some how with the issue.


sperows:
Just a quick question: How is this different from this? https://github.com/osTicket/osTicket/pull/4124

jesse-thomas:
@sperows I'm not an osTicket expert, I just did some debugging one day and got my setup to work. You would really have to take a good look at the code and do some testing to see. If you do please let me know if either worked. Thanks!

Issue #3991, opened 2017-10-09
Ticket Lock and Timezone problems for Australia/Brisbane after daylight saving change in AEST


perryk:
### Prerequisites

* [ Y] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [ N] Do you have any errors in the PHP error log, or javascript console?
* [ Y] Did you check the [osTicket forums](http://osticket.com/forum/categories)?
* [ Y] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

### Versions

![screenshot from 2017-10-08 23-59-09](https://user-images.githubusercontent.com/806269/31323141-98d7a5fe-ace6-11e7-8aa3-ad27f910b775.png)

### Description

After daylight savings started in the AEST timezone, osTicket installations set to Brisbane time are unable to reply or edit tickets as the ticket lock renew drop down windows immediately appears and if renew is selected it just waits forever. If the tick lock is returned off, the system still displays all time-dates as being 1 hour behind the actual time in Brisbane. The system report incorrect reports the database timezone as being interpreted as Australia/Melbourne.

I've been around and around a few times on this, initially, I thought it started before the change to daylight saving for all the other state capitals in the AEST timezone. Brisbane is different in that it doesn't observe daylight savings whilst Melbourne and Sydney do.

I found a couple of PR's which both looked to fix the lock issue, however they didn't fix the wrong time issue. In particular [this](https://github.com/osTicket/osTicket/pull/3734) PR is interesting but it only fixes cases in which there are multiple timezones which might be returned, not the case I have here, although it looks similar.

In the end, I started working with a stock standard v1.10.1 branch of the code and found there were no ticket lock problems if the system date was 30 Sept, and there _were_ problems if the system date is 1 Oct or later. This matches with the start of daylight saving AEST and confirms I don't need the other PRs. This also confirmed that the system report was always reporting the timezone being interpreted as Australia/Melbourne.

Further complicating this I worked out was that I was using a database backup from Sept for my testing and there was session data in the ost_session database which included a cached variable for the timezone.

From what I gather there is initially a call to "SELECT @@system_time_zone.global;" to obtain the database timezone before the result is saved to the session data.

The select statement above returns "AEST" for me, however, this is then passed through the format class where the PHP function "timezone_name_from_abbr" is incorrectly returning "Australia/Melbourne" instead of "Australia/Brisbane" and saving this to the session data.

[...]

PR #4186, opened 2018-03-02
oops: Fix System Timezone Variable

JediKev:
This addresses an issue where people are getting the incorrect time on tickets even though they have all the timezones set correctly. The current MySQL variable the system uses to determine the timezone is `system_time_zone`. This variable is set by the server and sometimes cannot be updated properly (or cannot be updated at all). 

This changes the variable from `system_time_zone` to `time_zone` so if in the event the system returns the incorrect time and they can't update the server timezone properly, they can just change the `@@global.time_zone` and `@@session.time_zone` MySQL variables and should get the correct time.

### DIFF
--- a/include/mysqli.php
+++ b/include/mysqli.php
@@ -115,7 +115,11 @@ function db_version() {
 }
 
 function db_timezone() {
-    return db_get_variable('system_time_zone', 'global');
+    $timezone = db_get_variable('time_zone', 'global');
+    if ($timezone == 'SYSTEM')
+        $timezone = db_get_variable('system_time_zone', 'global');
+
+    return $timezone;
 }
 
 function db_get_variable($variable, $type='session') {


perryk:
Thanks @JediKev for putting this into a PR.

It would be interesting to hear if this fixes things for others who are having the trouble previously too. Just to confirm it works for more than just my case of when in "Australia/Brisbane" however I'm pretty confident it will do so and should sort of other similar cases too.

Cheers.

protich:
@JediKev - I'm not sure I agree with the switch!

* Won't it break dates for those with `time_zone` set different from SYSTEM? 
* If so, perhaps we should fall back to `system_time_zone` if `time_zone` is not set?

Please note that  do heavy-lifting in `DbTimezone::determine()` - see `class.timezone.php`

I don't have the answer - just want to start the discussion on the proposed fix. For reference see [mysql manual](https://dev.mysql.com/doc/refman/5.7/en/time-zone-support.html) on timezone setting and the difference between the two variables.

perryk:
I'm happy to test this a little more when I have a chance considering it is my timezone that has a problem with the current option. I'm not sure when daylight savings finishes for the other states here however when that happens the problem will go away for me for a while.

Perhaps it is worth checking what some other php based applications use on github ?

Also, I remember that the very interesting looking code is in `DbTimezone::determine()` - see `class.timezone.php`  as described above, however from memory, I could never find a way to have this code called unless the database failed to report any type of timezone. 

It has been a few months now, my previous comments are in this issue: https://github.com/osTicket/osTicket/issues/3991

I'd be interested to hear from anyone else in an odd timezone if this fixes their other issues too. I think there are some cases which are worked around by setting the timezone in the database, then there is my case which was just a little more broken.

Thanks @JediKev and @protich for looking into this.

haijunsu (commented):
The 'global.time_zone' value can be "SYSTEM". It should handle the value "SYSTEM" as well. If mysql and osTicket are running on different servers, the SYSTEM may be different. see post: http://osticket.com/forum/discussion/comment/122427#Comment_122427

JediKev:
@haijunsu 

Right, that's why this isn't merged...not fully complete yet.

Cheers.

PR #4187, opened 2018-03-04
Fix #4185 - Extend staff username maxlength to 64

### LINKED ISSUE #4185: Username limited to 32 characters
Hi,
Why is the username limited to 32 characters?

Best regards

plenaerts:
This change increases the ost_staff table username field length to 64.
The user forms do not warn against too long usernames.


JediKev:
@plenaerts This will fix the length for _new_ installations but this will not fix the length for _current_ installations or _upgrades_. You might want to look into adding a fix for current installs and upgrades. Cheers.

plenaerts:
@JediKev I think these commits cover upgrades.

Issue #3468, opened 2016-11-04
stuck on "loading...." when closing a TASK within a ticket. 


jakobii:
![selection_006](https://cloud.githubusercontent.com/assets/19825478/20024789/f2201318-a2a5-11e6-964a-33263049ef42.png)

This is what it gets stuck on when I close a task within a ticket. If I refresh the page the ticket reloads with the task closed correctly. I can also close the task just fine when not within the ticket page. 

love being able to manage tasks within a ticket, awesome! thanks!

v1.10 11de92f
ubuntu 16.04.1
Apache/2.4.18
PHP 7.0.8-0ubuntu0.16.04.3

![selection_008](https://cloud.githubusercontent.com/assets/19825478/20024892/acee12e4-a2a6-11e6-8e04-212f7200f3c8.png)

#3058 - not sure if this could be related to my problem...  I doubt it though.
cgi.fix_pathinfo	 "1" is recommended if AJAX is not working
date.timezone	 Setting default timezone is highly recommended

jakobii:
I have rebuilt a new osticket server with php5.6. I am still having the same problem.

ntozier:
Your version of osTicket does not appear to be the one from osticket.com/download.
Which is: v1.10 (901e5ea).  I would recommend that you download the version from the official site and install that... not whatever this version you have installed now.

jakobii:
thank you for the suggestion. I download the osticket from osticket.com v1.10 (901e5ea). the problem still exists.

krishbang88:
@jakobii : I am experiencing similar issue. did you solve it? 
can somebody post a resolution for this issue plz.

Chefkeks:
+1
Same issue here - osTicket 1.10 (901e5ea) is stuck on "loading...." when ~~closing~~ creating a TASK within a ticket.
Apache 2.4.16 with PHP 5.6.9

warpedreality:
Also experiencing this issue.  Happens when posting updates and/or closing tasks.  Page has to be refreshed in order to resume normal functionality.  

Server is Ubuntu 16.04

osTicket Version 	v1.10 (901e5ea) — Up to date
Web Server Software 	Apache/2.4.18 (Ubuntu)
MySQL Version 	5.7.16
PHP Version 	5.6.29-1

Firefox 50.1.0

tykisson:
I am getting stuck on the loading screen when replying to a task with in a ticket. If I go to the task from the Task Tab and post a reply it works just fine.

[...]

PR #4200, opened 2018-03-16
Fix: stuck on "loading...." when closing a TASK within a ticket

nicoletta-maia:
This addresses issue #3468 where page is stuck on "loading...." when posting a reply or note at a TASK within a ticket.
_#overlay_ and _#loading_ were shown after the form submission and were no longer hidden.


nicoletta-maia:
This issue has been fixed in #4660

PR #4229, opened 2018-04-10
i18n: FAQ Local Questions & Answers

JediKev:
This addresses two issues, with the first being the Featured FAQs on the
client portal are not translated when selecting another language. The
other issue is when you click the FAQ Category to see all FAQs the FAQ
names are not translated. This adds the correct `getLocal*` functions to
the FAQ templates to get the correctly translated Questions & Answers.
This also adds a new function called `getLocalTeaser` which will return
the truncated, translated FAQ answer.

### DIFF
--- a/include/class.faq.php
+++ b/include/class.faq.php
@@ -79,6 +79,9 @@ class FAQ extends VerySimpleModel {
     function getTeaser() {
         return Format::truncate(Format::striptags($this->answer), 150);
     }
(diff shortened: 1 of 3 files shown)


IceKaching:
This works great with 1.15.2 --- please merge it.

There is still one small issue though and that is when Searching the KB, it does not find works from the other languages - only English. Has anyone else come across this?

PR #4230, opened 2018-04-10
Update class.avatar.php

HPringles:


HPringles:
Mistakenly submitted to the wrong repo

PR #4253, opened 2018-04-25
issue: Files - deleteOrphans()

JediKev:
This addresses the issue where files were being deleted before being sent
out in Agent responses. This was due to a bug in the query that gets the
orphaned files. This query was getting files created within the last 24
hours not after the last 24 hours. The query also had another bug that
would use the time from PHP instead of MySQL which could cause issues.
This updates the query as per @greezybacon's suggestions to delete
orphaned files that were created more than 24 hours ago.

### DIFF
--- a/include/class.file.php
+++ b/include/class.file.php
@@ -635,7 +635,7 @@ class AttachmentFile extends VerySimpleModel {
             ->filter(array(
                 'attachments__object_id__isnull' => true,
                 'ft' => 'T',
-                'created__gt' => new DateTime('now -1 day'),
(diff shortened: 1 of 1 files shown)


toyowheelin:
@JediKev Thanks for the pointer, I will see if I can apply the patch to my install.

toyowheelin:
So I thought this patch fixed the issue I was seeing where people would have their attachments go missing, but now today out of the blue with no changes I just got a few more reports of this same issue. Do you know if this fix works with the file backend storage plugin or is it only for storing data in SQL?

PR #4259, opened 2018-04-30
PHP 7.1 compatibility: session handler return var

clonemeagain:
Read handler needs to return a string in 7.1. So this pull adds explicit `(string)` casts to the success return on the custom handlers.

Compare with return value section here: http://php.net/manual/en/sessionhandlerinterface.read.php

Conveniently: Fixes login session errors when running on PHP 7.1.15-0ubuntu0.17.10.1

### DIFF
--- a/include/class.ostsession.php
+++ b/include/class.ostsession.php
@@ -190,7 +190,7 @@ extends SessionBackend {
         catch (OrmException $e) {
             return false;
         }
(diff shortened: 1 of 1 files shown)


pieleric:
We are using the v1.11.x branch, with PHP 7.1 and this fix is needed to be able to login.

PR #4260, opened 2018-05-01
Label translate problem in dynamic form fixed

mkysoft:


JediKev:
What exactly were you trying to fix?

Unfortunately, this pull is incorrect as you're putting the localization function around a function that gets the translation for the field's label.

Cheers.

mkysoft:
getLocal() function not return correct translate. I fixed it adding localization.I tested it, it look like working.
You can test it on create user form.

mkysoft:
I am sorry, I make detail analyze about this form. It is look like custom form, and need manually update.

PR #4267, opened 2018-05-03
fixed csrf-bug in PHP 7.2

cyberer:
the SessionBackend::read should always return a string, the SessionBackend class should implement class SessionHandlerInterface


JediKev:
@cyberer 

We already have two pull requests for this issue. One is already merged into the new version that's about to be released (v1.11rc-1). Thanks anyways.

Cheers.

Issue #4301, opened 2018-05-29
MimePart issue with counting NULL member subparts


shanept:
### Prerequisites

* [x] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [x] Do you have any errors in the PHP error log, or javascript console?
* [x] Did you check the [osTicket forums](http://osticket.com/forum/categories)?
* [x] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/osTicket/osTicket/wiki/Github-Issues-Guidelines)

### Description
`[php7:warn] [pid 9542] [client 10.1.5.5:4883] PHP Warning:  count(): Parameter must be an array or an object that implements Countable in /var/www/html/include/pear/Mail/mimePart.php on line 314`

Upon attempting to count the number of elements in the un-initialized member varible 'subparts', PHP logs a warning.

The above warning is from a fresh install.

### Steps to Reproduce

1. [First Step]
2. [Second Step]
3. [and so on...]

**Expected behavior:**
No error should be logged - variable should be initialized.

**Actual behavior:**
An error is logged.

### Versions

Admin panel -> Dashboard -> Information which also additionally gives you information about you server. - Unable to access.

osTicket: 1.10.1
PHP: 7.2
apache: 2.4.6
CentOS: 7.5.1804

JediKev:
You are using a PHP version which is not fully supported by osTicket just yet. We aim to fully support PHP 7.x+ in version 2.0. In the meantime please downgrade to PHP 5.6 or use the fix you literally just posted while I was typing this lol

Cheers.

shanept:
Oh, I just love to live dangerously!
I'm having a few other syntactical issues with osticket at the moment, I'll be popping a few more issues/fixes up shortly, but I'll keep the supported versions in mind, thank you.

PR #4302, opened 2018-05-29
Fix for warning on Mailer (#4301)

shanept:
Fixes issue opened at #4301


shanept:
As per comment by @JediKev this is an issue with a PEAR module, not osTicket itself.

PR #4335, opened 2018-07-12
issue: Ticket Lock On Disable

JediKev:
This addresses an issue where disabling Ticket Locks but setting a time
for the Lock Timeout will always throw a "lock required" error. This adds
a check to see if locks are enabled AND there is a time set. If locks are
disabled it will not throw the error.

### DIFF
--- a/scp/tickets.php
+++ b/scp/tickets.php
@@ -90,7 +90,7 @@ if($_POST && !$errors):
                 if(!$vars['response'])
                     $errors['response']=__('Response required');
 
-                if ($cfg->getLockTime()) {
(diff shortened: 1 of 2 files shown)


scp/tickets.php
```diff
@@ -90,7 +90,7 @@
                 if(!$vars['response'])
                     $errors['response']=__('Response required');
 
-                if ($cfg->getLockTime()) {
+                if (($cfg->getTicketLockMode() != Lock::MODE_DISABLED) && $cfg->getLockTime()) {
```
protich:
```php
$cfg->isTicketLockEnaned(); 
```

Issue #4331, opened 2018-07-09
Cross site scripting,Blind sql injection attack found in osticket


asha42:
[mission.pdf](https://github.com/osTicket/osTicket/files/2174855/mission.pdf)

JediKev:
@asha42 

Thank you for the report. Unfortunately almost all of the things it's complaining about are things that Server Admins should do or things that should be ignored (although there are a few issues we can look into). I have attached a document containing my response to each warning. Below is a summary of the attached document.


### Attachment
[xss_response.txt](https://github.com/osTicket/osTicket/files/2176752/xss_response.txt)



### Attachment Summary

**Things We Need To Look Further Into:**
- [High] Parameter System Call Code Injection _**(Fabric.js; not us but we will see how they addressed this)**_
- [Low] Unsafe third-party link (target="_blank") _**(not really necessary but will look into it just in case)**_
- [Medium] Cross-Site Request Forgery _**(look into validating the referrer)**_
- [Low] Unsafe third-party link (target="_blank") _**(add 'rel="noopener noreferrer"' to the links with 'target="_blank"')**_

**Things Server Admins Should Do:**
- [High] Unencrypted Login Request _**(Configure an SSL certificate for your server and force all traffic/requests to use HTTPS.)**_
- [Medium] Authentication Bypass Using HTTP Verb Tampering _**(Configure your server to allow only required HTTP methods.)**_
- [Low] Missing "X-XSS-Protection" header _**(Configure your server to use the "X-XSS-Protection" header.)**_
- [Low] Missing Cross-Frame Scripting Defence _**(Configure your server to use the "X-Frame-Options" header.)**_
- [Low] Missing "Content-Security-Policy" header _**(Configure your server to use the "Content-Security-Policy" header.)**_
- [Low] Missing "X-Content-Type-Options" header _**(Configure your server to use the "X-Content-Type-Options" header.)**_

**Things To Ignore:**
- [Low] Permanent Cookie Contains Sensitive Session Information
- [Information] Email Address Pattern Found _**(osTicket is an email-based system so of course you will find email addresses in some responses.)**_
- [Low] Autocomplete HTML Attribute Not Disabled for Password Field _**(Some people like password autofill and some people don’t.)**_
- [Information] Internal IP Disclosure Pattern Found _**(Don’t use an IP for your helpdesk url.)**_
[...]

PR #4337, opened 2018-07-13
Address Security Issues On #4331

JediKev:
It may be possible to steal or manipulate customer session and cookies, which might be used to impersonate a legitimate user, allowing the hacker to view or alter user records, and to perform transactions as that user. Sanitation of hazardous characters was not performed correctly on user
input.

osTicket did not properly sanitize array values in `Format::htmlchars()`. Some values in the Admin Interface were not properly sanitized and returned to the response.

### DIFF
--- a/include/class.config.php
+++ b/include/class.config.php
@@ -1110,6 +1110,8 @@ class OsticketConfig extends Config {
         $f['default_timezone']=array('type'=>'string',   'required'=>1, 'error'=>__('Default Timezone is required'));
         $f['system_language']=array('type'=>'string',   'required'=>1, 'error'=>__('A primary system language is required'));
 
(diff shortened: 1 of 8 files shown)


alvahdati:
@JediKev hi. this commit is in release 1.10.4 right? can we say that 1.10.4 is not affected by CVE-2017-15580 available at: https://becomepentester.blogspot.com/2017/10/osTicket-File-Upload-Restrictions-Bypassed-CVE-2017-15580.html? it seems 1.10.4 does the required checking while uploading html files but i just wanna be sure whether my decision is correct. thanks.

JediKev:
@alvahdati

Not this commit; that issue was addressed with:
https://github.com/osTicket/osTicket/pull/4337

Cheers.

PR #4363, opened 2018-07-24
issue: CLI Deploy Missing Bootstrap Fix

JediKev:
This addresses an issue introduced with #4332 where the deploy CLI breaks upon deploying to an existing location. This checks to see if there is a bootstrap file in the destination first and if so it uses that one if not uses the source.

### DIFF
--- a/include/cli/modules/unpack.php
+++ b/include/cli/modules/unpack.php
@@ -210,7 +210,11 @@ class Unpacker extends Module {
         ), $pipes);
 
         fwrite($pipes[0], "<?php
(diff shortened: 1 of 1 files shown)


bmorg:
Took me quite a bit of debugging to find this issue. Is this not breaking the upgrade process for everyone? It might be worth a 1.10.5 bugfix release?

bmorg:
@ntozier I see that it has been merged. I am saying that the issue might be grave enough to justify a bugfix release.

PR #4395, opened 2018-08-01
issue: Saved Searches Flags

JediKev:
This addresses an issue where class Queue sets SavedSearches flags to `FLAG_QUEUE` causing the SavedSearches to become Primary Queues.

### DIFF
--- a/include/class.queue.php
+++ b/include/class.queue.php
@@ -1293,7 +1293,8 @@ class CustomQueue extends VerySimpleModel {
 
         $queue = new static($vars);
         $queue->created = SqlFunction::NOW();
-        $queue->setFlag(self::FLAG_QUEUE);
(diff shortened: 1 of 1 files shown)


include/class.queue.php
```diff
@@ -1293,7 +1293,8 @@ static function create($vars=false) {
 
         $queue = new static($vars);
         $queue->created = SqlFunction::NOW();
-        $queue->setFlag(self::FLAG_QUEUE);
+        if (isset($vars['flags']) && $vars['flags'] != 0)
```
protich: I believe this is incorrect  - `FLAG_QUEUE` shouldn't be set when `$vars['flags']` is set.

JediKev:
@protich 

Updated. Tested and works!

Cheers.

PR #4412, opened 2018-08-06
Agent Default Queue:

aydreeihn:
This commit addresses an issue where the Agent could set a Default Ticket Queue in their profile, but the System Default queue would be shown instead.

### DIFF
--- a/scp/tickets.php
+++ b/scp/tickets.php
@@ -93,7 +93,7 @@ if (!$ticket) {
 
     $queue_key = sprintf('::Q:%s', ObjectModel::OBJECT_TYPE_TICKET);
     $queue_id = $queue_id ?: @$_GET['queue'] ?: $_SESSION[$queue_key]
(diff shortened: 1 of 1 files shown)


scp/tickets.php
```diff
@@ -92,8 +92,8 @@
     }
 
     $queue_key = sprintf('::Q:%s', ObjectModel::OBJECT_TYPE_TICKET);
-    $queue_id = $queue_id ?: @$_GET['queue'] ?: $_SESSION[$queue_key]
-        ?: $cfg->getDefaultTicketQueueId();
+    $queue_id = $queue_id ?: @$_GET['queue'] ?: $thisstaff->getDefaultTicketQueueId()
```
protich: I think `$_SESSION[$queue_key]` should take precedence over staff's settings - otherwise we will loose stickiness of queues.

PR #4414, opened 2018-08-06
Referral Assignment Issue:

aydreeihn:
This commit fixes an issue where referrals were being considered successfully saved even if that referral already existed, resulting in multiple thread events for the same referral.

Instead, if we find that the thread is already referred to the referee, refer should return false so that an error message is displayed stating that we were unable to refer the ticket.

### DIFF
--- a/include/class.thread.php
+++ b/include/class.thread.php
@@ -314,7 +314,7 @@ implements Searchable {
     function refer($to) {
 
         if ($this->isReferred($to, true))
-            return true;
+            return false;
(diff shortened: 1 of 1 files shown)


protich:
Make sure this doesn't mess up auto-referral

aydreeihn:
I tested auto-referral by cc'ing a department email where there were no referrals as well as cc'ing the department email where a referral to that department already existed, and both worked fine.

PR #4416, opened 2018-08-06
issue: Featured FAQs On Disable

JediKev:
This addresses an issue where the client portal shows Featured FAQs on the sidebar even though the Knowledgebase is Disabled.

### DIFF
--- a/include/client/templates/sidebar.tmpl.php
+++ b/include/client/templates/sidebar.tmpl.php
@@ -19,8 +19,9 @@ $BUTTONS = isset($BUTTONS) ? $BUTTONS : true;
         </div>
 <?php } ?>
         <div class="content"><?php
-    $faqs = FAQ::getFeatured()->select_related('category')->limit(5);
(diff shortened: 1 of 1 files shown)


include/client/templates/sidebar.tmpl.php
```diff
@@ -20,7 +20,7 @@
 <?php } ?>
         <div class="content"><?php
     $faqs = FAQ::getFeatured()->select_related('category')->limit(5);
-    if ($faqs->all()) { ?>
+    if ($cfg->isKnowledgebaseEnabled() && $faqs->all()) { ?>
```
protich: I think faqs query should be within if statement so we don't want a trip to db on each page when when knowledge base is disabled. 

```php

if ($cfg->isKnowledgebaseEnabled() 
  && ($faqs = FAQ::getFeatured()->select_related('category')->limit(5))
  && $faqs->all()) { ?>

```

JediKev:
@protich 

Updated.

Cheers.

PR #4420, opened 2018-08-07
Implement 'Select Active Collaborators':

aydreeihn:
Enable/Disable Collaborators based on which Collaborators are in the 'Select Active Collaborators' box


include/class.ticket.php
```diff
@@ -2916,6 +2916,23 @@ function postReply($vars, &$errors, $alert=true, $claim=true) {
         if (isset($vars['ccs']) && count($vars['ccs']))
             $this->addCollaborators($vars['ccs'], array(), $errors);
 
+        if ($collabs = $this->getRecipients('collabs', array(), false)) {
```
protich: Why could you call 'getCollaborators' directly and avoid mailing list / contact mess?

PR #4425, opened 2018-08-08
issue: Old Attachment Names

JediKev:
This addresses an issue where uploading a file you’ve uploaded before but renamed is not saving the new name in the Attachment table.

### DIFF
--- a/include/class.attachment.php
+++ b/include/class.attachment.php
@@ -108,15 +108,23 @@ extends InstrumentedList {
      */
     function keepOnlyFileIds($ids, $inline=false, $lang=false) {
         if (!$ids) $ids = array();
-        $new = array_fill_keys($ids, 1);
+        $new = array_flip($ids);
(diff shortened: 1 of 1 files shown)


include/class.attachment.php
```diff
@@ -107,16 +107,24 @@ function getType() { return $this->key['type']; }
...
-        $new = array_fill_keys($ids, 1);
+        $new = array_flip($ids);
         foreach ($this as $A) {
             if (!isset($new[$A->file_id]) && $A->lang == $lang && $A->inline == $inline)
                 // Not in the $ids list, delete
                 $this->remove($A);
             unset($new[$A->file_id]);
         }
+        // Format $new for upload() with new name
+        foreach ($new as $id=>$name) {
```
protich: Move `$attachments = array();` from line 110 down here... unit close to block of code that uses the variable to avoid overwrite errors as code get added in the future.

PR #4426, opened 2018-08-08
issue: Wrong Attachment Names

JediKev:
This addresses an issue where FileUploadFields are displaying the old names of existing files with new names.

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -2951,10 +2951,10 @@ class FileUploadField extends FormField {
 
     function display($value) {
         $links = array();
-        foreach ($this->getFiles() as $f) {
+        foreach ($this->getAttachments() as $a) {
             $links[] = sprintf('<a class="no-pjax" href="%s">%s</a>',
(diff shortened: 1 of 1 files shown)


include/class.forms.php
```diff
@@ -3896,28 +3896,27 @@ function render($options) {
             function($t) { return strpos($t, '/') !== false; }
         );
         $maxfilesize = ($config['size'] ?: 1048576) / 1048576;
-        $files = $F = array();
+        $files = array();
```
protich: unused variable??

JediKev:
Looking at it now, I need to replace it with `$attachments` and possibly replace `$F[] = $a->file;` with `$attachments[] = $a;`... what do you think?

include/class.forms.php
JediKev: Looking at it now, I need to replace it with `$attachments`. what do you think?

protich: Nothing to think about - delete it - it's not being used, right?

PR #4427, opened 2018-08-08
issue: Multiple File Display

JediKev:
This addresses an issue where FileUploadFields on the ticket header is not displaying all attachments in the field.

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -3426,7 +3426,7 @@ class FileUploadField extends FormField {
     function display($value) {
         $links = array();
         foreach ($this->getAttachments() as $a) {
-            $links[] = sprintf('<a class="no-pjax" href="%s">%s</a>',
+            $links[] = sprintf('<a class="no-pjax" href="%s"><i class="icon-paperclip icon-flip-horizontal"></i> %s</a>',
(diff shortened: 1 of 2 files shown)


protich:
I think testing for a positive (string in this case ) would be a better solution.

```php
if (is_string($v) && strlen($v) > 200)
  truncate... 

```

JediKev:
@protich

When you do that, if you only have one file it’ll truncate the name and put an icon in a weird place. Basically, if it’s a file field then we should display the entire file name(s) or at least a list of longer truncated names with no icon. I’ll play with it a little more.

protich:
My concern is `$a->getField() instanceof FileUploadField` - you're basically testing the last answer every time from foreach loop around line 600 - so if `FileUploadFiled` is in the middle of then the fix is mute!

I think the display format (e.g truncating or not) should go on display call - around line 601.

protich:
Closed the PR by mistake -

JediKev:
@protich 

Updated.

Cheers.

protich:
@JediKev - Please rebase this PR on top of `develop-next`

JediKev:
@protich Rebased.

PR #4428, opened 2018-08-08
issue: FAQ & Canned Attachments Dropping

JediKev:
This addresses an issue where FAQ & Canned Response attachments are dropping when another agent edits the FAQ or Canned Response.

### DIFF
--- a/scp/canned.php
+++ b/scp/canned.php
@@ -37,6 +37,13 @@ $canned_form = new SimpleForm(array(
    )),
 ));
 
+// Set fields' attachments so exsting files stay put
+if ($canned
+    && $canned->attachments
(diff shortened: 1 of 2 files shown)


scp/canned.php
```diff
@@ -37,6 +37,12 @@
    )),
 ));
 
+// Set fields' attachments so exsting files stay put
```
protich: Canned response might not necessarily have attachments.
Canned response might not necessarily have attachments. 

```php
if ($canned 
    && $canned->attachments
    && ($attachments = $canned_form->getField('attachments'))) {
     $attachments->setAttachments($canned->attachments);
 }
```

PR #4438, opened 2018-08-13
release: Options Enabled By Default

JediKev:
This enables the release option checkboxes by default.


protich:
You need to think about what happens on error - I would imagine you want to maintain what the user had unchecked or checked.

greezybacon:
What "options" are you enabling (by default)? What is the "release option"

protich:
@greezybacon - Ability to release assignment - which can be an agent or/and a team. Options here really means checkboxes for assignees.

protich:
The implementation for sticky checkboxes is a little dirty to me - will discuss it in person.

JediKev:
@protich Updated.

PR #4439, opened 2018-08-13
perms: Alphabetize Role/Staff Permissions

JediKev:
This alphabetizes all Role and Staff Permissions.

### DIFF
--- a/include/class.role.php
+++ b/include/class.role.php
@@ -320,6 +320,16 @@ class RolePermission {
     }
 
     static function allPermissions() {
+        static $sorted = false;
+
+        if (!$sorted) {
+            // Sort permissions in alphabetical order
+            foreach (static::$_permissions as $k => $v) {
+                asort(static::$_permissions[$k]);
+            }
+            $sorted = true;
+        }
+
         return static::$_permissions;
     }
 


include/staff/role.inc.php
```diff
@@ -80,6 +80,7 @@
                 if ($v['primary'])
                     continue;
                     $buckets[$g][$k] = $v;
+                asort($buckets[$g]);
```
greezybacon: I don't think you mean to put this here? It's inside the `foreach` for `$k`, so you'll be sorting it for each key added to `$buckets[$g]`

include/staff/staff.inc.php
```diff
@@ -337,6 +337,7 @@ class="error">*</span></div>
             if (!isset($permissions[$g]))
                 $permissions[$g] = array();
             $permissions[$g][$k] = $P;
+            asort($permissions[$g]);
```
greezybacon: I don't think you mean to put this here? It's inside the `foreach` for `$k`, so you'll be sorting it for each key added to `$permissions[$g]`

include/class.role.php
```diff
@@ -320,6 +320,10 @@ function merge($perms) {
     }
 
     static function allPermissions() {
+        // Sort permissions in alphabetical order
+        foreach (static::$_permissions as $k => $v) {
+            asort(static::$_permissions[$k]);
+        }
```
greezybacon: This is probably fine, but it could result in unnecessary sorting, since its sorted on every invocation. I would suggest
```php
static function allPermissions() {
    static $sorted = false;

    if (!$sorted) {
        // Sort permissions in alphabetical order
        foreach (static::$_permissions as $k => $v) {
            asort(static::$_permissions[$k]);
        }
        $sorted = true;
    }

    return static::$_permissions;
}
```

PR #4441, opened 2018-08-14
Address Performance Issues For Large Datasets

greezybacon:
For very large ticket databases (like over 1 million tickets), there seems to be a mix of MySQL having trouble identifying the most optimized query plan and unnecessary results sorting and such. This PR is an attempt to bring the page load timing of large systems closer to O(1) where pages load quickly regardless of how many million tickets your awesome staff has handled over the years--without changing how the system works or by adding unnecessary complexity to the codebase.


JediKev:
👌🏻

include/class.search.php
```diff
@@ -874,7 +882,24 @@ static function ticketsCount($agent, $criteria=array(),
...
+        return $counts;
+    }
+
+    static function clearCounts($agent=null) {
+        global $thisstaff;
+
+        $agent = $agent ?: $thisstaff;
+        if ($agent instanceof Staff && function_exists('apcu_store')) {
+            $key = "counts.queues.{$agent->getId()}.".SECRET_SALT;
+            apcu_delete($key);
```
greezybacon: Iterate over APCu and drop stats for all users

include/class.queue.php
```diff
@@ -1240,8 +1240,11 @@ function save($refetch=false) {
         $nopath = !isset($this->path);
         $path_changed = isset($this->dirty['parent_id']);
 
-        if ($this->dirty)
+        if ($this->dirty) {
             $this->updated = SqlFunction::NOW();
+            // Refetch the queue counts
+            SavedQueue::clearCounts();
+        }
```
greezybacon: Add this for tickets too

include/class.ticket.php
```diff
@@ -3175,6 +3175,9 @@ function deleteDrafts() {
     function save($refetch=false) {
         if ($this->dirty) {
             $this->updated = SqlFunction::NOW();
+            if (isset($this->dirty['status']))
```
greezybacon: `status_id`

protich:
@greezybacon - shouldn't we consider stashing the counts in the agent's session as an alternative when `APCu` is not available?

greezybacon:
Yes, we could. My concern is that the session data could overflow the
session data field in the database and force the agent to be logged out.
Plus, it still requires to do the operation for each agent logged in.

If placed in the session, perhaps we could limit it to only the queues to
which an agent has access? That would help keep the session data from over
flowing the database field?

I assume you mean both the queue exact and rough counts? Or just one of
them?

Cheers,

protich:
I was thinking only queues they have access to - just the numbers but on demand only; When queue counts is triggered or via rough counts we can simply stash already available data.

greezybacon:
Honestly, @protich. I'm wondering if we should just abandon the SQL_CALC_ROWS_FOUND. I think it only helps in small recordsets--where it isn't needed? Then the count and pagination on the queue page footer would be accurate. It it were cached, that would be a better optimization than the SQL_CALC_ROWS_FOUND?

protich:
@greezybacon - made a pull request to your branch to stash counts per user. Also added an entry to auto-cron to keep the cache fresh in the background.

greezybacon:
@protich what's missing from this request? I think the pagination needs to be revised? That and what else?

PR #4450, opened 2018-08-15
sessions: Clear On Password Set/Reset

JediKev:
This adds a Signal to clean Agent and User sessions upon setting/resetting their password. If an Agent/User resets their own password and has multiple sessions open it will log them out of every session except the one they’re on.

### DIFF
--- a/include/class.auth.php
+++ b/include/class.auth.php
@@ -1313,7 +1313,35 @@ abstract class PasswordPolicy {
     static function register($policy) {
         static::$registry[] = $policy;
     }
+
+    static function cleanSessions($model, $user=null) {
+        $criteria = array();
+
+        switch (true) {
+            case ($model instanceof Staff):
+                $criteria['user_id'] = $model->getId();
+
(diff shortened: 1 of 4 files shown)


protich:
I think you should user signals, so you can consolidate the code in `include/class.auth.php` as part of password policy. This will help eliminate the need to repeat same code blocks.

JediKev:
@protich 

Updated.

Cheers.

include/class.auth.php
```diff
@@ -1313,7 +1313,41 @@ static function allActivePolicies() {
...
+                $criteria['session_data__regex'] = $regexp;
+
+                if ($user)
+                    array_push($criteria,
+                        Q::not(array('session_id' => $user->session->session_id)));
+                break;
+            default:
+                return false;
+        }
+
```
protich: With the switch statement default returning `false` - you're guaranteed `$criteria`.  I would cut out the rest of the code by simply returning affected rows.

```php
return SessionData::objects()->filter($criteria)->delete();
```

PR #4451, opened 2018-08-15
cron: Clean Expired Passwd Resets

JediKev:
This adds functionality to clean expired password reset tokens on cron runs.

### DIFF
--- a/include/class.config.php
+++ b/include/class.config.php
@@ -176,6 +176,20 @@ extends VerySimpleModel {
             $this->updated = SqlFunction::NOW();
         return parent::save($this->dirty || $refetch);
     }
+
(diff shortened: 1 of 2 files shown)


include/class.config.php
```diff
@@ -176,6 +176,20 @@ function save($refetch=false) {
             $this->updated = SqlFunction::NOW();
         return parent::save($this->dirty || $refetch);
     }
+
+    // Clean password reset tokens that have expired
+    static function cleanPwResets() {
+        global $cfg;
+
+        if (!$cfg && !($period = $cfg->getPwResetWindow())) // In seconds
```
protich: should be or (`||`)

PR #4452, opened 2018-08-16
Relative Time Fixes:

aydreeihn:
This commit removes the logic that was used for time within 2 days which would give a result of either today or tomorrow when something could actually be in 2 days instead.

It also rounds relative time that is within 29 days so that the result will be more accurate.
Ex: 1.99 will round to 2 instead of showing 1


aydreeihn:
As a side note, if we wanted to keep the 2 days option, we need to compare the normal $timeDiff variable instead of $absTimeDiff because an absolute value will always be positive.

return $timeDiff >= 0 ? __('yesterday') : __('tomorrow');

PR #4462, opened 2018-08-20
Staff Profile Updates:

aydreeihn:
This commit fixes an issue where the ORM was trying to get invalid Staff fields (which are actually keys  in the config table), therefore throwing an error.


protich:
I'm curious the config key check failed - line 76.

aydreeihn:
I think it's because they're not set initially. An agent has to hit save on their profile before the config table has their configurations, so it throws the error on the initial save only.

PR #4113, opened 2017-12-08
issue: Fix Most Redactor Issues

JediKev:
This addresses most Redactor issues we’ve been having lately (ie. Image
Paste, Bold button doesn’t work, etc.). This updates Redactor from 10.2.2
to 10.2.5 which seems to have resolved all the issues in testing. This
should not affect the original functionality but if you notice any strange
behaviors please let us know.

PR #4464, opened 2018-08-21
redactor: Upgrade and Fix Formatting

JediKev:
This upgrades Redactor to 10.2.3 and addresses minor formatting issues with Chrome and Firefox.


JediKev:
Need some people to test this out for me on the latest pull from `develop-next`. See if you can still do everything redactor in osTicket.

JtheBAB:
I tested with the latest develop-next and this pull request. Looks like everything is working (Chrome). Just one small thing:

When i select "Formatting" --> "Header 1" (or something else) and i write nothing happen. But when i select my text and select "Header 1" it is working. I never used this feature so i don't know if it is normal or not.

JediKev:
@JtheBAB 

First off....

**THANK YOU FOR TESTING**!! 😀

Secondly, it never functioned like that previous so that's good!

Please keep testing and reporting (if you can)!

Cheers.

JtheBAB:
@JediKev 

So far it is working fine. We are using develop-next with this pull request within our production. All our team members can do there daily jobs to make customers happy ;)

JediKev:
Text alignment breaks with new redactor...gotta look into this.

wgalafassijr:
how to apply the patch to fix the issue about copy/paste?

peterson-dane:
I'm testing this on a just-updated 1.10.4, and, while some stuff is fixed, I'm still seeing a lot of the same issues that I'd reported in #3849. **Bold** and **Italic** are not working unless you select a whole paragraph (everything between two newlines) or start on a newline and select "backwards" up onto the previous line. The same is true of the **Change Font Family** formatting.

I've also tried out the fix from #4113 on 1.10.4. This older fix is what I had been using, with pretty good success, on the previous version of osTicket that I was running (1.10.2).

JediKev:
@peterson-dane 

I believe we are moving away from redactor in `2.0`...so please stay tuned!

Cheers.

aloostrom:
Just added this into 1.12.2 with PHP 7.1.9 and IIS7.5. Fixed the paste issue, so I can't complain too much. Will be testing with KB and a couple other cases and add any issues I stumble across. Alignment changes are still a little wonky, but I can make it work. I use this for work, so pasting straight from clipboard(greenshot) saves a ton of time. Looking forward to 2.0!

PR #4466, opened 2018-08-21
issue: mPDF SetAutoFont RTL

JediKev:
This addresses an issue where `SetAutoFont()` was making the Thai PDFs break. This adds the `AUTOFONT_RTL` flag to only autodetect RTL languages and nothing else.


JediKev:
More information on `SetAutoLang()` and it's flags here:
https://mpdf.github.io/reference/mpdf-functions/setautofont.html

PR #4468, opened 2018-08-22
Misc. Fixes

protich:
Collection of fixes for v1.11.x

### DIFF
--- a/include/class.filter.php
+++ b/include/class.filter.php
@@ -369,6 +369,7 @@ class Filter {
         $sql='DELETE FROM '.FILTER_TABLE.' WHERE id='.db_input($id).' LIMIT 1';
         if(db_query($sql) && ($num=db_affected_rows())) {
             db_query('DELETE FROM '.FILTER_RULE_TABLE.' WHERE filter_id='.db_input($id));
+            db_query('DELETE FROM '.FILTER_ACTION_TABLE.' WHERE filter_id='.db_input($id));
         }
 
         return $num;
--- a/include/class.queue.php
+++ b/include/class.queue.php
@@ -253,6 +253,7 @@ class CustomQueue extends VerySimpleModel {
                 'topic_id',
                 'created',
                 'est_duedate',
+                'duedate',
             )
         );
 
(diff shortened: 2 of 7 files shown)


include/class.filter.php
```diff
@@ -369,6 +369,7 @@ function delete() {
         $sql='DELETE FROM '.FILTER_TABLE.' WHERE id='.db_input($id).' LIMIT 1';
         if(db_query($sql) && ($num=db_affected_rows())) {
             db_query('DELETE FROM '.FILTER_RULE_TABLE.' WHERE filter_id='.db_input($id));
+            db_query('DELETE FROM '.FILTER_ACTION_TABLE.' WHERE filter_id='.db_input($id));
```
aydreeihn: This works, however, as a result of some of our latest commits, the filter_id on filter actions isn't being set. I'll look into it

include/class.format.php
```diff
@@ -529,6 +529,10 @@ function __formatDate($timestamp, $format, $fromDb, $dayType, $timeType,
         if (!$timestamp || !($datetime = DateTime::createFromFormat('U', $timestamp)))
             return '';
 
+        // Normalize timezone
+        if ($timezone)
+            $timezone = Format::timezone($timezone);
```
aydreeihn: Where were the timezones invalid before? From what I tested the timezone showed the same before and after this code.

include/class.queue.php
```diff
@@ -2825,7 +2826,8 @@ class DateTimeFilter
     static $desc = /* @trans */ "Date and Time";
 
     function filter($text, $row) {
-        return $text->changeTo(Format::datetime($text->value));
+        return $text ?
```
aydreeihn: Should this be $text->value instead of just $text?

protich: No text can be null, causing a fatal error.

include/class.format.php
protich: Parsing a timezone from a datetime from database you can end up with a timezone like  'GMT' or '+00:00' - which are not valid. 

Now that you mentioned it, it looks line 538 does the same thing - will drop this change.

PR #4472, opened 2018-08-22
issue: Duplicate Page Requests

JediKev:
This addresses an issue where clicking Open, Post Reply, etc. more than once whilst creating/responding to a ticket will generate duplicate requests. This causes duplicate responses to be sent out or duplicate tickets being created. This adds jQuery, on both Agent and User side, to disable the submit buttons when the "Loading" popup appears so you cannot click the button more than once.

### DIFF
--- a/js/osticket.js
+++ b/js/osticket.js
@@ -45,6 +45,9 @@ $(document).ready(function(){
 
     $('form').submit(function() {
         $(window).unbind('beforeunload');
+        // Disable client-side Post Reply/Create Ticket buttons to help
+        // prevent duplicate POST
+        $(':submit', $(this)).attr('disabled', true);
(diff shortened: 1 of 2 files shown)


js/osticket.js
```diff
@@ -45,6 +45,7 @@ $(document).ready(function(){
 
     $('form').submit(function() {
         $(window).unbind('beforeunload');
+        $('input[type="submit"]').attr('disabled', true);
```
protich: Limit the scope of selector to the form.
```js
$('input[type="submit"]', $(this)).attr('disabled', true);
```

JediKev:
@protich Updated.

protich:
I disagree with the fixes... let's talk when you have a second.

Issue #4470, opened 2018-08-22
Disabled list shown in portal


nathaelr:
### Description

Unabled form fields shown in customer portal.
If your unselect some form fields in a help topic, they won't appear when a customer select this category in the portal. However, if they don't complete the form (forget a required field), all the fields (even unselected fields) are showed when page reloads

### Steps to Reproduce
I've reproduced the problem in a fresh install.

1. Create two custom lists
2. Add them to the "ticket details" form as required fields
3. Modify one existing help topic. Unselect one list you just added.
4. Go to portal. Create a ticket of the topic you just modified. 
5. Don't fill a required field.
Result: the unselected list (in help topic) is shown

### Versions
osTicket Version | v1.10.4 (035fd0a)
Web Server Software | Microsoft-IIS/10.0
MySQL Version | 5.7.23
PHP Version | 5.6.31

I think its pretty serious as it can block a customer. If you created many custom fields, they will all be shown and customer won't understand what's happening. It may need a bugfix.

JediKev:
@nathaelr 

Try this and let me know:
https://github.com/osTicket/osTicket/pull/4476

Cheers.

PR #4476, opened 2018-08-23
forms: Disabled By Help Topic Users

JediKev:
This addresses issue #4470 where Users creating tickets via Client Portal and failing to fill out a required field makes disabled fields by Help Topic appear on the page. When the disabled fields appear they allow Users to populate and save data that they were not meant to submit.


nathaelr:
Tested, it solved the original issue #4470

PR #4509, opened 2018-09-10
feature: Form Tokens

JediKev:
This adds a hidden input called "form_token" to ticket creation forms, response forms, and message forms that should prevent form resubmissions. This will address issues like duplicate tickets created on page refresh, etc.

### DIFF
--- a/include/ajax.tickets.php
+++ b/include/ajax.tickets.php
@@ -118,7 +118,7 @@ class TicketsAjaxAPI extends AjaxController {
 
         return $this->json_encode(array(
             'id'=>$lock->getId(), 'time'=>$lock->getTime(),
-            'code' => $lock->getCode()
+            'code' => $lock->getCode(), 'token'=>$lock->getToken()
         ));
     }
 
--- a/include/class.lock.php
+++ b/include/class.lock.php
@@ -20,6 +20,8 @@
 
 class Lock extends VerySimpleModel {
 
+    var $_token;
+
     static $meta = array(
         'table' => LOCK_TABLE,
         'pk' => array('lock_id'),
@@ -79,6 +81,13 @@ class Lock extends VerySimpleModel {
         return $this->code;
     }
 
+    function getToken() {
+        if (!$this->_token)
+            $this->_token = md5($this->lock_id.$this->code.strtotime($this->expire));
+
+        return $this->_token;
+    }
+
     //Renew existing lock.
     function renew($lockTime=0) {
         global $cfg;
@@ -90,6 +99,7 @@ class Lock extends VerySimpleModel {
             SqlFunction::NOW(),
             SqlInterval::MINUTE($lockTime)
         );
+        $this->_token = null;
         return $this->save(true);
     }
 
@@ -111,7 +121,7 @@ class Lock extends VerySimpleModel {
     //Create a ticket lock...this function assumes the caller checked for access & validity of ticket & staff x-ship.
     static function acquire($staffId, $lockTime) {
 
-        if (!$staffId or !$lockTime)
+        if (!is_int($staffId) or !$lockTime)
             return null;
 
         // Create the new lock.
@@ -145,6 +155,19 @@ class Lock extends VerySimpleModel {
         return static::objects()->filter(array(
             'expire__lt' => SqlFunction::NOW()
         ))->delete();
+    }
+
+    function validateToken($token=null, $code) {
+        if (!$code)
+            return false;
+
+        if (!($lock = Lock::lookup(array('code' => $code))))
+            return false;
+
+        if (strcmp($token, $lock->getToken()) !== 0)
+            return false;
+
+        return $lock->delete();
     }
 }
 ?>
--- a/scp/js/ticket.js
+++ b/scp/js/ticket.js
@@ -178,6 +178,10 @@
       // to the lock.code retrieved (if any)
       if (lock.code)
         $(this.options.lockInput, this.$element).val(lock.code);
+      // If there is an input with the name 'form_token', then set the value
+      // to the lock.signature retrieved (if any)
+      if (lock.token)
+        $('input[name=form_token]', this.$element).val(lock.token);
 
       // Deadband renew to every 30 seconds
       this.nextRenew = new Date().getTime() + 30000;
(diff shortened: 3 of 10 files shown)


JediKev:
This is Round 1. Please review/test and let me know what you find.

Cheers.

include/class.lock.php
```diff
@@ -90,6 +99,7 @@ function renew($lockTime=0) {
             SqlFunction::NOW(),
             SqlInterval::MINUTE($lockTime)
         );
+        $this->_signature = md5($this->lock_id.$this->code.strtotime($this->expire));
```
protich: Don't set hash in multiple places - getSignature should be the only spot.

```php
$this->_signature = null;
```

include/class.lock.php
```diff
@@ -124,8 +134,10 @@ static function acquire($staffId, $lockTime) {
             ),
             'code' => Misc::randCode(10)
         ));
-        if ($lock->save(true))
+        if ($lock->save(true)) {
+            $lock->_signature = md5($lock->lock_id.$lock->code.strtotime($lock->expire));
```
protich: Remove...

include/class.lock.php
```diff
@@ -146,5 +158,22 @@ static function cleanup() {
             'expire__lt' => SqlFunction::NOW()
         ))->delete();
     }
+
+    function validateToken($token=null, $code) {
+        if (!$code)
+            return false;
+
+        $lock = Lock::objects()
```
protich:
```php
if (!($lock = Lock::lookup(array('code' => $code)))
   return false;
```

include/class.lock.php
```diff
@@ -146,5 +158,22 @@ static function cleanup() {
...
+            ->filter(array('code' => $code))
+            ->first();
+
+        if (!$lock)
+            return false;
+
+        if ($token == md5($lock->getId().$lock->getCode().$lock->getExpireTime()))
```
protich: Never do this... it spreads hash all over the place (see my other comments). Also I think we should change `signature` to `token`.

```php
if (strcmp($token, $token->getToken()) !== 0)
    return false;

return $lock->delete();

```

include/class.ticket.php
```diff
@@ -2624,6 +2624,17 @@ function changeOwner($user) {
     function postMessage($vars, $origin='', $alerts=true) {
         global $cfg;
 
+        // Validate Form Token to prevent form resubmissions
+        if (!$vars['form_token'] || !$vars['lockCode'])
+            $errors['err'] = __('Invalid Form Token');
+
```
protich: Validation shouldn't happen inside any model... it's a `form` validation, not part of data validation.

include/class.ticket.php
```diff
@@ -2903,6 +2914,17 @@ function postCannedReply($canned, $message, $alert=true) {
     function postReply($vars, &$errors, $alert=true, $claim=true) {
         global $thisstaff, $cfg;
 
+        // Validate Form Token to prevent form resubmissions
+        if (!$vars['form_token'] || !$vars['lockCode'])
+            $errors['err'] = __('Invalid Form Token');
+
+        if (!$vars['token_validated'] && !Lock::validateToken($vars['form_token'], $vars['lockCode'])) {
```
protich: As mentioned above - resubmission prevention should be at form level. In this case it should be part of `Response Form`.

tickets.php
```diff
@@ -85,6 +85,8 @@
                     'message' => $_POST['message']
                     );
             $vars['cannedattachments'] = $attachments->getClean();
+            $vars['lockCode'] = $_POST['lock_code'] ?: null;
+            $vars['form_token'] = $_POST['form_token'] ?: null;
```
protich: This is a result of doing validation in the wrong place... results is mudding `$vars`

protich (changes requested):
First pass of review... ;)

scp/js/ticket.js
```diff
@@ -178,6 +178,10 @@
       // to the lock.code retrieved (if any)
       if (lock.code)
         $(this.options.lockInput, this.$element).val(lock.code);
+      // If there is an input with the name 'form_token', then set the value
+      // to the lock.signature retrieved (if any)
+      if (lock.signature)
```
protich: Change signature to token.

include/ajax.tickets.php
```diff
@@ -118,7 +118,7 @@ function acquireLock($tid) {
 
         return $this->json_encode(array(
             'id'=>$lock->getId(), 'time'=>$lock->getTime(),
-            'code' => $lock->getCode()
+            'code' => $lock->getCode(), 'signature'=>$lock->getSignature()
```
protich: This will crash... should be `'token' => $lock->getToken()`

protich (changes requested):

include/class.lock.php
```diff
@@ -146,5 +156,18 @@ static function cleanup() {
...
+        if (!$code)
+            return false;
+
+        if (!($lock = Lock::lookup(array('code' => $code))))
+            return false;
+
+        if (strcmp($token, $lock->getToken()) !== 0)
+            return false;
+
+        return $lock->delete();
```
protich: Are you sure you want to delete the lock here -  What about if the action (e.g posReply) errors? Re-acquiring the lock will change the token (due to new expire date - please confirm) - so I don't the need to delete here. Lock cleaning mechanism should handle deleting expired sessions.

jthakilla9:
So basically i should follow all the stept from above to every single step?

include/class.lock.php
jthakilla9: @protich , do i have to implement only this $this->_signature = null; or this to $this->_signature = md5($this->lock_id.$this->code.strtotime($this->expire));.

JediKev:
@jthakilla9

What..?

Just apply the changes from the Changes tab.

Cheers.

include/class.lock.php
JediKev: @jthakilla9

The changes are already implemented...just apply the changes from the Changes tab.

Cheers.

jthakilla9: Thanks, i will use the changes tab!

jthakilla9:
> @jthakilla9
> 
> What..?
> 
> Just apply the changes from the Changes tab.
> 
> Cheers.

Thank you! IT worked!

mrkylegp:
Tried applying these edits to v1.18.2 and get error 500 when trying to reply to a ticket.

PR #4515, opened 2018-09-13
issue: Company Name Variable

JediKev:
This addresses an issue where manually typing the User's email and name on the create ticket form causes the %{company.name} variable in the New Ticket Notice template to become the User's full name. This is due to `$this->getForm()->getClean()` which returns the incorrect values for the correct form fields.


TimzOwen (approved):
tested ok!

PR #4516, opened 2018-09-14
issue: Australian Timezones

JediKev:
This addresses an issue where Australian timezones (specifically `AEST`) will throw a fatal error when creating a task. This is due to the way we create the time in the backend where DateTime class "cannot find `AEST` in the database".


JediKev:
Not a good way to do this...need to just format the time stamp with a Format function. Will update soon.

PR #4538, opened 2018-09-27
Primary Queues as "Buckets"

protich:
This PR makes Primary Queues "buckets" for sub queues with forced inheritance of the criteria. Primary Queues are now hidden from dropdown menu unless there are no sub queues attached. Agents can still get to primary queue by clicking on it.

The PR also adds default sub queues for Closed Queue, utilizing DateTime periods.


protich:
@aydreeihn - Any feedback based on your review? 

I'm thinking we need to think about those already using queues. The change to not show the primary queue on the dropdown might cause great confusion.

include/class.misc.php
```diff
@@ -143,6 +143,85 @@ function micro_time() {
...
+    // Date range for the period in a given time
+    function date_range($period, $time=false) {
+        $time = $time ?: self::gmtime();
+        if (!($dt = Format::parseDateTime($time)))
+            return null;
+        // Force UTC
+        $dt->setTimezone(new DateTimeZone('UTC'));
+
+        // Make dt Immutable.
+        $dt = DateTimeImmutable::createFromMutable($dt);
```
aydreeihn: Looks like I'm missing the DateTimeImmutable class

PR #4540, opened 2018-09-28
Attachment Names Issue

aydreeihn:
This commit ensures that we will always get the correct attachment name regardless of if the file content is the same. Additionally, it ensures that the file_ids for attachments are compared in the correct order (elseifs)

### DIFF
--- a/include/class.thread.php
+++ b/include/class.thread.php
@@ -1084,10 +1084,9 @@ implements TemplateVariable {
         $ids = array();
         foreach ($files as $id => $info) {
             $F = array('inline' => is_array($info) && @$info['inline']);
+            $AF = null;
 
-            if (is_numeric($id) && $id != 0)
-                $fileId = $id;
-            elseif ($info instanceof AttachmentFile)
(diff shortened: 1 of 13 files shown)


include/class.thread.php
```diff
@@ -1085,12 +1085,12 @@ protected function normalizeFileInfo($files, $add_error=true) {
         foreach ($files as $id => $info) {
             $F = array('inline' => is_array($info) && @$info['inline']);
 
-            if (is_numeric($id) && $id != 0)
+            if (is_array($info) && isset($info['id']))
+                $fileId = $info['id'];
+            elseif (is_numeric($id) && $id != 0)
```
protich: I wonder if we still need this case - we should track down possibles just to make sure we don't end up with wrong attachments again.

protich: One scenario you didn't cover is email attachments processed via mail parse (piped emails). This would result in $info being array of file data to be created! Since $info['id'] doesn't exist yet - we end up with wrong references.

PR #4558, opened 2018-10-22
Issue/remove enums

aydreeihn:
Remove the 'state' enum column from the Thread Events table and create an Events table in its place (using event_id in the Thread Events table). This will improve the upgrade process and performance for databases with large amounts of data. Moving forward, we aim to eliminate the use of enums in the database.


include/class.report.php
```diff
@@ -77,26 +77,32 @@ function getDateRange() {
 
     function getPlotData() {
         list($start, $stop) = $this->getDateRange();
+        $states = array("created", "closed", "reopened", "assigned", "overdue", "transferred");
+        $event_ids = array();
+        foreach ($states as $state) {
+            $eid = Event::getIdByName($state);
```
protich: This is expensive  ....

do something like  
```php
$events = Event::objects()->value_flat('id')->filter(array('name__in', $states));
```

include/class.report.php
```diff
@@ -148,8 +154,8 @@ function getTabularData($group='dept') {
                ))
             ->constrain(array(
                 'thread__events' => array(
-                    'thread__events__state' => 'created',
-                    'state' => 'closed',
+                    'thread__events__event_id' => Event::getIdByName('created'),
```
protich: $events->findFirst(array('name' => 'created'));

PR #4580, opened 2018-10-29
queue: Top-Level Ticket Counts

JediKev:
This adds a setting in the Ticket Settings to give the option of showing the ticket counts on Top-Level queues. If enabled, this will populate the ticket counts next to the Top-Level queue names like they used to be. The only caveat is the counts are triggered by mousing-over any of the queue names as we only load the counts when we need to display the sub-queues (for better performance).

This also adds a new format function called `number` (Phase 1). As of now it will format any integer to a comma separated number (eg. `1337 -> 1,337`).

### DIFF
--- a/include/class.config.php
+++ b/include/class.config.php
@@ -767,6 +767,10 @@ class OsticketConfig extends Config {
         return $sequence;
     }
 
+    function showTopLevelTicketCounts() {
+        return ($this->get('queue_bucket_counts'));
+    }
+
     function getDefaultTicketNumberFormat() {
         return $this->get('ticket_number_format');
     }
@@ -1277,6 +1281,7 @@ class OsticketConfig extends Config {
         return $this->updateAll(array(
             'ticket_number_format'=>$vars['ticket_number_format'] ?: '######',
             'ticket_sequence_id'=>$vars['ticket_sequence_id'] ?: 0,
+            'queue_bucket_counts'=>isset($vars['queue_bucket_counts'])?1:0,
             'default_priority_id'=>$vars['default_priority_id'],
             'default_help_topic'=>$vars['default_help_topic'],
             'default_ticket_status_id'=>$vars['default_ticket_status_id'],
--- a/include/class.format.php
+++ b/include/class.format.php
@@ -503,6 +503,22 @@ class Format {
         return implode( $separator, $string );
     }
 
+    function number($number, $locale=false) {
+        if (is_array($number))
+            return array_map(array('Format','number'), $number);
+
+        if (!is_numeric($number))
+            return $number;
+
+        if (extension_loaded('intl') && class_exists('NumberFormatter')) {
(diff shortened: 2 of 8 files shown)


greezybacon:
I think `number_format` may be influenced by the server locale over the configured Agent's locale? If so, this may be a better implementation of `Format::number`:?
```php
function number($number, $locale=false) {
    if (is_array($number))
        return array_map(array('Format','number'), $number);

    if (extension_loaded('intl') && class_exists('NumberFormatter')) {
        $nf = NumberFormatter::create($locale ?: Internationalization::getCurrentLocale(),
            NumberFormatter::DECIMAL);
        return $nf->format($number);
    }

    return number_format((int) $number);
}
```
Actually, it doesn't look like `number_format` respects the locale at all?

JediKev:
@greezybacon 

LOL I mentioned this is only Phase 1 but yes that's exactly what I was aiming for in Phase 2 😂

Might go ahead and test your suggestion and implement it here so it's already done.

Cheers.

include/class.config.php
```diff
@@ -1277,6 +1281,7 @@ function updateTicketsSettings($vars, &$errors) {
         return $this->updateAll(array(
             'ticket_number_format'=>$vars['ticket_number_format'] ?: '######',
             'ticket_sequence_id'=>$vars['ticket_sequence_id'] ?: 0,
+            'ticket_counts'=>isset($vars['ticket_counts'])?1:0,
```
protich: I wouldn't call this `ticket_counts` - I think something like `queue_bucket_counts` would be more appropriate.

include/class.config.php
```diff
@@ -767,6 +767,10 @@ function getDefaultTicketSequence() {
         return $sequence;
     }
 
+    function getTopLevelTicketCounts() {
```
protich: In the spirit of making it consistent - I think this function should be `showTopLevelTicketsCounts()` and `return ($this->get('ticket_counts'));`

include/class.format.php
```diff
@@ -503,6 +503,13 @@ function array_implode( $glue, $separator, $array ) {
         return implode( $separator, $string );
     }
 
+    function number($number) {
+        if(is_array($number))
+            return array_map(array('Format','number'), $number);
+
```
protich:
```php
if (!is_numeric($number)) 
   return $number;
```
Also recommend adding locale suggested by @greezybacon.

include/class.format.php
```diff
@@ -503,6 +503,22 @@ function array_implode( $glue, $separator, $array ) {
         return implode( $separator, $string );
     }
 
+    function number($number) {
```
protich: Should be 

```php

function number($number, $locale=null) { 
```

JediKev: swear I committed that change..updating in a sec

JediKev: updated

PR #4591, opened 2018-11-01
feature: Public Mark As Answered/Unanswered

JediKev:
This feature extends the ability to mark tickets as Answered/Unanswered to any Agent with the Role Permission to Post Reply in the Department. This also extends the Mark As Answered option to Internal Notes and Change Status. If you Post an Internal Note you will see a checkbox to "Mark As Answered" same goes for changing a Ticket's Status to a status that status is not "closed" or "resolved".

### DIFF
--- a/include/ajax.tickets.php
+++ b/include/ajax.tickets.php
@@ -1088,6 +1088,24 @@ function refer($tid, $target=null) {
         $state = strtolower($status->getState());
 
         if (!$errors && $ticket->setStatus($status, $_REQUEST['comments'], $errors)) {
+            // Mark as Answered
+            if ($_POST['status_answered'] && ($_POST['status_answered'] == 'on')) {
+                if (!$ticket->getDept()->isManager($thisstaff)
+                  && !$role->hasPerm(Ticket::PERM_REPLY))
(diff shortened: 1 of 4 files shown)


JediKev:
Would it be worth moving the js to `scp/js/scp.js` so it’s not repeated? It’s only repeated twice.. but still. Thoughts?

protich:
Not sure I like the idea of overloading status change & internal note... tickets are auto-marked answered or unanswered, meaning rarely do agents need to manually change the flag. Basically it's an exception to the rule and it shouldn't be part of core operation like changing status.

ntozier:
> Tickets are auto-marked answered or unanswered, meaning rarely do agents need to manually change the flag.

I'm going to have to disagree with this sentence.  The people I work with have a tendency to always respond.  So if I tell someone "Contacted vendor, will let you know what they say"  The staff (user) will usually reply with Thank you, or Okay or something.  Which makes the ticket bold in the ticket list.  I would love to have a way to just mark it answered (and deBold it) because bold to me means that it needs my attention.

protich:
The reason we’re adding the feature is for use cases like yours, but it shouldn’t be primary. Read my comment in context.

PR #4612, opened 2018-11-19
feature: Public Mark As Answered/Unanswered

JediKev:
This feature extends the ability to mark tickets as Answered/Unanswered to any Agent with the Role Permission to Post Reply in the Department. This also adds the ability to add an (optional) Internal Note when marking as (un)answered.


include/ajax.tickets.php
```diff
@@ -1256,6 +1256,64 @@ function setSelectedTicketsStatus($state) {
...
+                    elseif (!$ticket->markUnanswered())
+                        $errors['err'] - __('Cannot mark ticket as unanswered');
+                    break;
+
+                default:
+                    Http::response(404, __('Unknown action'));
+            }
+
+            $comments = $form->getComments();
+            if ($comments) {
```
protich: Comment should only be added on a successful action..... move the comment code block to else block below.

PR #4629, opened 2018-12-10
Improve the Staff login

pitbulk:
- Follow the same pattern that the client login https://github.com/pitbulk/osTicket/blob/develop/login.php#L48 . Otherwise StaffAuthenticationBackend that receives a POST are not possible.
- Refresh $msg with the value stored in session if the StaffAuthenticationBackend backend defines it, otherwise the initial "Authentication Required" is prompted.

### DIFF
--- a/scp/login.php
+++ b/scp/login.php
@@ -40,7 +40,8 @@ if($_POST) {
 
     // Rotate the CSRF token (original cannot be reused)
     $ost->getCSRF()->rotate();
-
+}
+if ($_POST && isset($_POST['userid'])) {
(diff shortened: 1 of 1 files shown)


perryk:
Not an issue, simply a note. This conflicts with #3955 , which I assume is no longer needed.

Cheers.

JediKev:
@perryk 

Just to add a note to your note (lol), #3955 has already been merged into `develop-next` (way before this was created) so it doesn't matter.

Cheers.

perryk:
Great, thanks @JediKev.

PR #4630, opened 2018-12-11
Issue: Duplicate Search Results

aydreeihn:
This commit ensures that we do not return duplicate results when searching for Tickets with an Advanced Search that contains keywords.


protich:
I think it would make sense to do the distinct in `include/staff/templates/queue-tickets.tmpl.php` and use `ticket_id` instead if `number`. See staged fix in SS.

aydreeihn:
Updated to mirror the production fix.

PR #4640, opened 2018-12-28
Task Collaborators

aydreeihn:
This commit changes the wording for Collaborators on Tasks to match the wording for Collaborators on Tickets.


koebelin:
Thank you!

protich:
I wonder if we should be more specific and use the word  "Email Collaborators" since collaboration happens purely via email on tasks.

aydreeihn:
Idk. I feel like it looks a little awkward. Maybe it's just because I'm so used to seeing it the other way.

screen shot 2019-01-08 at 4 00 22 pm

PR #4649, opened 2019-01-02
Issue: Edit Export Column Heading

aydreeihn:
This commit addresses an issue where we did not save headings for export fields.
We just need to leave the heading in the array rather than filtering it out so that we can save its new heading.

### DIFF
--- a/include/class.queue.php
+++ b/include/class.queue.php
@@ -1141,9 +1141,6 @@ class CustomQueue extends VerySimpleModel {
             return false;
 
         $order = array_keys($fields);
-        // Filter exportable fields
-        if (!($fields = array_intersect_key($this->getExportableFields(), $fields)))
-            return false;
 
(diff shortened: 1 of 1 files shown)


protich:
I'm curious to investigate why we had the filter in the first place. It has been a minute, but I think the thinking was limit fields to only exportable fields - which might be mute anyhow but please look into it.

aydreeihn:
I think that's what the intent was, however, all array_intersect_key really did was take the 'heading' key out of the $fields array which was what caused the code to lose the custom headings.

Ex:
 ["number"]=> string(13) "Ticket Number"

So now, $fields looks like this:
{ ["number"]=> array(2) { ["name"]=> string(13) "Ticket Number" ["heading"]=> string(16) "Ticket Number Custom Name" }

And the custom heading can be saved.

PR #4652, opened 2019-01-03
Issue: Can’t use “0” in Short Answer Field

aydreeihn:
This commit fixes an issue where a Short Answer Field would show as Empty if the value saved in it was '0'

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -1478,6 +1478,11 @@ class TextboxField extends FormField {
     function parse($value) {
         return Format::striptags($value);
     }
+
+    function display($value) {
+        $value = $value ?: $this->value;
+        return ($value == 0) ? '&#48' : Format::htmlchars($this->toString($value));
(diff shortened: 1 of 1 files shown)


protich:
I don't like the implementation of the fix - I think the logic should go to `TextboxField` class instead.

aydreeihn:
Updated. Looked like I was making it more complicated than it needed to be.

tomlaf:
Just try this merge and on my setup all the ticket number and ticket subject now just show 0.

Exemple : 

Without this merge request (on 1.11.x) : 
image
With :
image

Any idea how to correct it ?

greezybacon:
@aydreeihn what data are you testing this with?!

A triple equal would be better as it does not confuse types (0 == '0'), but (0 !== '0'). 

Furthermore, don't you need a semicolon after the HTML entity? (e.g. `&#48`**;**)

PR #4653, opened 2019-01-03
Issue: Tasks Within Tickets

aydreeihn:
This commit addresses issues we had with viewing Task(s) within a Ticket:

1. When viewing the table of all Tasks on a Ticket, the Options dropdown should only allow the Agent to Reopen or Close the Tasks based on the status of the Tasks.
Ex: If there is only 1 Open Task, you should only see the 'Close' Option
    If there are 2 Closed Tasks, you should only see the 'Reopen' Option
    If there are multiple Tasks in which some are Open and some are Closed, you should see both the 'Reopen' and 'Close' options

2. When viewing an individual Task within a Ticket, the status options were the opposite of what they should have been. Now, if the Task is Open, the Agent will see the option to 'Close' the Task. If the Task is Closed, the Agent will see the option to 'Reopen' the Task.

### DIFF
--- a/include/staff/ticket-tasks.inc.php
+++ b/include/staff/ticket-tasks.inc.php
@@ -40,12 +40,17 @@ $showing = $pageNav->showing().' '._N('task', 'tasks', $count);
             print __('Add New Task'); ?></a>
     <?php
     }
+    foreach ($tasks as $task)
+        $taskStatus .= $task->isOpen() ? 'open' : 'closed';
+
     if ($count)
         Task::getAgentActions($thisstaff, array(
                     'container' => '#tasks_content',
                     'callback_url' => sprintf('ajax.php/tickets/%d/tasks',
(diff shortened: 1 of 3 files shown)


include/staff/ticket-tasks.inc.php
```diff
@@ -40,12 +40,17 @@ class="green button action-button ticket-task-action"
             print __('Add New Task'); ?></a>
     <?php
     }
+    foreach ($tasks as $task)
```
protich: To me it looks like the '$taskStatus` will be set to the status of the last task - is that the intended?

aydreeihn: No, it is appending to one string, so it would have all of the values. Then in tasks-actions.tmpl.php, we determine which options belong in the list (Reopen/Close) based on that status string. Let me know if you would like this implemented a different way.

Issue #4382, opened 2018-07-31
SLA change Hours to Minutes


johnkalangha:
Osticket Version: v1.10
Mysql
Php 7.2
apache server

anyone can help me regarding in here issue.. my boss want to change all our SLA Help topics for about to 10 mins to 20 mins ...

JediKev:
@johnkalangha

That is not a feature of osTicket at this time.

Cheers.

johnkalangha:
huhu, is there any solution even though for temporary?

JediKev:
@johnkalangha

Not that I know of. You can search around on our Forum here for any possible temporary, third-party workaround:
http://osticket.com/forum

Cheers.

johnkalangha:
Copy sir, thanks and god bless

PR #4657, opened 2019-01-10
Give solution of issues #4382 SLA change Hours to Minutes

ghost:
And in this code changed It will reflect in

- SLAPlan list
- When we create SLA Plan
- When we Create Ticket that time SLA Plane list
- When cron runs


JediKev:
@sahil1986 

Thank you for this pull request, but we are going to revamp SLA as a whole very soon.

Also, a better improvement would be to provide the option for minutes or hours instead of forcing minutes. 😉

Cheers.

JediKev:
@mabibby

We don’t have public roadmaps/feature blueprints at this time. This is something on our roadmap for sure and will be implemented.

Please stay tuned! Cheers.

PR #4667, opened 2019-01-22
+ Make osTicket full-width

pedro-lb:
This css change makes osTicket full-width, improving the UI.

The change affects all pages on osTicket, including Admin and User panels, while respecting the actual layout.

Preview here:
image

### DIFF
--- a/include/client/header.inc.php
+++ b/include/client/header.inc.php
@@ -48,9 +48,9 @@ if ($lang) {
     <script src="<?php echo ROOT_PATH; ?>js/osticket.js"></script>
     <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/filedrop.field.js"></script>
     <script src="<?php echo ROOT_PATH; ?>scp/js/bootstrap-typeahead.js"></script>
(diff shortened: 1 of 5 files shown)


koebelin:
I've done this in 1.10.1 and 1.10.4 in our installation by going through each scp file containing UI and removing the width attributes and all the ancient html attributes (cellpadding, cellspacing, border) for all tables and columns, and only then changing #container to width:100%.  Then you don't need !important logic and the html is cleaned.

Like in tasks.inc.php:
 <table class="list" border="0" cellspacing="1" cellpadding="2" width="940"> becomes 
<table class="list">

Then you can define table.list in scp.css with "width:100%; border:1px solid #000".

I've just been too lazy and and intimidated to try to make a pull request involving dozens of files.

An easier fix is welcome for next release.

JediKev:
@pedro-lb 

Thank you for this but we are about to go through complete UI/UX changes in `v2.0`. Please stay tuned.

Cheers.

pedro-lb:
@JediKev 

Thanks for the info! I'll stay tuned for the news.

Cheers!

PR #4695, opened 2019-02-07
ui:New Ticket

jdelhome3578:
New Ticket is bold and showing active when any other top level queue is active. This fixed the issue.


protich:
Thank you for the PR - however removing the active link will kill auto-bolding systemwide. PR #4724 correctly addresses the root cause of the issue.

Issue #4698, opened 2019-02-07
1.11: Custom Queues Annotations


jdelhome3578:
Unable to edit and delete annotations in any queue. I am able to add annotations but when going back to edit them it shows non exist. They do populate on the queue list.

PHP 5.6.38
IIS

JediKev:
@jdelhome3578 

Here you go:
#4707

Cheers.

PR #4707, opened 2019-02-08
issue: Column Annotations

JediKev:
This addresses issue #4698 where you are not able to edit/delete column annotations in the queue edit. This is due to the JS not removing the hidden class.

### DIFF
--- a/include/staff/templates/queue-column.tmpl.php
+++ b/include/staff/templates/queue-column.tmpl.php
@@ -86,6 +86,7 @@ foreach (Internationalization::sortKeyedList($annotations) as $class=>$desc) {
           position.attr('name', position.data('name'));
           if (pos)
             position.val(pos);
+          clone.removeClass('hidden');
           template.closest('.tab_content').find('.empty').hide();
         };
         $('select.add-annotation', '#annotations').change(function() {


greezybacon:
I found this in my Redactor upgrade. Also, the `Preview` tab does not work. I fixed both in 2bdebfa3f189049fa1bec82c9eb9e04f656d4083 in my Redactor pull request.

For this issue, the root cause is on line 75: the `show()` function no longer drops the `hidden` class.

JediKev:
@greezybacon

Nice! I’ll close this then. Thanks for pointing that out!

ICYWW:
My thinking was to hide after the build of the clone but your fix seems cleaner.

Cheers.

greezybacon:
Well, perhaps I should propose a separate pull request with just that one commit? I hate to hold up the release on the stability of the new Redactor

JediKev:
@greezybacon

I think that seems like a good idea as it’s quite annoying to not be able to delete or edit existing/new annotations lol

Cheers.

greezybacon:
Allright, @JediKev you've got #4711

protich:
Addressed in PR #4711

PR #4711, opened 2019-02-09
queues: Fix compatibility issues with newer jQuery

greezybacon:
This is a continuation of the conversation from #4707


greezybacon:
Oops. This breaks sequence editing

PR #4728, opened 2019-02-13
Quick Filter Fixes:

aydreeihn:
Make the following quick filters work:
- Assigned
- Assignee
- Department Manager
- Assigned Team

Also, make sure we account for if there are no choices available for a quick filter 
Ex: The helpdesk does not have any teams

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -1980,6 +1980,24 @@ class ChoiceField extends FormField {
     }
 
     function applyQuickFilter($query, $qf_value, $name=false) {
+        global $thisstaff;
+
+        //special assignment quick filters
+        switch (true) {
+            case ($qf_value == 'assigned'):
+            case ($qf_value == '!assigned'):
+                $result = AssigneeChoiceField::getSearchQ($qf_value, $qf_value);
+                return $query->filter($result);
+            case (strpos($qf_value, 's') !== false):
+            case (strpos($qf_value, 't') !== false):
+            case ($qf_value == 'M'):
+            case ($qf_value == 'T'):
+                $value = array($qf_value => $qf_value);
+                $result = AssigneeChoiceField::getSearchQ('includes', $value);
+                return $query->filter($result);
+                break;
+        }
+
         return $query->filter(array(
             $name ?: $this->get('name') => $qf_value,
         ));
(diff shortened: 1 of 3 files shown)


include/class.search.php
```diff
@@ -1399,8 +1406,14 @@ class DepartmentManagerSelectionField extends AgentSelectionField {
     static $_members;
 
     function getChoices($verbose=false) {
-        if (isset($this->_members))
-            $this->_members = Staff::getStaffMembers();
+        if (!isset($this->_members)) {
+            $managers = array();
+            foreach (Staff::objects() as $staff) {
```
protich: Why not filter via ORM and select only managers?

include/class.forms.php
```diff
@@ -1980,6 +1980,24 @@ function getQuickFilterChoices() {
...
 
     function applyQuickFilter($query, $qf_value, $name=false) {
+        global $thisstaff;
+
+        //special assignment quick filters
+        switch (true) {
+            case ($qf_value == 'assigned'):
+            case ($qf_value == '!assigned'):
+                $result = AssigneeChoiceField::getSearchQ($qf_value, $qf_value);
+                return $query->filter($result);
```
greezybacon: Shouldn't this be defined in `AssigneeChoiceField::applyQuickFilter`?

aydreeihn: We could do that, but we would end up needing to define an applyQuickFilter function for AssignedField, AssigneeChoiceField, and TeamSelectionField. 

applyQuickFilter is hit in class.queue.php getQuery 
$qf->applyQuickFilter()
where $qf could be one of those classes depending on the quick filter.

After discussing it, it looks like it'll be fine to just leave it as is.

include/class.search.php
aydreeihn: Off of departments?

```
if (!isset($this->_members)) {
            $managers = array();
            $managerIds = Dept::objects()
                ->filter(Q::any(array(
                    'manager_id__gt' => 0,
                )))
                ->values_flat('manager_id');

            foreach ($managerIds as $key => $id) {
                $staff = Staff::lookup($id[0]);
                $managers['s'.$id[0]] = $staff->getName()->name;
            }
            $this->_members = $managers;
        }
```

Does this make it less expensive? I still ended up with a lookup on the staff member.

Issue #4720, opened 2019-02-12
Mail fetcher - new email fails to create ticket


TechCare-Aus:
osTicket Version | v1.11 (7b1eee8) —  Up to date
Web Server Software | Apache 2.4
MySQL Version | 10.2.21
PHP Version | 7.2.14

Updated today via Installatron.

**Issue:**
Prior to running PHP 7.2 I was running PHP 7.1 and if I sent an email to support@mycompany.com it would show up in the inbox (unread). Cron would poll and then mail fetcher would do it's thing and place it in the folder RETRIEVED as read. But nothing would show up in osTicket.

If I created a ticket inside osTicket, then replied to it (so that it would send an email out to the end user) they could reply and osTicket would fetch that and business as usual. 

But never did it create a NEW ticket off an email sent through to the primary email mailbox.

This was working previously. 

Now I just on the off hand decided to turn on PHP 7.2 and then bam... all is well again. The only error showing in php is:

\var\log\apache2\error_log = `PHP Warning:  array_merge(): Argument #2 is not an array in /home/USERNAME/public_html/support/include/class.plugin.php on line 290`

\home\USERNAME\log\cron_error.log = `PHP Fatal error:  APCuIterator::__construct(): APC must be enabled to use APCuIterator in /home/USERNAME/public_html/support/include/class.search.php on line 966`

This is from when I ran PHP 7.1

\home\USERNAME\log\cron_error.log = `[12-Feb-2019 00:53:01 UTC] PHP Warning:  Declaration of TransferForm::render($options) should be compatible with Form::render($staff = true, $title = false, $options = Array) in /home/USERNAME/public_html/support/include/class.forms.php on line 4487`

\var\log\apache2\error_log = `[core:error] [pid 15415:tid 46933651949312] (104)Connection reset by peer: [client XXX.XXX.XXX.XXX:50942] AH00574: ap_content_length_filter: apr_bucket_read() failed`

I couldn't find any other errors. But I'm happy to post anything you like. I'm running WHM and have full access to anything on the server.

Siriuo:
Its good to know I am not the only one having this issue with osTicket v1.11. Only difference here is that I am running apache 2.2, php 5.6, mysql 5.5.62 (not mariadb), also tried php 7.2 with the same issue.

For some reason I am not seeing any errors in the logs. However it appears only some emails are not being created. I am trying to find similarities in the email itself.

Siriuo:
[...]

PR #4731, opened 2019-02-13
issue: APC CLI

JediKev:
This addresses issue #4720 where creating tickets via CLI fails to set a status and crashes with an "APC must be enabled" error. This adds a check to see if the APC module is loaded before continuing.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -3173,7 +3173,7 @@ implements RestrictedAccess, Threadable, Searchable {
     function save($refetch=false) {
         if ($this->dirty) {
             $this->updated = SqlFunction::NOW();
-            if (isset($this->dirty['status_id']))
+            if (isset($this->dirty['status_id']) && PHP_SAPI !== 'cli')
                 // Refetch the queue counts
                 SavedQueue::clearCounts();
         }


pablopol24:
**it did not work in my case with:**
if (extension_loaded('apc') && function_exists('apcu_store')) {
**It works with:**
if (PHP_SAPI !== 'cli' && function_exists('apcu_store')) {

JediKev:
@pablopol24 

Hmmm...

What OS are you running? CentOS, Ubuntu, etc. 

What build of PHP?

Also, how did you apply this fix? Did you copy/paste, apply via git, etc.

Cheers.

JediKev:
@pablopol24 

The error is `APC must be enabled to use APCuIterator` which means the APC module is not enabled/loaded. In PHP the way you check that is `extension_loaded('apc')`.

Not sure why this is not working for you but working fine for me. Maybe you mistyped something? Try applying the diff or copy pasting from the pull request.

Cheers.

include/class.search.php
```diff
@@ -960,7 +960,7 @@ static function storeCounts($key, $counts, $ttl) {
     }
 
     static function clearCounts() {
-        if (function_exists('apcu_store')) {
+        if (extension_loaded('apc') && function_exists('apcu_store')) {
```
greezybacon: I'm not so sure. On PHP >= 5.6 the extension is called `acpu`. So I think your fix is just disabling the block altogether? Because `function_exists('acpu_store')` should be enough to detect the presence of ACPu?

greezybacon:
Should the fix maybe go into `class.ticket.php` around line 3173:
```php
    function save($refetch=false) {
        if ($this->dirty) {
            if (isset($this->dirty['status_id']) && PHP_SAPI !== 'cli')
                // Refetch the queue counts
                SavedQueue::clearCounts();
        }
        return parent::save($this->dirty || $refetch);
    }    
```

include/class.search.php
JediKev: `function_exists('apcu_store')` is not adequate enough to detect this as it fails for me every time when piping via CLI. I receive the error `APC must be enabled to use APCuIterator` which means APC is not enabled. So, I'm checking to make sure APC is enabled before continuing. I do see where this can be an issue so there has to be a better way other than detecting cli and this.

JediKev:
@greezybacon 

There has to be a better way other than detecting CLI right?

Cheers.

adlerweb:
> The error is `APC must be enabled to use APCuIterator` which means the APC module is not enabled/loaded. In PHP the way you check that is `extension_loaded('apc')`.

I can't test atm, but as far as I remember tell the APCu extension **is** loaded, just not active due to _apc.enable_cli_ beeing false in the default config. [symfony](https://github.com/symfony/symfony/issues/29877) has a similar problem and only proposal atm is also to check the used SAPI first.

pablopol24:
> @ pablopol24
> 
> Hmmm ...
> 
> ¿Qué sistema operativo estás ejecutando? CentOS, Ubuntu, etc.
> 
> ¿Qué compilación de PHP?
> 
> Además, ¿cómo aplicaste esta corrección? Copiaste / pegaste, aplicaste vía git, etc.
> 
> Aclamaciones.

Ubuntu 18.04
PHP 7.2
Osticket 1.11

copy and paste in both cases

greezybacon:
According to [StackOverflow](url
https://stackoverflow.com/a/13911327),  ```ini_get('apc.enabled')```?

adlerweb:
> According to [StackOverflow](https://stackoverflow.com/a/13911327), `ini_get('apc.enabled')`?
That is only true for other SAPIs, CLI like the cron used for mails ignores apc.enabled and uses apc.enable_cli instead.

bradgillap:
The fix op posted in this thread worked for me in production with two departments and several users.

Server Information
--
osTicket Version | v1.11 (0f229ae) —  Up to date
Web Server Software | Apache/2.4.25 (Debian)
MySQL Version | 10.1.37
PHP Version | 7.2.15-1+0~20190209065123.16+stretch~1.gbp3ad8c0

gdlib | Used for image manipulation and PDF printing
imap | Used for email fetching
xml | XML API
xml-dom | Used for HTML email processing
json | Improves performance creating and processing JSON
mbstring | Highly recommended for non western european language content
phar | Highly recommended for plugins and language packs
intl | Highly recommended for non western european language content
fileinfo | Used to detect file types for uploads
APCu | Improves overall performance
Zend Opcache | Improves overall performance

cgi.fix_pathinfo | "1" is recommended if AJAX is not working
date.timezone | America/Toronto

Schema | osticket (localhost)
Schema Signature | 
Space Used | 13.66 MiB
Space for Attachments | 10.47 MiB
Timezone | EST (Interpreted as America/New_York)




Thanks!

clonemeagain:
Typically APC is only enabled for cgi/fpm/Apache, not cli, enabling tools like composer and systems like laravel/symfony to work properly in cli mode, because of annotations and apc stripping them.
Each of those PHP use cases will have their own ini and module configs. This would explain why simply detecting the sapi type works

PR #4734, opened 2019-02-14
ui:add plugin location

jdelhome3578:
Show install location. Helps to know which file is installed/to install. Adds to the extra information #3486.


JediKev:
@jdelhome3578 

Just wondering, how is this useful? The plugin names are relevant to what they are (ie. `auth-ldap.phar` is ldap authentication, `storage-fs.phar` is file storage on the filesystem, etc.), no two plugins are named the same, and all plugins are always in the plugins folder...what is the benefit of this?

Cheers.

jdelhome3578:
@JediKev When testing and trying to help develop fixes. I have different versions of the same plugin.

Issue #4628, opened 2018-12-07
Removing Organization will result in error


clandmeter:
Im running an up to date 1.11.x branch (31c129a2667d3dc6e4bb74254b9686b66594c51d) on PHP 7.0.
When removing an organization the below error will be thrown.

Removing the users cookie will make osTicket work again.

```
[Fri Dec  7 07:43:37 2018] PHP Fatal error:  Uncaught DoesNotExist in /usr/share/webapps/osTicket/include/class.orm.php:1353
Stack trace:
#0 /usr/share/webapps/osTicket/include/class.orm.php(721): QuerySet->one()
#1 /usr/share/webapps/osTicket/include/class.orm.php(338): VerySimpleModel->refetch()
#2 [internal function]: VerySimpleModel->__wakeup()
#3 /usr/share/webapps/osTicket/include/class.ostsession.php(86): session_start()
#4 /usr/share/webapps/osTicket/include/class.ostsession.php(130): osTicketSession->__construct(86400)
#5 /usr/share/webapps/osTicket/include/class.osticket.php(60): osTicketSession::start(86400)
#6 /usr/share/webapps/osTicket/include/class.osticket.php(628): osTicket->__construct()
#7 /usr/share/webapps/osTicket/main.inc.php(33): osTicket::start()
#8 /usr/share/webapps/osTicket/client.inc.php(21): require_once('/usr/share/weba...')
#9 /usr/share/webapps/osTicket/index.php(16): require('/usr/share/weba...')
#10 {main}
  thrown in /usr/share/webapps/osTicket/include/class.orm.php on line 1353
[Fri Dec  7 07:43:37 2018] 0.0.0.0:34245 [500]: / - Uncaught DoesNotExist in /usr/share/webapps/osTicket/include/class.orm.php:1353
Stack trace:
#0 /usr/share/webapps/osTicket/include/class.orm.php(721): QuerySet->one()
#1 /usr/share/webapps/osTicket/include/class.orm.php(338): VerySimpleModel->refetch()
#2 [internal function]: VerySimpleModel->__wakeup()
#3 /usr/share/webapps/osTicket/include/class.ostsession.php(86): session_start()
#4 /usr/share/webapps/osTicket/include/class.ostsession.php(130): osTicketSession->__construct(86400)
#5 /usr/share/webapps/osTicket/include/class.osticket.php(60): osTicketSession::start(86400)
[...]

PR #4740, opened 2019-02-15
issue: Delete Org Session Failure

JediKev:
This addresses an issue on the Forum where deleting an Organization causes the session to fail. This is due to the Organization QuerySet being stored in the session and not being cleared out when deleted from the system. This causes the system to try to refetch the object from the database which it's not there as it's been deleted and therefore causes the session to crash.

### DIFF
--- a/include/class.organization.php
+++ b/include/class.organization.php
@@ -486,6 +486,9 @@ implements TemplateVariable, Searchable {
         if (!parent::delete())
             return false;
 
+        // Clear organization from session to avoid refetch failure
+        unset($_SESSION[':Q:orgs'], $_SESSION[':O:tickets']);
+
         // Remove users from this organization
         User::objects()
             ->filter(array('org' => $this))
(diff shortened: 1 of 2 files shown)


JediKev:
Jk 😅 This was raised here:
#4628

include/class.organization.php
```diff
@@ -471,6 +471,9 @@ function delete() {
         if (!parent::delete())
             return false;
 
+        // Clear organization from session to avoid refetch failure
+        unset($_SESSION[':Q:orgs'], $_SESSION[':O:tickets']);
```
protich, line 490: Why are you clearing tickets here?

JediKev: The org is stored there as well. Maybe I shoul just clear the org from tickets? I can do something like:
```
unset($_SESSION[':Q:orgs'], $_SESSION[':O:tickets'][':O:'.$id]);
```

protich: Let me try replicating the issue and revert back to you with thoughts...

JediKev: I can also send you a dump of my session if you want.

PR #4744, opened 2019-02-19
Assignment Restriction Issue

aydreeihn:
This commit fixes an issue where we were not checking the restrictions on Ticket assignment correctly. isAccessLimited returns true if showAssignedOnly is true, so we wouldn't want to negate the isAccessLimited check (that would return true if access is not limited).

Also, if an Agent's access is limited to assigned tickets only, we can return the assigned tickets once we have them without needing to go through any further processing.

Note: queue counts update themselves a little after you toggle the 'Limit ticket access to ONLY assigned tickets' check box. The 'See all tickets in search results, regardless of access' still works correctly as well.

### DIFF
--- a/include/class.staff.php
+++ b/include/class.staff.php
@@ -597,6 +597,10 @@ implements AuthenticatedUser, EmailContact, TemplateVariable, Searchable {
 
         $visibility = Q::any(new Q(array('status__state'=>'open', $assigned)));
 
+        // -- If access is limited to assigned only, return assigned
+        if ($this->isAccessLimited())
+            return $visibility;
+
         // -- Routed to a department of mine
         if (($depts=$this->getDepts()) && count($depts)) {
             $visibility->add(array('dept_id__in' => $depts));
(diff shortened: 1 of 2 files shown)


theking81:
Thank you :)

protich:
I agree to the change to `getTicketsVisibility()` - but  can you test a limited agent clicking a ticket they only have access to via assignment?

theking81:
I tested it and yes the can only have access to assigned tickets. They can see them in open status tab and in the my tickets tab.

aydreeihn:
When testing it, it's true that agents still have access to tickets that they only have access to via assignment as long as the ticket is open. Shouldn't they still have access though even if the ticket is closed? I modified my code to allow that access, but if we still want to restrict it, just let me know and I can put that restriction back.

protich:
@aydreeihn - No -  closed tickets are no longer assigned, since closing agent might not necessarily be the agent who was assigned.

To extend access to previously assigned agents or closing agents we would have to add auto-referral (which is really assignment) with a setting to enable it.

Issue #4743, opened 2019-02-19
V1.11: Unable to post internal note with ticket creation


Syphere:
When someone creates a new ticket with an internal note written, the ticket opens as expected but the internal note is ignored and not included.
The system is running a recent upgrade from 1.10, where ticket creation did work with internal notes.

Logs are empty and clean and have yet to show any errors since the upgrade.

JediKev:
@Syphere 

Here you go:
#4745

Cheers.

doczkal:
Though #4745 is merged now I can see this is still an open issue with osTicket (v1.14.3)

PR #4745, opened 2019-02-19
issue: Internal Note Ignored

JediKev:
This addresses issue #4743 where the system ignores Internal Notes on Ticket Creation. This is due to the ticket being assigned and not having the Internal Note passed along.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -4120,7 +4120,7 @@ implements RestrictedAccess, Threadable, Searchable {
         }
 
         // Not assigned...save optional note if any
-        if (!$ticket->isAssigned() && $vars['note']) {
+        if (!$vars['assignId'] && $vars['note']) {
             if (!$cfg->isRichTextEnabled())
                 $vars['note'] = new TextThreadEntryBody($vars['note']);
             $ticket->logNote(_S('New Ticket'), $vars['note'], $thisstaff, false);


protich:
Did you actually replicate the issue as descried and truly tested your PR? 

Asking because the changes  herein have nothing to do with the original issue as described by the poster -  the code you changed deals with auto-assignment via filters which shouldn't carryover internal notes. 

What you needed to look into was code block around line 4122 in `class.ticket.php` - 

```php
        // Not assigned...save optional note if any
        if (!$ticket->isAssigned() && $vars['note']) {
            if (!$cfg->isRichTextEnabled())
                $vars['note'] = new TextThreadEntryBody($vars['note']);
            $ticket->logNote(_S('New Ticket'), $vars['note'], $thisstaff, false);
        }

```

The issue I could see here is perhaps `$ticket->isAssigned()` is too general since it can include auto-assignment. If that's the case then we can chance it to `$vars['assignId']`

JediKev:
@protich 

>Did you actually replicate the issue as descried and truly tested your PR?

Yes and yes.

>Asking because the changes herein have nothing to do with the original issue as described by the poster

This is the only way I could replicate was to select an assignee and that didn't carry over the internal note.

>The issue I could see here is perhaps `$ticket->isAssigned()` is too general since it can include auto-assignment. If that's the case then we can chance it to `$vars['assignId']`

I knew I should've stuck with my original changes... That's exactly what I had before but since `!$ticket->isAssigned()` was added intentionally in [this](https://github.com/osTicket/osTicket/commit/ce81a4b293c011d7c6bbdb90256904ca65bb82ad) I didn't want to revert it so I worked around it. I will change this back to the original and update here.

Cheers.

protich:
> This is the only way I could replicate was to select an assignee and that didn't carry over the internal note.

I cannot see how that's even possible unless you didn't select assignee and you have auto-assignment set via dept, topic or filters. Agent selection takes precedent  - see logic starting at line `3980` in `class.ticket.php`

> was added intentionally in [this](https://github.com/osTicket/osTicket/commit/ce81a4b293c011d7c6bbdb90256904ca65bb82ad) I didn't want to revert it so I worked around it.

I think all changes are intentional but doesn't mean they're forever, that's why we have bug fixes.

JediKev:
@protich 

>I cannot see how that's even possible unless you didn't select assignee and you have auto-assignment set via dept, topic or filters

That's what I meant. Ticket was auto-assigned to Team via Help Topic and didn't carry over the note. Sorry, I had to look and remember.

Cheers.

JediKev:
@protich 

Updated 👍

Cheers.

Issue #4716, opened 2019-02-10
Multilanguage fields can't be saved - stuck on ajax 'Loading' on Save


pitzyn:
This is same for OST 1.10.x-1.11, PHP5.6/7.0/7.2, checked on two different hosting accounts

On Admin panel a few form fields are multilanguage and could be translated for each installed language.
Some of those forms can't be saved (regardles if any change has been done to its content nor default values been left unchanged).
The modal/Ajax message window is without css styling and it hungs forever on rotating gif and saying 'LOADING' 

It doesn't matter if only one/original language pack is installed or if additional packs are added.

This issue applies i.e. to tickets statuses (scp/lists.php?type=ticket-status) when you try to change its name, description or whatever value and then to save the form.
Same bug also applies to a few other places on admin/agent panel.

![2019-02-10](https://user-images.githubusercontent.com/47509322/52541264-4904bb00-2d93-11e9-985e-72747fcc4c1c.png)

Strange but for some places i.e. /scp/lists.php?type=ticket-status#properties tab to change & save its multilanguage content works fine.

ntozier:
I am unable to replicate this on my production sites.

What webserver are you running?
PHP errors?
console errors?

pitzyn:
How strange. I use Debian9/Apache2/PHP7.1 or PHP5.6/MariaDB.
I installed v1.10 and 1.11 and have the same issue with this multilanguage form fields - BUT with this particular one only. 
Any other multilang forms with editing & saving works normally.


I have found random osTicket demo installation, i.e. here (it is not related/hosted to me nor my system):
https://www.opensourcecms.com/osticket/

Please check this bug on their demo page. It does the same on editing ticket statuses /scp/lists.php?type=ticket-status

Admin Page: https://s1.demo.opensourcecms.com/osticket/scp/login.php
Username: opensourcecms
Password: opensourcecms

JediKev:
@pitzyn 

I still cannot reproduce this error with the latest codebase and translations installed/enabled. I'm able to change, update, and save any Ticket Status.

Cheers.

InToSSH:
I have exactly the same problem with my osTicket installation. Installed it yesterday from master. 
OS: Ubuntu 18.04 Server, PHP 7.1 (mostly stock config), MySQL 5.7.25. 
And same error: `VM1135:1 POST http://fqdn/scp/ajax.php/list/1/item/6/update 500 (Internal Server Error)`

JediKev:
[...]

PR #4758, opened 2019-02-26
issue: PHP 7.2 Ticket Status

JediKev:
This addresses issue #4716 where updating a Ticket Status throws a fatal error of "Too few arguments" when using PHP 7.2. This is due to `$errors` not being passed to the `update()` function causing PHP 7.2 to freak out. This creates an `$errors` array and passes it to `update()` so PHP 7.2 is happy and we are all happy. :)


kostek1:
I had the same problem, I checked works.

PR #4796, opened 2019-03-14
Issue: Ticket Export Headers

aydreeihn:
This commit fixes an issue where custom export headings were being set to numeric values rather than field labels.

For exports that are already saved, we can get the heading directly using getHeading. To get the heading for fields that are not already saved as exports, we can use getExportableFields, which will return path => heading.

### DIFF
--- a/include/class.queue.php
+++ b/include/class.queue.php
@@ -1144,25 +1144,23 @@ class CustomQueue extends VerySimpleModel {
 
         $new = $fields;
         foreach ($this->exports as $f) {
+            $heading = $f->getHeading();
             $key = $f->getPath();
             if (!isset($fields[$key])) {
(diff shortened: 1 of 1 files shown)


include/class.queue.php
```diff
@@ -1144,25 +1144,24 @@ function updateExports($fields, $save=true) {
...
-                $heading = $info;
-
             $f->set('heading', $heading);
             $f->set('sort', array_search($key, $order)+1);
             unset($new[$key]);
         }
 
         foreach ($new as $k => $field) {
-            if (is_array($field))
+            $exportableFields = CustomQueue::getExportableFields();
```
protich: Move this outside `foreach` loop.

aydreeihn: Fixed

PR #4804, opened 2019-03-18
Fix bug: send new ticket alert to account manager

hans-chen:
not sending new ticket alert to account manager in case:
"New ticket alert to Organization Account Manager" selected
"New message alert to Organization Account Manager" not selected


protich:
Thanks for the fix.

PR #4807, opened 2019-03-19
Active SLAs

nathankitchen:
This request includes updates to the way SLAs work and due dates are calculated. The main changes are:

- SLA "active hours" can be defined on a per weekday basis. This means that an SLA can apply 8-6 Mon-Fri, for example. The admin "Manage SLAs" screen is updated to capture the additional information.
 - When a new due date is calculated, the grace period is applied only within the designated weekday SLA active hours. For example, if an SLA is for 8:00-18:00 Mon-Fri and a new ticket arrives at 17:00 with a 4 hour grace period, the due date will be calculated as 11:00 on Monday. Previously this would have sent due date warnings at 21:00, when potentially nobody was available.
 - As part of the database migration to store the additional SLA weekday data, the SLA is set to 24/7, which results in an identical calculation of due dates until someone specifically changes these.
 - If for some bizarre reason someone sets every weekday to "no support", a very long due date will be added (150 days, I couldn't decide what an appropriate setting was).

Things it doesn't do, because we didn't need it to:
 - Define arbitrary support periods within a day, e.g. 9-12 and 13-18, for example giving no availability over lunch on a single day.
 - Operate on anything other than a per-weekday basis.

Things that need reviewing:
 - Use of input="time" element
 - Behaviour when every weekday is set to "no support"

ost-slas


lucize:
@nathankitchen 
do you think you can add something like sla1 -> sla2 -> sla3 (after each sla expires to be moved to another sla) ?
the idea is to have sla1 email overdue after x time, sla2 email overdue after y time and sla3 to email daily after z time

IT-Kat:
Is this tested with v1.12?? I'm excited for it!

IT-Kat:
> Is this working with v1.12?? I'm excited for it!

I have confirmed that this is working with v1.12 with one exception: it doesn't seem to realize that it's past due unless the ticket is opened sometime after the SLA expires. Any way to fix that?

nathankitchen:
Yes, you're right, we've spotted two bugs actually. One is that it's not calculating the SLA correctly if it is opened prior to the current day's start SLA, and ends within the window. Also, the code that marks a ticket as overdue doesn't use the same functions for calculating SLAs, so I'll need to patch that too.

We've got some code we're testing that should fix both, I'll issue a patch once we're happy with it.

protich:
@nathankitchen - thank you for the PR. Like we discussed on the phone - I personally preferred the Schedule approach and happy to say I now have a PR ( #4984) ready for some testing. 

Thanks and look forward to your feedback.

nathankitchen:
Closing to tidy up, as the Schedule approach released in 1.12 delivered this functionality.

PR #4808, opened 2019-03-20
Indicate browsers to show PDFs on another tab (as an inline attachment)

socratesit:


JediKev:
@socratesit 

Is this to open the PDF in a new tab to view it? Or to open in a new tab to download it? Or to show the PDF inline in the thread?

There is no pull description so someone who doesn't know how to read the code does not understand your intentions or what this even does.

Cheers.

socratesit:
@JediKev Apologies about the lack of the description on the pull request.

This a very simple patch. It just opens the PDF on the same tab as the current ticket (it will be viewed on the browser PDF reader), instead of saving it as an attachment.

 The user will have to hit "back" in order to return to the ticket once the document has been viewed.

dirigit:
Not a good solution IMHO. Ticket should remain on screen.

PR #4828, opened 2019-04-03
issue: sendAccessLink On NULL

JediKev:
This addresses an issue where entering a collaborator's email to send ticket email access link throws a fatal error. This is due to the method that checks for tickets with the User's email equal to the email provided. This only checks for User's emails not Collaborator emails. This adds a check for Collaborator emails as well so this will not crash out.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -2884,7 +2884,10 @@ implements RestrictedAccess, Threadable {
             ->filter(array('number' => $number));
 
         if ($email)
-            $query->filter(array('user__emails__address' => $email));
+            $query->filter(Q::any(array(
+                'user__emails__address' => $email,
+                'thread__collaborators__user__emails__address' => $email
+            )));
 
 
         if (!$ticket) {


include/class.ticket.php
```diff
@@ -1079,7 +1079,7 @@ function sendAccessLink($user) {
             'body' => $content->getLocalBody($lang),
         ), $vars);
 
-        $email->send($user, Format::striptags($msg['subj']),
+        $email->send($user->getEmail(), Format::striptags($msg['subj']),
```
protich: Any explanation for the change here? The type of user is important when creating message-id in `class.mailer.php`

JediKev: `$user` is always an object and `send()` needs the email address instead otherwise you run into the error `substr() expects parameter 1 to be string, object given`.

protich: I don't believe that's correct statement - see send function in class.mailer.php

JediKev: Okay, I see what you mean. `send()` is supposed to return the email address. I will have to look deeper into why that error occurs.

PR #4829, opened 2019-04-04
issue: sendAccessLink On NULL v1.11

JediKev:
This addresses an issue in v1.11 where sending an Access Link via email to a Collaborator fails with fatal error. This is due to the type of User class that is sent to the `Mailer::send()` function. The ClientSession User class is not appropriate for this method, rather, TicketOwner or Collaborator User classes are appropriate. This adds a function called `getSessionUser()` to the ClientSession class so we can retrieve the TicketOwner/Collaborator User class from the session.


JediKev:
This needs #4828 merged into `develop` then `develop` needs to be merged into `develop-next` first before merging this pull request to `develop-next`. 😁 

Cheers.

PR #4831, opened 2019-04-04
Issue: Queue with Teams in Criteria

aydreeihn:
This commit fixes an issue with queues or searches that have a criteria for team assignment when the Agent is not assigned to any teams. In this case, the queue should not return any tickets. Previously, it would create an invalid query which would cause system errors.

### DIFF
--- a/include/class.orm.php
+++ b/include/class.orm.php
@@ -2585,8 +2585,7 @@ class SqlCompiler {
         }
         $glue = $Q->ored ? ' OR ' : ' AND ';
         $clause = implode($glue, $filter);
-        if (($Q->negated || $parens) && count($filter) > 1)
-            $clause = '(' . $clause . ')';
+        $clause = '(' . $clause . ')';
         if ($Q->negated)
             $clause = 'NOT '.$clause;
         return new CompiledExpression($clause, $type);
--- a/include/class.orm.php
+++ b/include/class.orm.php
@@ -2560,8 +2560,10 @@ class SqlCompiler {
                         $criteria["{$field}__{$f}"] = $v;
                     }
                 }
-                $filter[] = $this->compileQ(new Q($criteria), $model,
-                    $Q->ored || $Q->negated);
+                // New criteria here is joined with AND, so if the outer
+                // criteria is joined with OR, then parentheses are
+                // necessary
+                $filter[] = $this->compileQ(new Q($criteria), $model, $Q->ored);
             }
             // Handle simple field = <value> constraints
             else {
@@ -2585,7 +2587,8 @@ class SqlCompiler {
         }
         $glue = $Q->ored ? ' OR ' : ' AND ';
         $clause = implode($glue, $filter);
-        $clause = '(' . $clause . ')';
+        if (($Q->negated || $parens) && count($filter) > 1)
+            $clause = '(' . $clause . ')';
         if ($Q->negated)
             $clause = 'NOT '.$clause;
         return new CompiledExpression($clause, $type);
@@ -2593,16 +2596,10 @@ class SqlCompiler {
 
     function compileConstraints($where, $model) {
         $constraints = array();
-        $prev = $parens = false;
         foreach ($where as $Q) {
-            if ($prev && !$prev->isCompatibleWith($Q)) {
-                $parens = true;
-                break;
-            }
-            $prev = $Q;
-        }
-        foreach ($where as $Q) {
-            $constraints[] = $this->compileQ($Q, $model, $parens);
+            // Constraints are joined by AND operators, so if they have
+            // internal OR operators, then they need to be parenthesized
+            $constraints[] = $this->compileQ($Q, $model, $Q->ored);
         }
         return $constraints;
     }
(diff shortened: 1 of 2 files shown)


aydreeihn:
@protich The only issue I see with this is in the WHERE clause of an administratively added queue with the condition that assignee includes one of my teams.

The criteria are already there in queue-tickets-tmpl.php before filtering the tickets for the agent's ticket visibility.

Custom Queue WHERE:
WHERE A2.`state` = 'open' AND A2.`state` = 'open' AND A1.`team_id` IS NULL AND ((A2.`state` = 'open' AND (A1.`staff_id` = 1 OR A5.`staff_id` = 1)) OR A1.`dept_id` IN (1, 2, 3, 4, 5, 6, 7, 9, 10, 11) OR A6.`id` IN (1, 2, 3, 4, 5, 6, 7, 9, 10, 11)) LIMIT 50

Admin Queue WHERE:
WHERE A1.`team_id` IS NULL AND (A2.`state` = 'open' AND (A1.`staff_id` = 1 OR A5.`staff_id` = 1)) OR A1.`dept_id` IN (1, 2, 3, 4, 5, 6, 7, 9, 10, 11) OR A6.`id` IN (1, 2, 3, 4, 5, 6, 7, 9, 10, 11) LIMIT 50

What the Admin Queue WHERE needs to be:
WHERE A1.`team_id` IS NULL AND ((A2.`state` = 'open' AND (A1.`staff_id` = 1 OR A5.`staff_id` = 1)) OR A1.`dept_id` IN (1, 2, 3, 4, 5, 6, 7, 9, 10, 11) OR A6.`id` IN (1, 2, 3, 4, 5, 6, 7, 9, 10, 11)) LIMIT 50

We just need that extra set of parenthesis.

include/class.search.php
```diff
@@ -1244,7 +1244,10 @@ function getSearchQ($method, $value, $name=false) {
                     $agents[] = (int) substr($id, 1);
                     break;
                 case 'T':
-                    $teams = array_merge($thisstaff->getTeams());
+                    if ($thisstaff && $staffTeams = $thisstaff->getTeams())
```
protich:
```php 

if ($thisstaff && ($staffTeams = $thisstaff->getTeams()))

```

include/class.search.php
```diff
@@ -1429,7 +1432,8 @@ function getSearchQ($method, $value, $name=false) {
                 && ($teams = $thisstaff->getTeams())) {
             unset($value['T']);
```
protich, line 1435: I believe the logic here is wrong here -  should be something like;

```php
        // Unpack my teams
        if (isset($value['T']) {
            if (!$thisstaff ||  !($teams = $thisstaff->getTeams()))
                return Q::any(['team_id' => null]);

	    unset($value['T']);
	    $value = $value + array_flip($teams);
	}
```

protich (changes requested):

include/class.orm.php
```diff
@@ -2585,8 +2585,7 @@ function compileQ(Q $Q, $model, $parens=true) {
         }
         $glue = $Q->ored ? ' OR ' : ' AND ';
         $clause = implode($glue, $filter);
-        if (($Q->negated || $parens) && count($filter) > 1)
-            $clause = '(' . $clause . ')';
+        $clause = '(' . $clause . ')';
```
protich: Hey @greezybacon - do you foresee any issues with this change? See the PR conversation for more details.

Thanks,

greezybacon: I would suggest this patch for the ORM. There is definitely a bug:
```diff
diff --git a/include/class.orm.php b/include/class.orm.php
index 132986a4..f7bbfac4 100644
--- a/include/class.orm.php
+++ b/include/class.orm.php
@@ -2614,8 +2614,10 @@ class SqlCompiler {
                         $criteria["{$field}__{$f}"] = $v;
                     }
                 }
-                $filter[] = $this->compileQ(new Q($criteria), $model,
-                    $Q->ored || $Q->negated);
+                // New criteria here is joined with AND, so if the outer
+                // criteria is joined with OR, then parentheses are
+                // necessary
+                $filter[] = $this->compileQ(new Q($criteria), $model, $Q->ored);
             }
             // Handle simple field = <value> constraints
             else {
@@ -2649,16 +2651,10 @@ class SqlCompiler {
 
     function compileConstraints($where, $model) {
         $constraints = array();
-        $prev = $parens = false;
         foreach ($where as $Q) {
-            if ($prev && !$prev->isCompatibleWith($Q)) {
-                $parens = true;
-                break;
-            }
-            $prev = $Q;
-        }
-        foreach ($where as $Q) {
-            $constraints[] = $this->compileQ($Q, $model, $parens);
+            // Constraints are joined by AND operators, so if they have
+            // internal OR operators, then they need to be parenthesized
+            $constraints[] = $this->compileQ($Q, $model, $Q->ored);
         }
         return $constraints;
     }
```
Ultimately, the logic in `compileConstraints` cannot properly determine if parentheses are necessary or not.

And, my apologies if the line numbers do not match up well. This edit was made on the `canned-tasks` branch which has another ORM patch for cleaning up the expression API.

protich (approved):

PR #4839, opened 2019-04-10
Update README.md

ntozier:


README.md
```diff
@@ -22,7 +22,7 @@ easy to setup and use. The best part is, it's completely free.
 Requirements
 ------------
   * HTTP server running Microsoft® IIS or Apache
-  * PHP version 5.4 or greater, 5.6 is recommended
+  * PHP version 5.6 to 7.2, 5.6 is recommended
```
protich: I have a pull request I'll be making to fix issues with 7.3 and we absolutely should recommend 7.2. This should be `PHP version 5.6 to 7.3, 7.2 is recommended`

ntozier: Until the 7.3 pulls are in, and confirmed i highly recommend that we not mention 7.3 at all.
Otherwise I completely agree it's time to start recommending 7.2.

PR #4841, opened 2019-04-10
feature: ACL (Access Control List)

JediKev:
This adds a new feature called ACL that offers the ability to control what IP addresses are allowed to access the Login Page(s). This adds a new textfield to the System Settings to add a comma separated list of IPs. This also adds a checkbox called "Apply To Backend Only" that, if enabled, will only apply the ACL to the Agent Login Page. Lastly, this adds a validator for a simple comma-separated list of IP addresses. (eg. `111.111.111.111, 222.222.222.222, 333.333.333.333`)

If the requester's IP is not in the ACL the system will show an "Access Denied" page. If the requester's IP is in the ACL they will be able to access the Login Page(s) as usual. If the ACL field is left blank anyone will be able to access the User/Agent Login Pages.

### DIFF
--- a/include/class.config.php
+++ b/include/class.config.php
@@ -424,6 +424,26 @@ class OsticketConfig extends Config {
         return str_replace(array(', ', ','), array(' ', ' '), $this->get('allow_iframes')) ?: 'self';
     }
 
+    function getACL() {
+        if (!($acl = $this->get('acl')))
+            return null;
+
+        return explode(',', str_replace(' ', '', $acl));
+    }
+
+    function getACLBackendOpts() {
+        return array(
+            0 => __('Disabled'),
+            1 => __('All'),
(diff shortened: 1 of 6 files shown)


ntozier:
I would recommend that you institute ranges and not just single IPs.

Example:
If I wanted 192.168.44.1 to 192.168.45.255 you could enter it either as:

192.168.44.1-192.168.45.255
or as
192.168.44.0/24,192.168.45.0/24

JediKev:
@ntozier

That will be implemented in version 2 of this feature. Version 1 is basic.

Cheers.

ntozier:
@JediKev fair enough.  I just can't use this feature until it exists.... 19 locations spread across the state. :)

A separate lists for Users and SCP would also be a welcome edition.

PR #4850, opened 2019-04-16
Issues since v1.11 release

protich:
Bug fixes and enhancements for v1.11

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -2672,6 +2672,27 @@ class ThreadEntryField extends FormField {
 }
 
 class PriorityField extends ChoiceField {
+
+    var $priorities;
+    var $_choices;
+
+    function getPriorities() {
+        if (!isset($this->priorities))
+            $this->priorities = Priority::objects();
+
+        return $this->priorities;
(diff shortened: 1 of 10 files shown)


include/class.forms.php
```diff
@@ -2198,6 +2198,9 @@ function toString($value) {
         if (is_array($value))
             return '';
 
+        if (is_array($value))
+            return '';
+
```
Chefkeks: Despite that I don't have any idea where this code is used, but does it make sense to `return '';` twice?

protich: Most definitely unnecessary - fixed. Thanks

aydreeihn:
I see one more continue 2 warning:

PHP Warning:  "continue" targeting switch is equivalent to "break". Did you mean to use "continue 2"? in /var/www/html/develop-next/osTicket/setup/test/tests/class.php_analyze.php on line 78

Should this one be changed as well or was it intentionally left out?

protich:
@aydreeihn - good catch, I'll add it to the PR shortly.

PR #4867, opened 2019-04-23
Issue: Tickets Visibility

aydreeihn:
This commit ensures that an Agent with the 'See all tickets in search results, regardless of access' permission checked can see all Tickets in a search or a saved search, but does NOT see all Tickets within queues.

### DIFF
--- a/include/class.search.php
+++ b/include/class.search.php
@@ -900,7 +900,7 @@ class SavedQueue extends CustomQueue {
        $counts = array();
         $query = Ticket::objects();
         // Apply tickets visibility for the agent
-        $query = $agent->applyVisibility($query);
+        $query = $agent->applyVisibility($query, true);
         // Aggregate constraints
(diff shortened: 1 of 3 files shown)


aydreeihn:
@protich I may need some extra info on the patch you sent me for the exclude_archived bit. For now, I just added the diff to my code.
Your comment for reference:
"add below to your Visibility commit but make excluding archived true all the time.  We also need ability to exclude archived when visibility is being ignored."

Also, I wanted to explain why I ended up writing the visibility bit this way rather than changing up some of the queue functions (isASubQueue, isAQueue, isASearch, isASavedSearch). I noticed that the fields we get for 'Add Other Fields' in the queue/search edit counts on queues being only those we ship or ones created Administratively. We only allow Answered, Assigned, and Overdue for those queues (Supplemental Filters) whereas other queues and searches allow you to filter by all of the fields. Let me know what you think on this.

PR #4879, opened 2019-05-02
Issue: Reopen Assignment

aydreeihn:
This commit fixes an issue where Tickets were not being assigned to the last respondent if ‘Claim on Response’ was disabled and a Ticket was reopened.


aydreeihn:
@protich Shouldn't tickets still reopen to the last respondent regardless of if 'Claim on Response' is disabled for Tickets and Departments? Seems like they just shouldn't auto assign if an agent replies to an unassigned ticket. 

If this isn't the case, we should probably update this tool tip:
Screen Shot 2019-05-02 at 10 32 55 AM

aydreeihn:
@protich Updates:

Claim on response tool tip:
Screen Shot 2019-05-09 at 3 56 35 PM

Reopen Auto Assignment option:
Screen Shot 2019-05-09 at 3 59 32 PM

PR #4888, opened 2019-05-09
issue: API Validate Request Structure

JediKev:
This addresses an issue where validating the request structure for an API call throws warnings when field values are arrays. This adds a new variable called `$level` to keep track of the level of the request structure we are in. Once we reach a certain level we will break and continue with the other fields. This will prevent warnings when a value is an array like `["cc"] => { [0] => "email2", [1] => "email1"}`.

### DIFF
--- a/include/class.api.php
+++ b/include/class.api.php
@@ -244,14 +244,17 @@ class ApiController {
      * expected. It is assumed that the functions actually implementing the
      * API will further validate the contents of the request
      */
-    function validateRequestStructure($data, $structure, $prefix="", $strict=true) {
+    function validateRequestStructure($data, $structure, $prefix="", $strict=true, $level=0) {
         global $ost;
 
         foreach ($data as $key=>$info) {
+            if($level == 6)
+                break;
             if (is_array($structure) && (is_array($info) || $info instanceof ArrayAccess)) {
+                $level++;
                 $search = (isset($structure[$key]) && !is_numeric($key)) ? $key : "*";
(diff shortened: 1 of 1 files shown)


include/class.api.php
```diff
@@ -244,14 +244,17 @@ function getRequestStructure($format, $data=null) { return array(); }
...
 
         foreach ($data as $key=>$info) {
+            if($level == 6)
+                break;
             if (is_array($structure) && (is_array($info) || $info instanceof ArrayAccess)) {
+                $level++;
                 $search = (isset($structure[$key]) && !is_numeric($key)) ? $key : "*";
                 if (isset($structure[$search])) {
-                    $this->validateRequestStructure($info, $structure[$search], "$prefix$key/", $strict);
+                    $this->validateRequestStructure($info, $structure[$search], "$prefix$key/", $strict, $level);
```
greezybacon: You should probable do `$level + 1` in the nested call so that `$level` is not corrupted inside the `foreach` loop. Otherwise, the loop will validate combinations of six items in a loop, including nesting.

JediKev: Thank you, will look into this.

ericLemanissier:
I just got the error message `API Unexpected Data thread_entry_recipients/cc/1: Unexpected data received in API request` (running v1.15.1). This feels related to this PR.  Is there some progress on this topic ?

EDIT: the corresponding message is actually visible in osticket

PR #4894, opened 2019-05-15
Reformat Incorrect Reply-To Headers

aydreeihn:
In the event that we receive an email where the reply-to header is formatted with the name being an unquoted email, we should correct the name by inserting the quotes.

Incorrect Format:
adriane@enhancesoft.com <adriane@enhancesoft.com>

Correct Format:
"adriane@enhancesoft.com" <adriane@enhancesoft.com>

### DIFF
--- a/include/class.mailfetch.php
+++ b/include/class.mailfetch.php
@@ -671,6 +671,18 @@ class MailFetcher {
         // attachment. Download the body and pass it along to the mail
         // parsing engine.
         $info = Mail_Parse::splitHeaders($mailinfo['header']);
+
+        //make sure reply-to headers are correctly formatted
+        if ($mailinfo['reply-to'] && !Validator::is_email($mailinfo['reply-to']) && $info['Reply-To']) {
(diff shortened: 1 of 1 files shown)


include/class.mailfetch.php
```diff
@@ -671,6 +671,16 @@ function createTicket($mid) {
...
         // parsing engine.
         $info = Mail_Parse::splitHeaders($mailinfo['header']);
+
+        //make sure reply-to headers are correctly formatted
+        if ($mailinfo['reply-to'] && !Validator::is_email($mailinfo['reply-to']) && $info['Reply-To']) {
+            $replyto = Mail_Parse::parseAddressList($info['Reply-To']);
+            if ($replyto[0]) {
+                $mailinfo['reply-to'] = sprintf('%s@%s', $replyto[0]->mailbox, $replyto[0]->host);
+                $mailinfo['reply-to-name'] = $replyto[0]->personal;
+            }
```
protich: add below as a failsafe.
```php
else { 
  $mailinfo['reply-to'] = null;
} 
```

PR #4895, opened 2019-05-15
Issue: Ticket Alerts vs Dept Recipients

aydreeihn:
If an alert is enabled for only the Department Manager and the Department Recipients field is set to No one, the Department Manager still receives a notification.

If the recipients field is set to no one, nobody should receive an alert regardless of what is checked on the New Ticket Alert selections.

To fix this, we can do a count of what the getMembersForAlerts function returns. This function compares what is set for a Department's group_membership in the database to the ALERTS_DISABLED constant to return a query set of members that should receive an alert. If the count is 0, we know that alerts should be disabled.

### DIFF
--- a/include/class.dept.php
+++ b/include/class.dept.php
@@ -326,6 +326,10 @@ implements TemplateVariable, Searchable {
         return $rv;
     }
 
+    function getNumMembersForAlerts() {
+        return count($this->getMembersForAlerts());
+    }
+
     function getSLAId() {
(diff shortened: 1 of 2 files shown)


protich:
I don't like the sprinkles of `counts` here - why not add a routine `$dept->getNumMembersForAlerts()` and return the count. This is cleaner, in my opinion.

aydreeihn:
Updated

PR #4900, opened 2019-05-21
xss: Overridetype Broken Images

JediKev:
This addresses a security issue reported by Vincent Monier (Xenos) where someone can upload a broken image containing arbitrary code, add an `overridetype` to the download URL, send the link to anyone, and if clicked will display the image as HTML instead of a broken image ergo executing the code. This is due to an old if statement that set the file type to the `overridetype` if one is provided. We don’t use this if statement anymore so we are going to retire it. This will prevent anyone from faking/overriding the file type which will prevent XSS in broken images. In addition, this adds a new check with `exif_imagetype()` to ensure the file contents match the expected file type.

### DIFF
--- a/include/class.file.php
+++ b/include/class.file.php
@@ -247,8 +247,6 @@ class AttachmentFile extends VerySimpleModel {
         $ttl = ($expires) ? $expires - Misc::gmtime() : false;
         $this->makeCacheable($ttl);
         $type = $this->getType() ?: 'application/octet-stream';
-        if (isset($_REQUEST['overridetype']))
-            $type = $_REQUEST['overridetype'];
         Http::download($this->getName(), $type, null, 'inline');
         header('Content-Length: '.$this->getSize());
         $this->sendData(false);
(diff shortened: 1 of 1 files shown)


include/class.file.php
```diff
@@ -358,6 +356,10 @@ static function uploadLogo($file, &$error, $aspect_ratio=2) {
     }
 
     static function create(&$file, $ft='T', $deduplicate=true) {
+        // Crap-out if file contents do not match expected file type
+        if (strpos('image/', $ft) !== 0 && !exif_imagetype($file['tmp_name']))
```
protich: `$ft` is wrong where - it should be `$file['type']`. And on the second thought - this check should go to FileUploadField.

include/class.file.php
```diff
@@ -358,6 +356,10 @@ static function uploadLogo($file, &$error, $aspect_ratio=2) {
     }
 
     static function create(&$file, $ft='T', $deduplicate=true) {
+        // Crap-out if file contents do not match expected file type
+        if (strpos('image/', $file['type']) !== 0 && !exif_imagetype($file['tmp_name']))
```
protich: This will crap out when an image is emailed - liked I mentioned in this thread, the logic  NEEDS to go to FileUploadField. I'll make a PR once I have it tested.

JediKev: My mistake, I apologize, I don't know how I missed that part of the comment. Must've had a lot going on that day I guess.

Issue #4909, opened 2019-05-31
HTML tags are visible in the secondary language (help topic/instructions)


slicki:
### Prerequisites

* [y] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [n] Do you have any errors in the PHP error log, or javascript console?
* [y] Did you check the [osTicket forums](http://osticket.com/forum/categories)?
* [y] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

### Description

When the customer changes the language, html-tags are shown in the instrutions of the help-topic (default language everything is fine). 

### Steps to Reproduce

1. [First Step]
**Create new help topic and write instructions in german and english - including breaks and bold text**
![ht_ger](https://user-images.githubusercontent.com/11290633/58691891-98e0af80-838d-11e9-9c4a-5eb30364dba9.png)
![ht_us](https://user-images.githubusercontent.com/11290633/58691895-9b430980-838d-11e9-83a5-24d401aaabc8.png)

2. [Second Step]

**Login as customer and choose the help topic when opening a new ticket**

![result_ger](https://user-images.githubusercontent.com/11290633/58692558-22dd4800-838f-11e9-9c13-c78bb700f77f.png)

![result_us](https://user-images.githubusercontent.com/11290633/58692690-6b950100-838f-11e9-9fae-9714ffdc0d63.png)

**Expected behavior:**  HTML-tags not shown as strings
**Actual behavior:** HTML-tags are visible in the second language

### Versions

osticket-Version:  v1.12 (a076918)

OS: debian-stretch (9.9)
PHP:  7.0

JediKev:
@slicki 

Here you go:
https://github.com/osTicket/osTicket/pull/4910

Cheers.

slicki:
@JediKev 

Works like a charm - thanks for the quick response!

Have a nice weekend
slicki

Chefkeks:
@JediKev Great work! 👍 

@slicki 
As I told you... JediKev will take care if you open an issue 😁

JediKev:
@Chefkeks  Ah nice to see/hear from you again! Thank you for the kind words as always! :)

Chefkeks:
@JediKev 
[...]

PR #4910, opened 2019-05-31
issue: Form Instructions Translation

JediKev:
This addresses issue #4909 where setting a translated Form Instruction fails to render properly on the client side. This is due to us formatting the instructions but not decoding before we render. This adds the decode method to the translated Form Instructions so they are rendered properly on the client side.


JediKev:
Don’t merge yet, need to move to `getInstructions()` instead. No need to decode everything that’s translated; only the instructions.

PR #4913, opened 2019-06-03
ticket-view: Add ZIP export option

greezybacon:
This adds the ability to dowload a ticket thread PDF in a ZIP file with all of the attachments included. And, optionally, also include the associated tasks, again with a PDF of the thread and all attachments included.

It requires the PHP `zip` extension. So a new line item was added to the system information page, and to prevent errors, the option does not show up on the ticket view page if the extension is not available.

It would be possible to also include this as a feature on the ticket listing page, but exporting lots of tickets could easily exceed the max_execution_time and create confusion. So I left that out for this request.

This could be mocked up as a plugin; however, it doesn't look like the drop-downs on the ticket view page currently support plugins?

### DIFF
--- a/include/class.export.php
+++ b/include/class.export.php
@@ -555,3 +555,103 @@ class DatabaseExporter {
         $this->write_block(array('end-table'));
     }
 }
+
+class TicketZipExporter {
+    var $ticket;
+    var $tmpfiles;
+
+    function __construct(Ticket $ticket) {
+        $this->ticket = $ticket;
+        $this->tmpfiles = array();
+    }
+
+    function addTicket($ticket, $zip, $prefix, $notes=true, $psize=null) {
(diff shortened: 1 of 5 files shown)


JediKev:
Nicceeee. Will have to test this out!

protich:
PR requires rebasing on upstream/develop-next to resolve the conflicts.

philbertphotos:
adding this ... thanks so much .

mind-drive:
> Nicceeee. Will have to test this out!
what happen to your testing?

JediKev:
@mind-drive 

Please look right above your comment; we rewrote and merged this way back in 2019 with [this pull request](https://github.com/osTicket/osTicket/pull/5159).

Cheers.

mind-drive:
Thank you so much.

PR #4926, opened 2019-06-19
issue: Single Quotes In DB Password

JediKev:
This addresses an issue reported on twitter where single quotes in a db password causes 500 Error after installation. This is due to the system not escaping the single quotes before writing to the config file. This adds `addcslashes()` to the password before it’s saved to the config file so it escapes any single quotes properly.

### DIFF
--- a/setup/inc/class.installer.php
+++ b/setup/inc/class.installer.php
@@ -272,7 +272,7 @@ class Installer extends SetupWizard {
             '%CONFIG-DBHOST' => $vars['dbhost'],
             '%CONFIG-DBNAME' => $vars['dbname'],
             '%CONFIG-DBUSER' => $vars['dbuser'],
-            '%CONFIG-DBPASS' => $vars['dbpass'],
+            '%CONFIG-DBPASS' => addcslashes($vars['dbpass'], "'"),
             '%CONFIG-PREFIX' => $vars['prefix'],
             '%CONFIG-SIRI' => Misc::randCode(32),
         ));


protich:
Doesn't this mean you need to strip slashes ( `stripcslashes`) to use it?  Which brings up another issue - what happens if you have a slash on the password.

protich:
I almost want to say let's simply hash like email passwords - but it's slow operation by design, that would need to be done on every page load. Just thoughts...

JediKev:
PHP renders a string like `echo 'pass\'word';` as `pass'word` by default _BUT_ you're right in that if you have `pass\'word` as your actual password in mysql, the system writes it to the config file as `define('DBPASS','pass\\'word');` which totally messes things up. Will have to think deeper on this one.

Cheers.

greezybacon:
What if you just hex-encoded or base64-encoded it?

greezybacon:
```php
define('DBPASS', base64_decode('cGFzc1wnd29yZA=='));
```

JediKev:
@greezybacon That could totally be an option. Makes it more secure and we don’t have to worry about escaping anything. 

@protich What are your thoughts?

greezybacon:
But that said--and sorry to keep bombing this issue--`var_export` is used to represent a value in a way that can be re-interpreted by PHP as the original value:

if `$qq` is the literal `pass\'word`, then
```
php> var_export($qq);
'pass\\\'word'
```

Or more specifically:
```php
echo("DEFINE('DBPASS', ".var_export($qq, true).");");
```
will emit
```
DEFINE('DBPASS', 'pass\\\'word');
```

PR #4930, opened 2019-06-28
Custom Form Filed in Ticket page to available for Export in CSV

apyash:
Nothing


rmiranda16:
Hello @apyash  i try this, but not is available Custom Form Filed for Export in CSV.

Issue #1068, opened 2014-06-27
Merging Tickets (Same user but different email accounts)


C4RLOCO:
Hi folks,
I have this problem: with the addition of Users and Organization in v1.9.2, we have a myriad of tickets that kind of belongs to the same users but they are spread in different accounts. Some of this accounts are result of email typos, etc.

I would like to clean up the user accounts a little bit and i could do this by merging all these tickets into the appropriate account. However, I can't find the way to effectively do this without having to go ticket by ticket and manually change the ownership.

Any thoughts?

szemy:
+1 on virtual ticket. That looks like the best option as there are many issues with just merging (ticket properties, timeline ...)

simonnzg:
I note this subject has been closed. However, the duplicate issue that prompted the closure has also been closed!!!

szemy:
This one is still open: https://github.com/osTicket/osTicket-1.8/issues/1225

simonnzg:
On 26/01/2015 12:47, szemy wrote:
> This one is still open: #1225
> <https://github.com/osTicket/osTicket-1.8/issues/1225>
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/osTicket/osTicket-1.8/issues/1068#issuecomment-71455612>.
>


That's OK then!  ;-)

Simon

richbodo:
This should be MERGED with [1225] (https://github.com/osTicket/osTicket-1.8/issues/1225) ;)

PR #4939, opened 2019-07-04
union of several cases

ParmaU:
Hello
sometimes I get a response to a case and the system creates a new ticket,
in these cases it would be nice to be able to combine the two cases under one


ntozier:
Did you mean to make a pull request?

This seems to be a duplicate of https://github.com/osTicket/osTicket/issues/1068

PR #4948, opened 2019-07-10
security: CSV Formula Injection

JediKev:
This addresses a security issue discovered by Aishwarya Iyer where a User can change their Full Name to a windows formula and when an Agent exports a list of Users containing said User and opens the export file, the formula will be executed on their computer (if it's windows of course). This adds a new validator called `is_formula()` to all text fields disallowing the use of the following characters `= + - @` at the beginning of text. This should mitigate CSV Formula injections for any text field that allows user-input in the system. To further prevent CSV Formula injections this adds an escape mechanism to the Exporter that will escape any content matching the formula regex with a single quote (as mentioned in many posts about this subject).

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -1300,7 +1300,8 @@ class TextboxField extends FormField {
         parent::validateEntry($value);
         $config = $this->getConfiguration();
         $validators = array(
-            '' =>       null,
+            '' =>       array(array('Validator', 'is_formula'),
+                __('Content cannot start with the following characters: = - + @')),
             'email' =>  array(array('Validator', 'is_valid_email'),
                 __('Enter a valid email address')),
             'phone' =>  array(array('Validator', 'is_phone'),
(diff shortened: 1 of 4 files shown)


include/class.forms.php
```diff
@@ -1379,6 +1380,29 @@ function getConfigurationOptions() {
...
 
+    function validateEntry($value) {
+        parent::validateEntry($value);
+        $config = $this->getConfiguration();
+        $validators = array(
+            '' =>       array(array('Validator', 'is_formula'),
+                __('Content cannot start with the following characters: = - + @')),
+        );
+        // Support configuration forms, as well as GUI-based form fields
+        $valid = $this->get('validator');
```
protich: I would say consolidate this block of code as shown below.
```php        
if (!($valid = $this->get('validator')) && isset($config['validator']))
    $valid = $config['validator'];
```

include/class.forms.php
```diff
@@ -1379,6 +1380,29 @@ function getConfigurationOptions() {
         );
     }
 
+    function validateEntry($value) {
+        parent::validateEntry($value);
+        $config = $this->getConfiguration();
```
```php 
if (!$value) return;
```

I would say return early of value is empty - why get configuration etc only to break when `$value` is empty in line 1395

JediKev: Done 👍

PR #4954, opened 2019-07-17
issue: Advanced Search Default Sorting

JediKev:
This addresses an issue where conducting an Advanced Search will show the results in ASC order (from oldest to most recent) by default. The search is supposed to show the results in DESC order (from most recent to oldest) by default. This adds default sorting to Advanced Searches so that the results appear correctly again. This also adds a check for Adhoc Searches where if queue is Adhoc Search the sort is not saved to the session (this avoids `DoesNotExist` errors on `osTicketSession::__wakeup()`).


aydreeihn:
Looks good to me! Tests check out as well

protich:
Overall this looks good but what happens when a search has a keyword - I believe the system sorts the results based on `relevance` (which it should - given the expanse of possible matches) and limits it to 500 or so. Just a thought as something to test and make sure the changes don't override it.

JediKev:
After some testing this seems to be deeper than just adding a `getDefaultSort`. This appears to stem from a `GROUP BY A1.ticket_id` clause in the SQL statement. Looking further into this.

include/staff/templates/queue-tickets.tmpl.php
```diff
@@ -90,8 +90,10 @@
     $criteria->annotations = $criteria->related = $criteria->aggregated =
         $criteria->annotations = $criteria->ordering = [];
     $tickets->constraints = $tickets->extra = [];
+    $criteria->extra(array('select' => array('relevance' => 'Z1.relevance')));
     $tickets = $tickets->filter(['ticket_id__in' =>
             $criteria->values_flat('ticket_id')]);
+    $tickets->order_by(new SqlCode('A2.relevance'), QuerySet::DESC);
```
protich: Table aliases are auto-assigned via ORM - while they're predictable based on table relationships (joins), it's not hard-coded. So this makes me a little concerned. I don't have a solution - just raising the concern here.

JediKev:
@protich Updated

PR #4955, opened 2019-07-17
Add/Remove Collabs Without Refresh

aydreeihn:
This code allows Agents to Manage Collaborators without the page refreshing so that they will not lose any drafts that might not have been saved yet. Instead, adding/removing collaborators will show up instantly in the active collaborators section.

This code also removes the http response from the updateCollaborators function. This triggers collaborators.tmpl.php to correctly update the active collaborators section without needing to refresh the page.

### DIFF
--- a/include/ajax.thread.php
+++ b/include/ajax.thread.php
@@ -222,20 +222,7 @@ class ThreadAjaxAPI extends AjaxController {
             Http::response(404, 'No such thread');
 
         $errors = $info = array();
-        if ($thread->updateCollaborators($_POST, $errors))
-            $users = array();
-            foreach ($thread->getCollaborators() as $c)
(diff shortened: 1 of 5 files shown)


aydreeihn:
Apparently I already wrote this a long time ago? Who knew lol!

JediKev:
So far in my testing this looks great. Tests check out, no errors in logs, works as expected. 👍

ntozier:
> This code allows **users** to Manage Collaborators

Should that say "users"?  I would think is should say "Agents". :) (but haven't looked at the code or tested it)

JediKev:
@ntozier She meant Agents. It works as expected.

ntozier:
Right but if the commit note stays as it is, anyone reading it might be confused later.

PR #4968, opened 2019-07-29
issue: Set Staff Password On Creation

JediKev:
This addresses an issue reported on the Forum where setting a password for an Agent upon creation doesn't save the password in the backend. This means if you set a password, click Create, and then try to login as the agent you will be denied access (as there is no password saved). We store the PasswordResetForm data in the session for later use but we do not actually use the form data anywhere in the `Staff::update()` method. This updates the `Staff::update()` method to include a check for the form data and if exists sets the password and saves it in the backend. This maintains current functionality where if `Send the agent a password reset email` is enabled then we do not set the password, instead we send the Agent a registration email.

In addition, this updates the `StaffAjaxAPI::setPassword()` to include password validation so if a password is not valid we do not save it. Instead, we will keep the modal open and throw validation errors until a valid password is given. This means the form data will not be stored in the session until we have a valid password.

### DIFF
--- a/include/ajax.staff.php
+++ b/include/ajax.staff.php
@@ -35,12 +35,14 @@ class StaffAjaxAPI extends AjaxController {
 
       if ($_POST && $form->isValid()) {
           $clean = $form->getClean();
-          if ($id == 0) {
-              // Stash in the session later when creating the user
-              $_SESSION['new-agent-passwd'] = $clean;
-              Http::response(201, 'Carry on');
-          }
           try {
+              // Validate password
+              PasswordPolicy::checkPassword($clean['passwd1']);
+              if ($id == 0) {
+                  // Stash in the session later when creating the user
+                  $_SESSION['new-agent-passwd'] = $clean;
+                  Http::response(201, 'Carry on');
+              }
               if ($clean['welcome_email']) {
                   $staff->sendResetEmail();
               }
--- a/include/class.staff.php
+++ b/include/class.staff.php
@@ -1064,6 +1064,12 @@ implements AuthenticatedUser, EmailContact, TemplateVariable {
         $this->mobile = Format::phone($vars['mobile']);
         $this->notes = Format::sanitize($vars['notes']);
 
+        // Set staff password if exists
+        if (!$vars['welcome_email'] && $vars['passwd1']) {
+            $this->setPassword($vars['passwd1'], null);
+            $this->change_passwd = $vars['change_passwd'] ? 1 : 0;
+        }
+
         if ($errors)
             return false;
 


stfrbrntyu:
Hi
I am wondering, was this merged into RC2 ? (i see RC2 label in the commit)
I am having issues with the staff password when I create a new staff member, it creates a NULL value in the database, doesn't send a mail and when I try to send the agent a reset e-mail nothing happens.
I can set the password manually.

I see ajax.staff.php is the same as in the commit, but class.staff.php looks totally different. I started clean with the RC2 zip file.

1.14-git (4b9a699)
MySQLversie | 5.5.64
PHPversie | 7.2.10

disclaimer: I am not a coder, nor really familliar with github ... 

edit:
I discovered that creating an agent from the Departments page, via Access, there is a button at the bottom where you can add agents, from here it works fine, the mail is sent to the agent with a reset link.

gstrub:
Hi jedikev and stfrbrntyu,

I encountered the same issue as stfrbrntyu (no email being sent when creating an user or resetting a password from the admin interface) on version 1.14.1 as well as on 1.12.5. Both are being listed in the affected versions of commit d9108b.

When using the "Set Password" function for a new (or existing user), clicking the "Update" button when the "Send the agent a password reset email" checkbox is ticked leads to the following:
- The "Loading..." spinner overlay flashes briefly
- The "Set Password" popup box stays open on the same page
- Unchecking the "Send the agent a password reset email" box shows the password entry fields (as expected) but also the error message "Password must be at least 6 characters"

If I set a dummy 6-characters password such as "aaaaaa", then check "Send the agent a password reset email" (the password fields are then hidden) and validate, the email is sent as expected.

In ajax.staff.php, if I move line 40 (calling PasswordPolicy::checkPassword) to line 50 (else-block when not using the reset password function), or if I just comment it, everything works as expected.
The issue is that PasswordPolicy::checkPassword($clean['passwd1']) throws an exception (since passwd1 is empty when using the reset function) before we could send the reset password email

JediKev:
@gstrub 

Have you tried this patch yet?
- https://github.com/osTicket/osTicket/pull/5308

Cheers.

gstrub:
Hi @JediKev,

Patch #5308 indeed solves the issue I had (no email sent on agent creation).

However, without changing `include/ajax.staff.php`, the reset password function (for existing agents) from the admin interface still does not work due to the `PasswordPolicy` being checked before evaluating `welcome_email `([line 40 in `include/ajax.staff.php`](https://github.com/JediKev/osTicket/blob/d9108b1102ddab11caf5f62f66e4846d3a0fdec9/include/ajax.staff.php#L40))

Best regards,
Guillaume

JediKev:
@gstrub 

You were close. This isn't because it's checked _before_ `welcome_email`, it's because we are checking it _at all_ when we have `welcome_email` checked lol

There is no need to check the `$clean['passwd1']` variable as we are sending reset email and there should be no password being set. Here is the patch for that part of the issue:
- https://github.com/osTicket/osTicket/pull/5319

Thanks for the report. Cheers.

stfrbrntyu:
Installed both patches and it's working now. thanks alot.

PR #4970, opened 2019-07-30
issue: Image Attachment View

JediKev:
This enhances security patch (33ed10) to allow image attachments to be opened in a new tab if the setting is enabled. This adds a new setting to the Agent Profile labeled "Image Attachment View" that determines how image attachments are handled. The two options are "Download" and "New Tab" with "Download" being the default. If the option is set to "Download" all image attachments will be forced to download. If the option is set to "New Tab" all image attachments will open in a new tab when you right click + open in tab.


include/class.file.php
```diff
@@ -240,7 +240,9 @@ static function _genUrlSignature($id, $key, $signature, $expires) {
     }
 
     function download($disposition=false, $expires=false) {
-        $disposition = ($disposition && strcasecmp($disposition, 'inline') == 0
+        $thisstaff = StaffAuthenticationBackend::getUser();
+        $disposition = ((($disposition && strcasecmp($disposition, 'inline') == 0)
+              || $thisstaff->getImageAttachmentView() === 'New-Tab')
```
protich: Wouldn't this crash when you don't have an authenticated staff - e.g file view from Client Portal?

include/class.staff.php
```diff
@@ -88,6 +88,7 @@ function getConfig() {
                         'default_from_name' => '',
                         'datetime_format'   => '',
                         'thread_view_order' => '',
+                        'img_att_view' => 'Download',
```
protich: I think it should default to `New-Tab` - which by the way is a weird choice of word. It should be `inline` in my humble opinion.

include/class.file.php
JediKev: lol it did, updated now

include/class.staff.php
JediKev: I disagree with the defaulting to New Tab...we implemented Download by default so that by default Agent's aren't susceptible to malicious/malformed images. If Agents want to be susceptible to possibly tainted images then they can change it to New Tab (or `Inline` as it's called now).

I have updated the wording to `Inline` instead. I agree that this is more understandable and much better wording.

PR #4971, opened 2019-07-30
Fix Admin Alert

aydreeihn:
This commit fixes an issue where alerts are not sent out to the Admin Email if a Department does not have any primary or extended members despite Admin Email being enabled for the New Ticket Alert.

Now, if the Admin Email is checked for the New Ticket Alert and a Department does not have any members at all, an Alert will still go out to the Admin unless the Department recipients field is set to 'No one (diable Alerts and Notices)'

To make things more clear, an option has been added to have the 'Admin Only' as the recipient, so the Alert will go out ONLY to the Admin if  selected and the New Ticket Alert is enabled for the Admin Email.

If 'No one (diable Alerts and Notices)' is selected, no alert will go out at all, regardless of what is checked in Settings | Tickets | Alerts and Notices


include/staff/department.inc.php
```diff
@@ -312,6 +312,7 @@
                 <select name="group_membership">
 <?php foreach (array(
     Dept::ALERTS_DISABLED =>        __("No one (disable Alerts and Notices)"),
+    Dept::ALERTS_ADMIN_ONLY =>       __("Admin Only"),
```
protich: I'll change wording here to `Admin Email Only` - to avoid, "but I'm Admin" type of complaints.

aydreeihn: Updated!

PR #4973, opened 2019-07-31
issue: Reset Role Permissions

JediKev:
This addresses an issue reported on the Forum where clicking the "Reset Permissions" action in the Agent Directory shows a blank popup. Inspecting the console shows 500 error and checking the error logs shows `Too few arguments to function` error. This is due to the `RolePermission::__construct()` method expecting 1 parameter but in `ajax.staff.php` we are not passing anything to it. This updates `resetPermissions()` to pass `null` to `RolePermission::__construct()` so that it doesn't complain and error out.

### DIFF
--- a/include/ajax.staff.php
+++ b/include/ajax.staff.php
@@ -145,7 +145,7 @@ class StaffAjaxAPI extends AjaxController {
         $form = new ResetAgentPermissionsForm($_POST);
 
         if (@is_array($_GET['ids'])) {
-            $perms = new RolePermission();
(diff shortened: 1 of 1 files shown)


protich:
I would personally approach the solution from - should we have be passing something instead of allowing nothing... if its ok to pass null then _explain so to justify the change_. It's approach I prefer when fixing bugs instead of simply squashing the critter manifestation.

JediKev:
@protich Updated. I needed to pass `null` instead of allowing `null`. Thanks for this.

PR #4988, opened 2019-08-14
issue: Update Installer PHP Requirements

JediKev:
This updates the Installer PHP Requirements from `5.4` to `5.6`.


include/upgrader/prereq.inc.php
```diff
@@ -15,7 +15,7 @@
             <?php echo __('These items are necessary in order to run the latest version of osTicket.');?>
             <ul class="progress">
                 <li class="<?php echo $upgrader->check_php()?'yes':'no'; ?>">
-                <?php echo sprintf(__('%s or later'), 'PHP v5.4'); ?> - (<small><b><?php echo PHP_VERSION; ?></b></small>)</li>
+                <?php echo sprintf(__('%s or later'), 'PHP v5.6'); ?> - (<small><b><?php echo PHP_VERSION; ?></b></small>)</li>
```
protich: Use -  something like `SetupWizard::$prereq['php']` so we will have one spot to change it going forward. Better yet adding static methods might be better so you can make $prereq protected -> `protected $prereq = ....`

```php

static function getPHPVersion() {
     return self::$prereq['php']; 
}

```

setup/inc/install-prereq.inc.php
```diff
@@ -15,7 +15,7 @@
             <?php echo __('These items are necessary in order to install and use osTicket.');?>
             <ul class="progress">
                 <li class="<?php echo $installer->check_php()?'yes':'no'; ?>">
-                <?php echo sprintf(__('%s or greater'), '<span class="ltr">PHP v5.4</span>');?> &mdash; <small class="ltr">(<b><?php echo PHP_VERSION; ?></b>)</small></li>
+                <?php echo sprintf(__('%s or greater'), '<span class="ltr">PHP v5.6</span>');?> &mdash; <small class="ltr">(<b><?php echo PHP_VERSION; ?></b>)</small></li>
```
protich: See prior comment on upgrader...

JediKev:
@protich Updated

PR #4990, opened 2019-08-15
Feature: Merge/Link Tickets

aydreeihn:
This new feature adds the ability to merge or link tickets. Please see the link below for the full documentation:

https://github.com/osTicket/docs/blob/fc00e5fa8e34c5c1a868a18ad2662da5df8415dc/Features/Ticket%20Merge%20%26%20Link.rst

I'm open to feedback and suggestions!

Also, please note, this feature requires an upgrade, so do not merge this to your production codebase/database. You should test the feature on a new branch and test database until it is officially merged to the main codebase.


include/upgrader/streams/core/8b923d61-9b5550da.patch.sql
```diff
@@ -1,5 +1,5 @@
 /**
-* @signature 49fb8af84d9ac386a573f3eb80d8a0ee
+* @signature 9b5550dad8d3f036d62eb1340f230317
```
protich, line 2: Change @version on this file to 1.14

protich (changes requested):

include/upgrader/streams/core/8b923d61-9b5550da.patch.sql
aydreeihn: Done!

protich (approved):

ntozier:
@aydreeihn you are my favorite person right now. :)

philbertphotos:
This is life now.

Mike-Benoit:
Finally! This will be amazing and it has a ton of interesting functionality, great job! 

However after reading the full documentation, I didn't see anything about splitting existing tickets? In our case we would use this probably almost as much as linking/merging tickets and it seems like the two features would be quite similar to implement together.

aydreeihn:
@Mike-Benoit What exactly do you mean by splitting? We do have the option to create a ticket from a thread. So for example, if an Agent felt it might be better to reply to a User's message in a separate ticket/conversation, they could create a new ticket where that thread would be the first message of a new ticket.

Screen Shot 2019-08-29 at 11 12 00 AM

Screen Shot 2019-08-29 at 11 12 25 AM

Screen Shot 2019-08-29 at 11 12 41 AM

Screen Shot 2019-08-29 at 11 12 56 AM

TechCare-Aus:
@aydreeihn this is going to save me SOOO much time. So thank you very much. Just curious as to why the release date is pending v1.14 and we are on v1.12 :(

JediKev:
@TechCare-Aus

`v1.14` will be our next major realease. We are skipping `v1.13` as it’s an unlucky/bad number; like you don’t see hotels with a 13th floor. (triskaidekaphobia)

Cheers.

philbertphotos:
This good enough for production in a 1.12 environment?

f1ynng8:
Any rough plan to release v1.14?

Mike-Benoit:
> @Mike-Benoit What exactly do you mean by splitting? We do have the option to create a ticket from a thread. So for example, if an Agent felt it might be better to reply to a User's message in a separate ticket/conversation, they could create a new ticket where that thread would be the first message of a new ticket.
> 
> Screen Shot 2019-08-29 at 11 12 00 AM
> 
> Screen Shot 2019-08-29 at 11 12 25 AM
> 
> Screen Shot 2019-08-29 at 11 12 41 AM
> 
> Screen Shot 2019-08-29 at 11 12 56 AM

Creating a new ticket from a thread appears like it might help in some of the cases. The situation that seems to occur (way too often) for us is the customer creates the original ticket regarding the topic of "Apples", then replies to that same ticket for a question related to "Oranges". Its not always obvious right away that the two aren't related, so sometimes we could be 3+ exchanges in before we figure this out.

So ideally we would have a way to select X thread entries and move them to a completely new ticket, or even merge them with some other existing ticket. The PHPBB forum software has a really nice moderation tool for doing exactly that, but essentially its just checkboxes that allow you to select which thread entries you want to move, and if you want a pointer to the new topic (ticket) or not. 

Either way I look forward to giving this feature a try!

neoand:
> @aydreeihn you are my favorite person right now. :)

Me too !!!!!

Issue #4985, opened 2019-08-12
Changing help topic can lose response.


Mike-Benoit:
### Prerequisites

* [ N/A ] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [ NO ] Do you have any errors in the PHP error log, or javascript console?
* [ YES ] Did you check the [osTicket forums](http://osticket.com/forum/categories)?
* [ YES ] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/osTicket/osTicket/wiki/Github-Issues-Guidelines)

### Description

When replying to a ticket, setting a help topic causes the page to refresh and any response that may be written to be lost.

### Steps to Reproduce

We have the setting enable to require all tickets to have a help topic before they can be closed. So what often happens is the agent starts writing a reply to a ticket, then just before they are about click "Post Reply" they see the notice that a help topic needs to be set, so they scroll up to the help topic and click on the "Help Topic: None" link, set a help topic, the page refreshes and they lose their reply that they had written but were not yet allowed to post it.

### Versions

1.12.0, Ubuntu 18.04, Chrome.

JediKev:
@Mike-Benoit 

This is intended behavior. You're supposed to wait for a Draft to save (which is every 30 seconds after you stop typing) then you can do actions that refresh the page.

Cheers.

protich:
Let me clarify that this is not the intended behavior by design - it's how it works now _but_ we should most definitely do in-place edit on in-line edits **instead of** refreshing the entire page.

JediKev:
Please re-open if necessary.

PR #4994, opened 2019-08-19
Don't require refresh for inline edit

aydreeihn:
If an Agent uses inline edit to modify a single field, rather than saving the value and refreshing the page, the value will now populate in the form without having to refresh. This is useful in the event that an agent has begun writing a response to the user and does an inline edit before the draft has had a chance to save, causing them to lose their draft.

This fixes issue #4985.

### DIFF
--- a/include/ajax.tickets.php
+++ b/include/ajax.tickets.php
@@ -492,7 +492,6 @@ class TicketsAjaxAPI extends AjaxController {
                         );
                 Http::response(201, $ticket->getId());
             }
-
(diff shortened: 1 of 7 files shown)


protich:
Is this dependent on #5002 ?

aydreeihn:
Yes, I basically fixed everything with them combined. Together, everything works correctly.

protich:
@aydreeihn - please resolve conflicts.

PR #4996, opened 2019-08-20
issue: Support Exchange Shared Mailbox Auth

JediKev:
This addresses an issue where the system cannot successfully authenticate to an Exchange server using shared mailbox auth credentials _(eg. **Username:** `main-user@domain.com\shared-mailbox@domain.com`)_. The system passes the entire Username value as the `username` parameter for `imap_open()` and Exchange fails to authenticate. This updates the `MailFetcher::__construct()` method to explode the Username value at the backslash giving us two parts: the [0] Authentication Username and the [1] Mailbox Username. Once we break the Username value into two parts, we can then add the Authentication Username to the `mailbox` parameter as the `authuser` flag. In addition, this updates the `MailFetcher::getUsername` method to do the same thing, except this method will return the Mailbox Username. Once we get the Mailbox Username, we can then pass it as the `username` parameter for `imap_open` which successfully authenticates to Exchange.


protich:
Whenever you start duplicating logic e.g splitting username in this case then it's time to rethink the implementation otherwise you're setting up yourself for unmaintainable code. In my opinion - you should keep the logic handling specific case in one spot (duplicating even comments should have been a hint). 


```diff

JediKev:
@protich

Both sets of changes do separate things. The first set of changes splits the name and returns the authuser and the getUsername changes splits the name and returns the mailbox username.

Cheers.

protich:
@JediKev - Not really, it can still be consolidated - see the diff I posted. Can you imagine if we had another scenario of usernames format? It will be a mess.

JediKev:
@protich

The diff wasn’t posted/loaded when I responded. My bad.

Cheers.

JediKev:
@protich Updated.

PR #5002, opened 2019-08-21
Fix Saving For Fields:

aydreeihn:
This commit fixes/cleans up how we do the getChanges function in the FormField class. It also fixes saving/displaying the following field types:

- Multiselect Choices (clearing all options)
- Assignee Field
- Department Field
- Priority Field

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -655,6 +655,7 @@ class FormField {
         }
         return $this->_clean;
     }
+
     function reset() {
         $this->value = $this->_clean = $this->_widget = null;
     }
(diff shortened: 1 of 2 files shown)


include/class.forms.php
```diff
@@ -3083,6 +3078,10 @@ function setChoices($choices) {
         $this->_choices = $choices;
     }
 
+    function display($value) {
+        return $this->getAnswer()->value;
```
protich: Wouldn't this crash if a field is not a DynamicField?  That said, this should call display on answer and also note that an answer might not exist at all (static fields) or not yet on a particular object e.g new field gets added to Ticket Details form. System auto-creates form field answer when getting dynamic fields by adding missing fields but it's not a safe assumption to depend on.

Also curious where this is getting called from - what might `$value` be?

```php

function display($value) {
    if (($a=$this->getAnswer())
        return $a->display();

    return Format::htmlchars((string) $value);
} 
```

PR #5008, opened 2019-08-27
Add thread_type Field

aydreeihn:
In order to be able to delete thread and entries of deleted tickets, we need to add a thread_type field to the thread_event table so that we can accurately get dashboard statistics without having to rely on the object_type on the thread table.

**Note: This is rebased on top of the Schedules and Ticket Merge feature in order to be the latest patch. We must merge those two features before merging this!


include/upgrader/streams/core/9b5550da-e6978154.patch.sql
```diff
@@ -0,0 +1,55 @@
...
+* @version v1.14.0
+* @title Thread Type
+*
+* This patch adds a new field to the Thread Event table called thread_type
+* it allows us to be able to delete threads and thread entries when a ticket
+* is deleted while still maintaining dashboard statistics
+*
+*/
+-- Create a temporary table
+CREATE TABLE `tmp_table`
```
protich: Let's chat about this when you get a moment

PR #5009, opened 2019-08-27
issue: Organization Name Field Type-Changeable

JediKev:
This addresses an issue where Admins can change the Type of the Name field on the Organization Form. If changed to anything but Short Answer it will break anything having to do with Organizations' names. This adds a new, reusable method called `canChangeType()` to DynamicFormFields so we can set any one field to be non-type-changeable (not just special fields) by utilizing a new flag called `FLAG_MASK_TYPE` (0x80000). This updates the Installer yaml file to set the flag on new installs so new Installations cannot change the Organization Name Field type. This also adds to the upgrader so we can add the flag to existing Org. Name fields and change the Type if anything other than Short Answer.


protich:
When you get a chance please redo the PR to use INTERNAL mask as we discussed above.

JediKev:
Closing in favor of #5025.

Issue #5016, opened 2019-08-31
Wrong Queue for Today/Yesterday


philbertphotos:
The default queues seem to be off since tickets opened today is showing up in yesterdays Queue. 

OSticket recently updated to 1.12.2.
osTicket Version | v1.12.2 (a5d898b)     —  Up to date
Web Server Software | Apache/2.4.18 (Ubuntu)
MySQL Version | 5.7.26
PHP Version | 7.2.2

![image](https://user-images.githubusercontent.com/2892474/64057786-5aff4d00-cb6f-11e9-9df9-de85106fc8db.png)

ntozier:
You're not looking at tickets opened today or yesterday.
You are looking at ticket CLOSED yesterday or today.
So I do not follow you...

philbertphotos:
@ntozier the closed tickets in question were 8/30/2019 closed that day but shows up in the Yesterday Queue sorry for the confusion it does the same inour Open Queue after a few hours it moves to yesterdays queue even if its the middle of the afternoon

protich:
Question - What time zone do you have set on your profile? Asking because it does affect the day cutoff times.

philbertphotos:
@protich From what I can tell the times on tickets just created are pretty accurate.
![image](https://user-images.githubusercontent.com/2892474/64305177-76e26480-cf5c-11e9-81ae-c134aec0a2fa.png)

protich:
I'll look into it and get back to you - thanks.

protich:
Bro! I'll someday admit date time is confusing and hard! 

Try the patch below (if it applies) and let me know if it addresses the issue. I'll make an official PR tomorrow. 

```diff 

diff --git a/include/class.forms.php b/include/class.forms.php
index 2bfa740..93f691c 100644
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -2452,18 +2452,14 @@ class DatetimeField extends FormField {
                 "{$name}__gte" => $now->plus($interval),
             ));
         case 'period':
-            // Get the period range boundaries - timezone doesn't matter
-            $period = Misc::date_range($value, Misc::gmtime('now'));
+            // User's effective timezone
             $tz = new DateTimeZone($cfg->getTimezone());
-            // Get datetime boundaries in user's effective timezone
-            $tz = new DateTimeZone($cfg->getTimezone());
-            $start = new DateTime($period->start->format('Y-m-d H:i:s'),
-                    $tz);
[...]

PR #5026, opened 2019-09-05
Date Range Period  / User's Timezone

protich:
This commit fixes an issue where date range boundaries didn't consider user's timezone resulting in UTC times being used.


protich:
Addresses issue #5016

PR #5037, opened 2019-09-09
Custom Priority Field Blank

aydreeihn:
If no value is saved to a custom Priority field, we need to make sure we don't try to treat a null value as an array. We should instead just return the null priority.

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -2972,10 +2972,14 @@ class PriorityField extends ChoiceField {
         return $this->getPriority($id);
     }
 
-    function to_database($prio) {
-        return ($prio instanceof Priority)
-            ? array($prio->getDesc(), $prio->getId())
-            : array($prio[key($prio)],key($prio));
+    function to_database($value) {
+        if ($value instanceof Priority)
+            return array($value->getDesc(), $value->getId());
+
+        if (is_array($value))
+            return array(current($value), key($value));
+
+        return $value;
     }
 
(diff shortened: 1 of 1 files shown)


include/class.forms.php
```diff
@@ -2975,7 +2975,9 @@ function to_php($value, $id=false) {
     function to_database($prio) {
         return ($prio instanceof Priority)
             ? array($prio->getDesc(), $prio->getId())
-            : array($prio[key($prio)],key($prio));
+            : (is_array($prio)
```
protich: This is getting too nested for me - how about something like this.

```diff

diff --git a/include/class.forms.php b/include/class.forms.php
index 8bc21f2..914d83f 100644
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -2972,10 +2972,14 @@ class PriorityField extends ChoiceField {
         return $this->getPriority($id);
     }
 
-    function to_database($prio) {
-        return ($prio instanceof Priority)
-            ? array($prio->getDesc(), $prio->getId())
-            : array($prio[key($prio)],key($prio));
+    function to_database($value) {
+        if ($value instanceof Priority)
+            return array($value->getDesc(), $value->getId());
+
+        if (is_array($value))
+            return array(current($value), key($value));
+
+        return $value;
     }
 
     function display($prio, &$styles=null) {
```

protich (changes requested):

protich (approved):

Issue #5035, opened 2019-09-08
Feature "Ticket Filter" listed in ticket Event


philbertphotos:
### Prerequisites
![image](https://user-images.githubusercontent.com/2892474/64491391-2b6ad780-d236-11e9-9f14-3074a3c093da.png)

* [ ] Can you reproduce the problem in a fresh installation of the "develop" branch?
Yes
* [ ] Do you have any errors in the PHP error log, or javascript console?
No
* [ ] Did you check the [osTicket forums](https://forum.osticket.com/)?
yes
### Description
Would be nice for the Ticket Filter to get credit when it assigns a task to user or ticket. Makes troubleshooting way way easier and would make more sense since a user cannot assign a ticket to an agent.

JediKev:
@philbertphotos 

Take a look at #5042. It is not approved yet and is subject to change.

Cheers.

philbertphotos:
Will try it out.

PR #5042, opened 2019-09-12
issue: Ticket Filter Assignment Event

JediKev:
This addresses #5035 where the ticket Assignment Event show the Ticket Owner's name or "SYSTEM" as the assigner when tickets are created via Client Portal, API, or Email Fetching. This updates the event to include "Ticket Filter" as the assigner if the assignment was triggered by a Ticket Filter Action. This makes for more accurate events and is very useful for statistics/debugging. This keeps current functionality of Agents creating tickets shows the creating Agent's name as the assigner. This also adds a new parameter to `assignToAgent()` and `assignToTeam()` called `$user` (defaults to null) that allows us to pass different usernames/user objects for the events. This makes the two methods versatile so if we add new ways to assign to Staff/Teams in the future we can pass the new name/object with no issues.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -2708,7 +2708,7 @@ implements RestrictedAccess, Threadable, Searchable {
         return $this->assignToStaff($assignee, $form->getComments(), false);
     }
 
-    function assignToStaff($staff, $note, $alert=true) {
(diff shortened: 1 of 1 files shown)


philbertphotos:
Will implement it ...thanks.

PR #5046, opened 2019-09-13
Draft Saving in New Redactor

aydreeihn:
- Make auto save time every 10 seconds
- Modify saveDraft button for new redactor
- Fix inline edit auto saves

### DIFF
--- a/include/staff/ticket-view.inc.php
+++ b/include/staff/ticket-view.inc.php
@@ -153,7 +153,7 @@ if($ticket->isOverdue())
                     <li><a class="change-user" href="#tickets/<?php
                     echo $ticket->getId(); ?>/change-user"
                     onclick="javascript:
-                        $('#response').redactor('draft.saveDraft');"
+                        saveDraft();"
                     ><i class="icon-user"></i> <?php
                     echo __('Change Owner'); ?></a></li>
                 <?php
@@ -318,7 +318,7 @@ if($ticket->isOverdue())
                               data-redirect="tickets.php?id=<?php echo $ticket->getId(); ?>"
                               href="#statuses"
                               onclick="javascript:
-                                  $('#response').redactor('draft.saveDraft');"
+                                  saveDraft();"
                               >
                               <?php echo $ticket->getStatus(); ?>
                           </a>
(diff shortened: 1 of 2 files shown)


Mike-Benoit:
Would be really nice if the draft saving feature was a little smarter. Such as auto-saving everytime the user leaves focus on the textarea, or the browser goes to another page. That can help prevent lost input between the auto-save interval and the above actions, which unfortunately happens to me way too frequently. :(

JediKev:
@Mike-Benoit

With this pull request you can wait 10 seconds or click the Save Draft button. So before you leave the reply box click the Save Draft button and you’re good.

Cheers.

Mike-Benoit:
@JediKev I understand it can be done manually, but the entire point of saving a draft is to prevent the user from losing what they typed. 10 seconds is actually quite a long time, easily a sentence if the agent can type at 60WPM or faster. 

The real issue though is that its difficult for the agent to know that they lost something, I've personally had several cases where I left the screen (lost the entire last sentence), gone back to it, gave it a quick once over proof read and sent the response with it still missing. Since the majority of the response is already there, its even easier to overlook that something is missing. 

Its further made worse by issues like this: #4985 

A simple onblur event trigger to save the draft would greatly reduce the chance of this happening without having to remember to do anything manually.

aydreeihn:
@Mike-Benoit The 10 seconds is reduced from the previous 60 seconds, and it's there because if drafts are saved too frequently, it could make your helpdesk unstable. If you'd prefer though, you can reduce the wait time so that you can guarantee you won't lose your draft, just realize it could have some unwanted side effects. Issue #4985 has been addressed and the page no longer refreshes when updating fields using inline edit. We've also added an indication on the bottom of redactor that says either 'unsaved' or 'all changes have been saved' so you can easily see if you draft is saved or not.

Mike-Benoit:
@aydreeihn The reasons that cause input to be lost is often the same reason why displaying a message/save draft button on the screen isn't helpful, the agent is distracted. They are in the middle of finishing up a response and the phone rings, or a colleague interrupts them and they navigate somewhere else. 

Being an agent on the front-line is a tough, stressful job filled with distractions and juggling multiple tasks all day long. A message showing changes are unsaved, or a button to manually push to save them is certainly going to be missed and adds yet another thing for them to remember to do when they are least likely to do so. 

A timer setting, especially one that can't be easily configured without modifying source code, is always going to be a sore point for people. Too short for some, too long for others. But adding simple cases where you know with a high level of certainty when a good point to save the input is, has very little downside. 

onblur, onbeforeunload, when a character that ends a sentence is typed (".", "?", "!") and as a last resort timer of X seconds. Run them all through a short debouncer so the server doesn't get hammered and you should be able to eliminate lost input entirely.

PR #5054, opened 2019-09-18
Issue: Empty Due Date

aydreeihn:
This commit fixes an issue that allowed Agents to clear the Duedate on a ticket using inline edit. We were saving the value 0000-00-00 00:00:00 to the database which the ticket would then display as 12/1/02 06:09 pm.

Instead, if an Agent clears the Duedate field, we should save the duedate as null and the ticket should continue using the est_duedate.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -3339,16 +3339,20 @@ implements RestrictedAccess, Threadable, Searchable {
                 $fid = $field->get('name');
 
                 // Convert duedate to DB timezone.
-                if ($fid == 'duedate'
-                        && ($dt = Format::parseDateTime($val))) {
-                          // Make sure the due date is valid
-                          if (Misc::user2gmtime($val) <= Misc::user2gmtime())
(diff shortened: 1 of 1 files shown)


include/class.ticket.php
```diff
@@ -3339,16 +3339,20 @@ function updateField($form, &$errors) {
...
-                          // Make sure the due date is valid
-                          if (Misc::user2gmtime($val) <= Misc::user2gmtime())
-                              $errors['field']=__('Due date must be in the future');
-                          else {
-                              $dt->setTimezone(new DateTimeZone($cfg->getDbTimezone()));
-                              $val = $dt->format('Y-m-d H:i:s');
-                          }
-                }
+                if ($fid == 'duedate') {
+                    if ($dt = Format::parseDateTime($val)) {
```
protich: Why not something like... 

```php
if (empty($val)) {
   $val = null;
} elseif ($dt = Format::parseDateTime($val)) {
...
}

PR #5055, opened 2019-09-18
Show list of nested help topics on edit.

protich:
Inline edit didn't show nested help topics on the edit modal.

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -2703,17 +2703,15 @@ class TopicField extends ChoiceField {
 
     function getTopics() {
         if (!isset($this->topics))
-            $this->topics = Topic::objects();
+            $this->topics = Topic::getHelpTopics(false, false, true);
 
         return $this->topics;
     }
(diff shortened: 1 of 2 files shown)


include/class.forms.php
```diff
@@ -2732,10 +2730,8 @@ function hasIdValue() {
 
     function getChoices($verbose=false) {
         if (!isset($this->_choices)) {
-            $choices = array('' => '— '.__('Default').' —');
-            foreach ($this->getTopics() as $t)
-                $choices[$t->getId()] = $t->getName();
-            $this->_choices = $choices;
+            $this->_choices = array('' => '— '.__('Default').' —') +
```
aydreeihn, line 2733: Just want to note that this default value doesn't actually choose the default topic, but that's part of what I'm working on in another PR. Other than that, looks good.

protich: Dully noted, thanks

PR #5056, opened 2019-09-19
Queue Sort on Search

protich:
This commit adds a default sort of ticket create date for adhoc and saved searches.

TODO: 

- [x]  Ability to create / select default sort for Saved Queues.

### DIFF
--- a/include/class.search.php
+++ b/include/class.search.php
@@ -655,7 +655,7 @@ class SavedQueue extends CustomQueue {
     private $_columns;
     private $_settings;
     private $_form;
-
+    private $_sorts;
 
 
     function __onload() {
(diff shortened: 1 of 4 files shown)


aydreeihn:
Should I have to log out and log back in for the default sorting to take effect in saved searches?

include/class.search.php
```diff
@@ -841,8 +854,13 @@ function update($vars, &$errors=array()) {
             }
         }
 
-        if (!$errors && $this->_config->update($vars, $errors))
+        if (!$errors && $this->_config->update($vars, $errors)) {
+            // reset settings
             $this->_settings = $this->_criteria = null;
+            // Reset chached queue options
+            unset($_SESSION['sort'][$this->getId()]);
```
protich, line 884: No this line is supposed to clear it.

protich: PR updated to fix the issue. Thanks for the find.

PR #5058, opened 2019-09-19
Oops: Lint Fixes

aydreeihn:
- Misspelled method calls


aydreeihn:
@protich There was one more that it found, but I'm not sure what it needs to be changed to in class.file.php:

```
function getContents() {
        $this->fseek(0);
        return $this->fread($this->getSize());
    }
```

PR #5059, opened 2019-09-20
fix user email size.

ntozier:
address is an email, and should be varchar(255) like it is for the %TABLE_PREFIX%email table


protich:
Missing upgrade patch - correctly reimplemented by #5063

PR #5063, opened 2019-09-23
issue: Staff/User Email Length

JediKev:
This implements #5059 to make it valid for upgrades and new installs. This updates the Staff/User email lengths to match the `ost_email.email` column length. This makes all email address lengths consistent throughout the helpdesk.


include/upgrader/streams/core/e6978154-cc6d1b03.patch.sql
```diff
@@ -0,0 +1,24 @@
...
+*/
+
+-- Increase Staff Email Address length
+ALTER TABLE `%TABLE_PREFIX%staff` MODIFY COLUMN `email` VARCHAR(255);
+
+-- Increase User Email Address length
+ALTER TABLE `%TABLE_PREFIX%user_email` MODIFY COLUMN `address` VARCHAR(255);
+
+-- Update Contact Form > Email Address field length
+UPDATE `%TABLE_PREFIX%form_field` SET `configuration` = REPLACE(`configuration`,'"length":64','"length":255') WHERE `name` = 'email';
```
protich, line 19: I don't think this is necessary. It's not max length, I doubt you want a input of such size.

Issue #5053, opened 2019-09-18
New or editing notes is broken


TechCare-Aus:
### Prerequisites

* [ ] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [X] Do you have any errors in the PHP error log, or javascript console?
* [X] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [X] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/osTicket/osTicket/wiki/Github-Issues-Guidelines)

### Description

Problem with the notes since upgrading to the latest version of osticket-develop (presumably very similar to 1.14rc1). Organisation notes are broken.

### Steps to Reproduce

1. Click organisation
2. Click on notes
3. Edit existing note

**Expected behavior:** 
Same behaviour as before. Edit by default opens up in WYSIWYG editor, then option to change to code mode, then ability to save is displayed. Ideally an option to undo changes or at least see the changes made too (this is a sneaky attempt to add a brilliant option to notes)

**Actual behavior:** [What actually happened]
Display is cut to small version, only shows up as code (no WYSIWYG editor mode) and no save button

### Versions

osTicket Version | 1.14-git (5ea4fb5)
Web Server Software | Apache
MySQL Version | 10.3.18
PHP Version | 7.3.9

JediKev:
@TechCare-Aus 

I see you say you are using the `develop` branch but that branch actually correlates to `1.12-git`. I think you meant you're using the latest from `develop-next` branch if you're using v1.14-rc1. 😉

I notice the same issue with v1.14-rc1 and I do believe it was due to the Redactor upgrade. We will look into this. 👍

Cheers.

TechCare-Aus:
> I see you say you are using the develop branch but that branch actually correlates to 1.12-git. I think you meant you're using the latest from develop-next branch if you're using v1.14-rc1. 😉

haha sorry mate. yes typo.

On a side note. When I've updated to the latest git on branch develop-next and go to information next to `osTicket Version | 1.14-git (5ea4fb5)` it tells me to upgrade. Is this due to it thinking that v1.14RC1 is a higher version?

JediKev:
@TechCare-Aus

It’s just not in our dns yet as it’s an RC so it’s going to say that for now.

Cheers.

TechCare-Aus:
[...]

PR #5066, opened 2019-09-25
issue: Redactor QuickNotes

JediKev:
This addresses issue #5053 where since the Redactor Upgrade, QuickNotes are not working correctly. This is due to using old redactor methods instead of the new methods. This updates QuickNotes to use the new methods so they work properly again.

### DIFF
--- a/scp/js/scp.js
+++ b/scp/js/scp.js
@@ -1199,7 +1199,7 @@ $(document).on('click.note', '.quicknote .action.edit-note', function(e) {
     if (note.closest('.dialog, .tip_box').length)
         T.addClass('no-bar small');
     body.replaceWith(T);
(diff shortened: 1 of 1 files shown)


TechCare-Aus:
Still not working properly. Can create the notes now, but not edit them. Same issue as before.

TechCare-Aus:
Hmmm. I don't know why but today I tried it again and it is working. Previously I used Chrome's incognito and it didn't work, so maybe it was a CloudFlare caching thing. 

My apologies guys.

protich:
@JediKev - I'll hold off the merging until you can confirm latest comment. Thanks

JediKev:
@protich 

Everything is working as expected when applying the diff:
diff

Cheers.

PR #5076, opened 2019-10-01
Datetime Formats

protich:
This PR addresses two issues related to date / time format.

* 24 hrs format, when selected, is now used system-wide even on thread entries
* PHP-to-JS date / time format translation is now down in the backend - bug on year format fixed.


aydreeihn:
Based on my testing this looks good.

Issue #5095, opened 2019-10-09
Call to undefined function openssl_pkey_get_public() when installing a Language Pack


bes-19:
### Prerequisites

* [Y] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [Y] Do you have any errors in the _PHP error log_ (IIS Failed Request Logging!)?
[W3SVC1.zip](https://github.com/osTicket/osTicket/files/3707959/W3SVC1.zip)
Did not upload a Video because of the FileSize...
Files can be opened with Internet Explorer 11 (Doubleclick the xml Files)
* [Y] Did you check the osTicket forums?
* [Y] Did you perform a cursory search to see if your bug or enhancement is already reported?

### Description

Installing a Language pack (tested it with 10 different Language Packs from osticket.com/download) causes to crash the system.php Site.

### Steps to Reproduce

1. Put a Language Pack of your Choice in the include/i18n folder
2. Open _localhost_/scp/system.php

**Expected behavior:** [What you expected to happen]
Showing the Site correctly

**Actual behavior:** [What actually happened]
Showing a HTTP 500 Error with attached PHP Error

### Versions

osTicket: 1.12-git (develop branch downloaded on 09.10.2019 16:40 UTC+2)
Webserver: IIS 10.0
mySQL: 8.0.17
PHP: 7.3.8
Firefox Browser 69.0

JediKev:
@bes-19 

>Call to undefined function openssl_pkey_get_public()

This means you do not have the OpenSSL library installed on your server. More information on this here:
- https://www.openssl.org/

With this being said we will add a check for OpenSSL Library and if it's not available we just won't run the corresponding methods. 👍

Cheers.

JediKev:
@bes-19 

Here you go:
- https://github.com/osTicket/osTicket/pull/5096

Cheers.

bes-19:
Thanks JediKev.

Maybe add the OpenSSL Extension to the PHP Extensions Table:
![OpenSSL](https://user-images.githubusercontent.com/54275066/66546491-6fa30d80-eb3d-11e9-94cb-e73f41a6aa6e.PNG)
(Also in install.php...)

JediKev:
@bes-19 

Well, to be fair, OpenSSL is not a PHP extension. OpenSSL is a Software Library and is not required to use osTicket.

Cheers.

bes-19:
Hm ok... Don't know this much about Programming :)
Thanks for Explanation :)

PR #5096, opened 2019-10-09
i18n: OpenSSL Error

JediKev:
This addresses issue #5095 where installing a Language Pack and visiting the Dashboard > Information page without OpenSSL installed/configured will throw a fatal error of `Call to undefined function openssl_pkey_get_public()`. This is due to no check for the `openssl` extension before the dependant method is called. This adds a check so that if OpenSSL is not available we can return a useful error code of `VERIFY_EXT_MISSING` (which is error code `2`) and continue without disrupting the content.

### DIFF
--- a/include/class.plugin.php
+++ b/include/class.plugin.php
@@ -534,7 +534,7 @@ abstract class Plugin {
     static function isVerified($phar) {
         static $pubkey = null;
 
-        if (!class_exists('Phar'))
+        if (!class_exists('Phar') || !extension_loaded('openssl'))
             return self::VERIFY_EXT_MISSING;
         elseif (!file_exists(INCLUDE_DIR . '/plugins/updates.pem'))
             return self::VERIFY_NO_KEY;


include/class.plugin.php
```diff
@@ -539,7 +539,7 @@ static function isVerified($phar) {
         elseif (!file_exists(INCLUDE_DIR . '/plugins/updates.pem'))
             return self::VERIFY_NO_KEY;
 
-        if (!isset($pubkey)) {
+        if (!isset($pubkey) && extension_loaded('openssl')) {
```
protich: I know this technically resolves the issue but if you look closely we end up returning "wrong" exit code when a more suitable one, `self::VERIFY_EXT_MISSING`, is available. This is especially important if UI is (or will be) dependent on the returned value. I would refactor and put the check around line 537 as shown below. 

```php
       if (!class_exists('Phar') || !extension_loaded('openssl'))
            return self::VERIFY_EXT_MISSING;
```

protich (changes requested):

protich (approved):

PR #5099, opened 2019-10-10
Auto-refer tickets to assigned Agent on close

protich:
This PR adds a setting to Admin Panel > Settings > Tickets to enable ability to auto-refer tickets, on close, to assigned or closing agent. This is necessary when you want to give agents with limited access continued access to assigned tickets after they're closed.

It also addresses a bug where `closed` date was set to `reopened` date when a closed ticket is reopen.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -1443,7 +1443,7 @@ implements RestrictedAccess, Threadable, Searchable {
 
     // Ticket Status helper.
     function setStatus($status, $comments='', &$errors=array(), $set_closing_agent=true) {
-        global $thisstaff;
+        global $cfg, $thisstaff;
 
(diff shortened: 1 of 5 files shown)


include/class.ticket.php
```diff
@@ -1518,6 +1520,10 @@ function setStatus($status, $comments='', &$errors=array(), $set_closing_agent=t
         if (!$this->save())
             return false;
 
+        // Refer thread to previously assigned or closing agent
+        if ($refer && $cfg->autoReferTicketsOnClose())
+            $this->thread->refer($refer);
```
aydreeihn, line 1525: I think we should log a referral thread event here.

aydreeihn:
The code looks good. I just wonder if we should refer the assigned agent and the closing agent if they are different?

PR #5109, opened 2019-10-16
feature: Fetch From Mail Folder

JediKev:
This feature adds a textbox in the Email configurations that allows you to configure a specific Folder to fetch from. If no folder is provided the system will fetch from the INBOX as usual. This also adds a check so that if a Folder is given but doesn't exist and can't be created then the system will error out with a detailed message.

**TODO:** 
- [x] Sanitize user input (`mail_folder`).
- [x] Add errors for using Folder on POP.


include/staff/email.inc.php
```diff
@@ -277,7 +287,7 @@
                 <label><input type="radio" name="postfetch" value="archive" <?php echo ($info['postfetch']=='archive')? 'checked="checked"': ''; ?> >
                 <?php echo __('Move to folder'); ?>:
                 <input type="text" name="mail_archivefolder" size="20" value="<?php echo $info['mail_archivefolder']; ?>"/></label>
-                    &nbsp;<font class="error"><?php echo $errors['mail_folder']; ?></font>
+                    &nbsp;<font class="error"><?php echo $errors['postfetch']; ?></font>
```
aydreeihn: Wouldn't the error message be displayed twice if we echo it here and then below as well? (below is on line 298)

JediKev: (discussed in person but putting comment here as well)

This error is not actually used so I will change it to something else or remove it altogether.

PR #5111, opened 2019-10-17
In-reply-to header in postReply

mchobbel:
This patch solves the issue of a missing In-reply-to-header. This header was unfortunately not sent, when, as an Agent you add a Reply to an existing Ticket.  The problem, for the user who recieves such email messages in their INBOX (outside osTicket), is that the messages are not displayed in proper thread-format.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -3008,6 +3008,16 @@ implements RestrictedAccess, Threadable, Searchable {
             $attachments = $cfg->emailAttachments() ?
                 $response->getAttachments() : array();
 
+            $thread = $this->getThread();
+            $lastm = $thread->getLastEmailMessage();
+            if($lastm){
+                $mid = $lastm->getEmailMessageId();
+                if($mid){
+                    $options += array( 'inreplyto'=> $mid );
+                }
+            }
+
+            //$this->qdebug(print_r($lastm,true));
             //Send email to recepients
             $email->send($recipients, $msg['subj'], $msg['body'],
                     $attachments, $options);


protich:
@mchobbel - Thank you for the PR. I believe we already have the logic to add `Reply-To` header in addition to other `References`  in `include/class.mailer.php`. 

```vim
329         // Check if the email is threadable
330         if (isset($options['thread'])
331             && $options['thread'] instanceof ThreadEntry
332             && ($thread = $options['thread']->getThread())) {
333 
334             // Add email in-reply-to references if not set
335             if (!isset($options['inreplyto'])) {
336 
337                 $entry = null;
338                 switch (true) {
339                 case $recipients instanceof TicketOwner:
340                 case $recipients instanceof Collaborator:
341                     $entry = $thread->getLastEmailMessage(array(
342                                 'user_id' => $recipients->getUserId()));
343                     break;
344                 case $recipients instanceof Staff:
345                     //XXX: is it necessary ??
346                     break;
347                 }
348 
349                 if ($entry && ($mid=$entry->getEmailMessageId())) {
350                     $options['inreplyto'] = $mid;
351                     $options['references'] = $entry->getEmailReferences();
352                 }
353             }
```

I'm curious why that's not working for you.

mchobbel:
Hi Peter, thanks for looking into this. 
The fragment you pointed out (in function Mailer->send() )  depends on option 'inreplyto' to be set (line 335).  If you look at Ticket -> postReply, you can see (at the very end of the function block)  email->send() ; that's where you enter the Mailer->send(),  passing the options-paramater. But  here it goes wrong because the 'inreplyto' option is not provided. That's what my patch addresses.

protich:
> fragment you pointed out (in function Mailer->send() ) depends on option 'inreplyto' to be set (line 335).

That's not true - it's when it's not set -> `!isset($options['inreplyto'])` and thread-able (`thread` set).

PS: We're testing further.

protich:
@mchobbel - the root issue ended up being `$recipients` being an instance of `MailingList` - which wasn't handle in `class.mailer.php`. We will make a PR to address the issue.

JediKev:
@mchobbel 

Below is the patch for this issue:
- https://github.com/osTicket/osTicket/pull/5138

Cheers.

protich:
Thanks for the PR - #5138 fixes the issue.

mchobbel:
Very nice, thank you.

PR #5112, opened 2019-10-17
Add GTID compatibility

kler:
GTID-based replication does not allow `CREATE TABLE ... SELECT`
See https://dev.mysql.com/doc/refman/5.7/en/replication-gtids-restrictions.html

This PR change the routine to perform the operation using a temporary table, instead of a `CREATE TABLE ... SELECT`.

To verify the result, start MySQL using these params:
```--gtid-mode=ON --enforce-gtid-consistency=ON --log-bin --server-id=0```

Current osTicket will fail when running MySQL using these params, but with PR it will work and the result are the same (i.e. __cdata table is populated in the same way).

### DIFF
--- a/include/class.dynamic_forms.php
+++ b/include/class.dynamic_forms.php
@@ -286,10 +286,31 @@ class DynamicForm extends VerySimpleModel {
     }
 
     static function buildDynamicDataView($cdata) {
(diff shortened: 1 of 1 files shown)


kler:
Realised I already pushed this well over 2 years back...

Pro-tip: accept that PR, as it's a small, well contained patch and will provide huge value - since no one as of now can run the application in MySQL replicated clusters basically.

See PR #3832

Issue #5049, opened 2019-09-17
Merge ticket Issue


johnkalangha:
OSticket 1.14.RC1 
PHP 7.3
Mysql 5.6

hi, report issue, when i merge the two ticket, the another ticket will disappear in Open queued and it will merge in the parent ticket(agent side) and when i closed the ticket in the agent side, for user side only the parent ticket will be closed how about the other ticket, how can i change the status to closed since there are no open ticket in agent side?

see for the screenshot here: https://forum.osticket.com/d/95484-sla-something-wrong-in-1-14rc1/2

aydreeihn:
This is actually the intended behavior. When a ticket is merged, only the parent remains in the queue because the child ticket no longer has any thread entries. If the child ticket is kept in the system, its only purpose is to provide historical field data for the ticket. Replies can no longer come to the child. They will automatically be routed to the parent ticket. You can still find the children tickets that were merged by doing an advanced search where merged is checked. From there, you would be able to close that ticket if you wanted to. Additionally, there is an option to delete children tickets when doing a merge so that the parent ticket can take over from there.

Here is a link to the documentation for an in depth explanation of the feature as a whole:
https://github.com/osTicket/docs/blob/fc00e5fa8e34c5c1a868a18ad2662da5df8415dc/Features/Ticket%20Merge%20%26%20Link.rst

johnkalangha:
so basically, in the user view, the ticket will remain open? so there no workaround to change it in closed status?

aydreeihn:
The agent can close the ticket as usual and then the ticket will be moved to the closed queue for users as well.

Before closing child:
![Screen Shot 2019-09-24 at 8 53 07 AM](https://user-images.githubusercontent.com/22985133/65517688-c73c4a80-dea8-11e9-8a7d-90fb5c0657bd.png)

Search for Merged Tickets:
![Screen Shot 2019-09-24 at 8 53 55 AM](https://user-images.githubusercontent.com/22985133/65517841-f0f57180-dea8-11e9-9a09-a9b29b9e7068.png)

Merged Tickets Including Children:
![Screen Shot 2019-09-24 at 8 54 13 AM](https://user-images.githubusercontent.com/22985133/65517853-f9e64300-dea8-11e9-9f15-09b78f76b32e.png)


Child in Agent Panel:
[...]

PR #5114, opened 2019-10-17
Ticket Merge: Close Children

aydreeihn:
When tickets are merged, we should automatically close children Tickets.

This references issue #5049

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -2533,6 +2533,7 @@ implements RestrictedAccess, Threadable, Searchable {
                     $child->getThread()->setExtra($parentThread);
 
                 $child->setMergeType($options['combine']);
+                $child->setStatus(intval($options['statusId']));
 
                 if ($options['delete-child'] || $options['move-tasks']) {
                     if ($tasks = Task::objects()
--- a/include/staff/templates/merge-tickets.tmpl.php
+++ b/include/staff/templates/merge-tickets.tmpl.php
@@ -97,6 +97,26 @@ foreach ($tickets as $t) {
     </label>
 </div>
 <br/><br/>
+<div id="child-status">
+&nbsp;&nbsp;&nbsp;
+    <label class="inline checkbox">
+        <?php echo __('Child Status');?>
+        <select id="statusId" name="statusId">
+        <?php
+        $states = array('closed');
+        foreach (TicketStatusList::getStatuses(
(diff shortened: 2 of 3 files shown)


include/class.ticket.php
```diff
@@ -2533,6 +2541,7 @@ function merge($tickets) {
                     $child->getThread()->setExtra($parentThread);
 
                 $child->setMergeType($options['combine']);
+                $child->setState('closed');
```
protich: This is problematic - since wording can be different depending on install language. I don't have a solution at hand!

include/staff/templates/merge-tickets.tmpl.php
```diff
@@ -97,6 +97,26 @@
...
+        <?php echo __('Child Status');?>
+        <select id="statusId" name="statusId">
+        <?php
+        $states = array('closed');
+        foreach (TicketStatusList::getStatuses(
+                    array('states' => $states)) as $s) {
+            if (!$s->isEnabled()) continue;
+            echo sprintf('<option value="%d">%s</option>',
+                    $s->getId(),
+                    __($s->getName()));
```
protich: I believe this need to be `$s->getLocalName()` for translated status name.

include/class.ticket.php
```diff
@@ -1445,6 +1445,14 @@ function setStatus($status, $comments='', &$errors=array(), $set_closing_agent=t
...
             return false;
 
+        if (is_string($status)) {
+            $status = TicketStatus::objects()
+                ->filter(array('name'=>$status))
+                ->values_flat('id')
+                ->first();
+            $status = $status[0];
+        }
+
```
protich: This block of code is no longer needed.

protich (changes requested):

include/staff/templates/merge-tickets.tmpl.php
aydreeihn: It looks like it's good since it has __() around it. Not sure why, but when I dump just getLocalName(), it's not actually translated.

protich: That's not how inline translation works - I'll explain in person when you have a second.

aydreeihn: Kevin just explained it to me. Didn't realize those were supposed to be manually edited.

protich (approved):

PR #5119, opened 2019-10-22
feature: Separate SMTP Credentials

JediKev:
This feature adds the ability to separate the IMAP/POP and SMTP credentials for a System Email. This adds a checkbox next to the Authentication Required setting where, if Enabled, will show two additional inputs for the SMTP Username and Password. If Authentication Required is set to No the SMTP Username and Password inputs will disappear as they are not needed. If the checkbox is Disabled the system will use the IMAP/POP credentials for SMTP.

**Disabled:**
unchecked

**Enabled:**
checked


include/class.mailfetch.php
```diff
@@ -126,14 +126,18 @@ function getMaxFetch() {
         return $this->ht['max_fetch'];
     }
 
+    function getMailFolder() {
+        return $this->mailbox_encode($this->ht['folder']) ?: 'INBOX';
```
protich: I know mailbox_encode returns a null on null folder...  but that's not guaranteed forever should the logic get modified...  something like below will make me comfortable. 

```php
return $this->mailbox_encode($this->ht['folder'] ?: 'INBOX');
```

include/staff/email.inc.php
```diff
@@ -324,9 +334,28 @@
                  &nbsp;
                  <label><input type="radio" name="smtp_auth"  value="0"
                      <?php echo !$info['smtp_auth']?'checked':''; ?> /> <?php echo __('No'); ?></label>
+                 &nbsp;
+                 <label><input type="checkbox" name="smtp_auth_creds" value="1"
+                     <?php echo $info['smtp_auth_creds']?'checked':''; ?> />Use Separate Authentication</label>
```
protich: Put a space before and make text translatable.

protich (changes requested):

include/class.mailfetch.php
JediKev: Done

include/staff/email.inc.php
JediKev: Done

Issue #5120, opened 2019-10-23
[KB] Main FAQ Categories not translatable v1.12.3


datenbrei:
### Prerequisites

* [ ] Can you reproduce the problem in a fresh installation of the "develop" branch?
I can't
* [x] Do you have any errors in the PHP error log, or javascript console?
* [x] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [x] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/osTicket/osTicket/wiki/Github-Issues-Guidelines)

### Description

All own **main** categories of our FAQ, which are translated in the backend, do not appear in english, but all **non-main**-categories. All are translated in the backend, they are our own categories, it has nothing to do with the translation of the software.

![translations-missing](https://user-images.githubusercontent.com/13457963/67394095-ec090800-f5a3-11e9-9d9d-4f10770fdc39.png)

In /var/log/apache/error.log I see the following:
~~~
[Wed Oct 23 07:17:47.097469 2019] [:error] [pid 13885] [client 217.149.165.223:2918] script '/var/www/sites/support/kb/ajax.php' not found or unable to stat, referer: https://support.human-connection.org/kb/index.php
~~~
This file is indeed missing and it is also not part of the distribution:
![no-ajax](https://user-images.githubusercontent.com/13457963/67394520-a13bc000-f5a4-11e9-9c29-78badc3b53cc.png)

### Steps to Reproduce

1. Translate all your own categories to german or English
2. Write some FAQs
3. See, that only the translations of the sub-categories appear, but not the main categories

**Expected behavior:** [What you expected to happen]
Any translated main-category should be displayed correctly.

**Actual behavior:** [What actually happened]
see above

### Versions

osTicket-Version: v1.12.3 (bcf1a6f)  
Server-Software: Apache/2.4.41 (Ubuntu)
MySQL-Version: 10.0.38
PHP-Version: 5.6.40-12+ubuntu16.04.1+deb.sury.org+1

Browser: Firefox 69

JediKev:
@datenbrei 

Here you go:
- https://github.com/osTicket/osTicket/pull/5121

It was actually two separate issues; thanks for the report!

Cheers.

PR #5121, opened 2019-10-23
i18n: KnowledgeBase Category and JS

JediKev:
This addresses #5120 where main (or Parent) FAQ Category Titles are not translated even though the child Category titles are translated correctly. This is due to to the system pulling the Full Name instead of the Local (translated) Name.

In #5120, the reporting party also mentioned that they receive an error of `script '/xxx/osTicket/kb/ajax.php' not found or unable to stat`.  This is due to the AJAX call being inside the KB directory, and `/kb/ajax.php` truly does not exist. This adds `ROOT_PATH` to the beginning of the AJAX call so that no matter where it is, it will point to the correct location.

### DIFF
--- a/include/client/kb-categories.inc.php
+++ b/include/client/kb-categories.inc.php
@@ -41,7 +41,7 @@
             <li><i></i>
             <div style="margin-left:45px">
             <h4><?php echo sprintf('<a href="faq.php?cid=%d">%s %s</a>',
-                $C->getId(), Format::htmlchars($C->getFullName()),
+                $C->getId(), Format::htmlchars($C->getLocalName()),
                 $count ? "({$count})": ''
                 ); ?></h4>
             <div class="faded" style="margin:10px 0">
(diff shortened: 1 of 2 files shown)


include/client/kb-categories.inc.php
```diff
@@ -41,7 +41,7 @@
             <li><i></i>
             <div style="margin-left:45px">
             <h4><?php echo sprintf('<a href="faq.php?cid=%d">%s %s</a>',
-                $C->getId(), Format::htmlchars($C->getFullName()),
+                $C->getId(), Format::htmlchars($C->getLocalName()),
```
JediKev, line 44: @protich Would it be better for me to keep this as `getFullName()` and change `getFullName()` to return `getLocalName()` first instead of `self::getNameById`?

protich: No - that will force translation when it might not be needed. `getFullName` should return what's stored in database.

Issue #5122, opened 2019-10-24
[Issue/Question] Login


johnkalangha:
Hi, i just want to ask in login wherever it is user or staff login, If my email as space before and after the email, it will have an error access denied, i think it is need to modify like in gmail or yahoo, in the first place username and gmail/yahoo is not allowed to use space in registration, so it is okey not to read the space in login, thanks

JediKev:
@johnkalangha

I will make a pull tomorrow that will trim the rogue spaces.

Cheers.

johnkalangha:
Thank you sir

JediKev:
@johnkalangha 

Here you go:
- https://github.com/osTicket/osTicket/pull/5124

Cheers.

PR #5124, opened 2019-10-24
issue: Spaces In Username

JediKev:
This addresses issue #5122 where having a space at the beginning and/or end of the login Username/Email on the Client Portal the system will return `Access Denied`. This is due to the system not trimming the rogue spaces. This adds `trim()` around the Username/Email so that the spaces are removed and login will be successful.


johnkalangha:
Thank you sir, i will try this
    On Friday, 25 October 2019, 03:12:07 am GMT+8, Peter Rotich <notifications@github.com> wrote:  
 
 
Merged #5124 into develop.

PR #5126, opened 2019-10-25
Custom email notifications

Rossi84:
We would like to know if you can set a custom notification scheme for each team ?

### DIFF
--- a/include/class.config.php
+++ b/include/class.config.php
@@ -1188,7 +1188,7 @@ class OsticketConfig extends Config {
             // Check if Admin's IP is in the list, if not, return error
             // to avoid locking self out
             if (!in_array($vars['acl_backend'], array(0,2))) {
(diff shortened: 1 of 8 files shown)


JediKev:
You seemed to have opened a Pull Request instead of an Issue Report by accident. I will close this so there is no confusion.

Also, you're asking a question and not reporting an issue nor submitting any additions to the code. If you have questions about osTicket you should make a post on our community supported [Forum here](https://forum.osticket.com). With this being said I will give you a quick answer below:
There is no way currently to configure different Alert Templates per Team. You can however configure different Alert Templates per Department. So, for your case you can configure your Help Topics to route to specific Departments and Assign specific Teams via Help Topic so that when a ticket is created it gets routed to a specific Department, Assigns the Team, and use the Template configured for the Department.

Issue #5123, opened 2019-10-24
Attachment filters are not filtering extensions, banned attachments are still coming through


hot-steamer:
### Prerequisites

* [NA] Can you reproduce the problem in a fresh installation of the "develop" branch? Issue was reproduced by Kevin and confirmed. See this forum post:
https://forum.osticket.com/d/95746-attachments-are-not-being-filtered-in-v1-12
* [N] Do you have any errors in the PHP error log, or javascript console?
* [Y] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [Y] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

### Description

A ticket filter has been setup to reject all but a list of acceptable extensions.  However, I can send in an attachment that should have been blocked and it was allowed through.

### Steps to Reproduce

1. Enable attachments and set a few allowable extensions.
2. I sent a word doc from gmail to the ticket system's email address

**Expected behavior:**
I expected to see the attachment blocked and a note on the ticket stating such.

**Actual behavior:**
Attachment was added to the ticket.

### Versions
osTicket Version | v1.12 (a076918) —  Up to date
Web Server Software | Apache/2.2.15 (CentOS)
MySQL Version | 5.1.66
PHP Version | 5.6.40
Chrome

### Troubleshooting
For testing, I added some logging to the PHP to see where in the code it got with an attachment that should have been blocked. I can see it throw the exception when it processes the extensions, but it still adds it to the ticket.  I did not debug it past this point.

![image](https://user-images.githubusercontent.com/4528005/67499128-c7616e80-f63d-11e9-8974-01e78e7e971d.png)

JediKev:
@hot-steamer 

Here you go:
- https://github.com/osTicket/osTicket/pull/5133

Cheers.

visiniaga:
@JediKev, test with osticket 12.2.3 with php 7.2.x, 
sending attachment with extension ods, is still acceptable, while in filter only accept some file extension
![Screenshot 2019-10-29 at 06 51 20](https://user-images.githubusercontent.com/43309334/67726579-d9198c00-fa18-11e9-9966-6f16a3194f2c.png)

![Screenshot 2019-10-29 at 06 54 06](https://user-images.githubusercontent.com/43309334/67726596-f4849700-fa18-11e9-9480-4e1b92829b1b.png)

hot-steamer:
@JediKev 
[...]

PR #5133, opened 2019-10-28
issue: Attachment Filter

JediKev:
This addresses issue #5123 where fetched Emails do not reject attachments that are not allowed. This is due to setting a File ID inside of an array called `$file[]` but not unsetting it when the attachment hits an FileUploadError. This updates class MailFetch to set the file array back to empty so that the ID is cleared and the error is added correctly which rejects the attachment.

### DIFF
--- a/include/api.tickets.php
+++ b/include/api.tickets.php
@@ -91,7 +91,9 @@ class TicketApiController extends ApiController {
                     $file['id'] = $F->getId();
                 }
                 catch (FileUploadError $ex) {
-                    $file['error'] = $file['name'] . ': ' . $ex->getMessage();
+                    $name = $file['name'];
+                    $file = array();
+                    $file['error'] = $name . ': ' . $ex->getMessage();
                 }
             }
             unset($file);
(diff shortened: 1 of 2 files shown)


visiniaga:
@JediKev after apply update 9f4fbc2, this works now in osticket 12.2.3 and php 7.2.x. but still not yet block for file attachment via web upload

JediKev:
@visimitra 

When I upload a non-allowed attachment via Client Portal it denies the attachment like it's supposed to. Where exactly in the web UI is it not working? Could you provide steps to replicate + any screenshots?

Cheers.

visiniaga:
@JediKev, yes in my testing it show upload and posted successfully
Screenshot 2019-10-29 at 22 47 53
Screenshot 2019-10-29 at 22 45 51

JediKev:
@visimitra 

Ah, okay I see what you're talking about now. You are a bit confused, as the Agents are exempt from File Restrictions on purpose. This pull only applies to _User_ submitted attachments via Email, API, or UI. So this is working as intended. 👍

Cheers.

visiniaga:
@JediKev 

Ok, very well noted and thanks.

PR #5135, opened 2019-10-29
Feature: Mark as Answered permission option

RikkiMasters:
Some roles need the ability to mark a ticket as Answered/Unanswered but must not be able to reply directly to the customer. Both are controlled by the Post Reply permission. This seperates the two functions so they can be managed independently.

### DIFF
--- a/include/ajax.tickets.php
+++ b/include/ajax.tickets.php
@@ -1273,7 +1273,7 @@ function refer($tid, $target=null) {
         if (!($ticket=Ticket::lookup($tid)))
             Http::response(404, __('No such ticket'));
 
-        if (!$ticket->checkStaffPerm($thisstaff, Ticket::PERM_REPLY) && !$thisstaff->isManager())
+            if (!$ticket->checkStaffPerm($thisstaff, Ticket::PERM_REPLY) && !$ticket->checkStaffPerm($thisstaff, Ticket::PERM_MARKANSWERED) && !$thisstaff->isManager())
             Http::response(403, __('Permission denied'));
 
         $errors = array();
--- a/include/ajax.tickets.php
+++ b/include/ajax.tickets.php
@@ -1273,7 +1273,7 @@ function refer($tid, $target=null) {
         if (!($ticket=Ticket::lookup($tid)))
             Http::response(404, __('No such ticket'));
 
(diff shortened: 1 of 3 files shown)


JediKev:
@RikkiMasters 

Thank you for this but the original thinking behind this was that in order to have Mark as Answered/Unanswered you have to actually be able to answer the ticket.

Cheers.

RikkiMasters:
@JediKev 

No problem.

It was an issue for us as there were cases of staff posting internal notes as replies to customers mistakenly. We email outside of osTicket so logically we removed the box using the permission, but still needed to be able to mark as answered.

Just thought I'd add the solution here if it would be of benefit.

include/ajax.tickets.php
```diff
@@ -1273,7 +1273,7 @@ function markAs($tid, $action='') {
         if (!($ticket=Ticket::lookup($tid)))
             Http::response(404, __('No such ticket'));
 
-        if (!$ticket->checkStaffPerm($thisstaff, Ticket::PERM_REPLY) && !$thisstaff->isManager())
+            if (!$ticket->checkStaffPerm($thisstaff, Ticket::PERM_REPLY) && !$ticket->checkStaffPerm($thisstaff, Ticket::PERM_MARKANSWERED) && !$thisstaff->isManager())
```
JediKev: You should fix the spacing issue and change this to the following:
```
if (!$ticket->checkStaffPerm($thisstaff, Ticket:: PERM_MARKANSWERED) && !$thisstaff->isManager())
```

Since you are making this its own Permission there is no need to check wether or not they have Reply Permission, only the MarkAs Permission (and if they are Manager of course). The same concept as what you did with the `$canMarkAnswered` variable below.

RikkiMasters: Correct, thanks!

JediKev (approved):
Alright, looks good now 👍🏻

JediKev:
@RikkiMasters 

Just to let you know, we plan on merging this to core. 👍🏻 With this being said, I have added a [database patch](https://github.com/osTicket/osTicket/pull/5119/commits/ea1c55ff3e75ab0dcd8fa06dd230866d92f86b27) for this so we can add the Mark Answered permission by default for Roles with the Reply Permission Enabled. This is so that people used to the way it works now will not be affected. If we merged this without a patch then people who used the Mark Answered with Reply Permission will not be able to Mark Answered anymore and will freak out lol

Cheers.

RikkiMasters:
@JediKev That's great! Glad I could contribute

PR #5151, opened 2019-11-06
Ticket Merge Upgrade Patch

aydreeihn:
This patch modifies the upgrader for ticket merges to add a new table, thread_entry_merge that will store data for thread entries related to ticket merges. This modification improves the upgrade process.


include/class.thread.php
```diff
@@ -1539,25 +1555,38 @@ function lookupByRefMessageId($mid, $from) {
...
-            $this->setFlag(ThreadEntry::FLAG_CHILD, true);
-        }
-        $this->thread_id = $thread_id ?: $thread_id;
+        $this->setFlag(ThreadEntry::FLAG_CHILD, true);
+        $this->thread_id = $thread_id;
         $this->save();
     }
 
+    function getMergeData() {
+        $row = ThreadEntryMergeInfo::objects()
```
protich: Why are you doing direct lookup here? thread entry already has a relationship.

```php
return $this->merge_info ? $this->merge_info->data : null;
```

include/staff/templates/thread-entries.tmpl.php
```diff
@@ -34,6 +34,8 @@
         // TODO: Consider adding a date boundary to indicate significant
         //       changes in dates between thread items.
         foreach ($buckets as $entry) {
+            $entry = ThreadEntry::lookup($entry->id);
```
protich, line 37: Why lookup? You already have entry object. right?

aydreeihn: It's an AnnotatedModel__ThreadEntry.

include/class.thread.php
aydreeihn: That makes more sense... My bad lol

include/staff/templates/thread-entries.tmpl.php
protich: I thought it delegates to the object.

aydreeihn: It ended up the way it did because of the sortEntries function. Buckets had to end up being an array of thread entries so that the thread entries show up in the right order. When you return them as an array and then loop through them, it says they're in the AnnotatedModel__ThreadEntry class.

Issue #5155, opened 2019-11-07
Between filter behaviour


Naghal:
### Prerequisites

* [ No] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [No error ] Do you have any errors in the PHP error log, or javascript console?
* [x ] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [x ] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

### Description

The filtering using the "between" method is not as expected. 

### Steps to Reproduce

1. Do an advanced search
2. Filter any date field by "between" and select for example 2019-04-11 - 2019-05-11

**Expected behavior:** Data between the interval is filtered, as well as 2019-04-11 and 2019-05-11

**Actual behavior:**  Data between the interval is filtered, as well as 2019-04-11, but **NOT 2019-05-11**

### Versions

OsTicket | v1.12.3 (bcf1a6f) 
Logiciel serveur Web | Apache/2.2.15 (CentOS)
Version de MySQL | 5.5.52
Version PHP | 7.3.11

Running on Windows 10 on latest Chrome

TechCare-Aus:
I checked this on my system too, running the following:

> osTicket Version | v1.14-rc2-104-g3f42833c (3f42833)  
> Web Server Software | Apache
> MySQL Version | 10.3.19
> PHP Version | 7.3.11
> Timezone | AEDT (Interpreted as Australia/Melbourne)

Same issue.

Ticket came in on 04/11/19 7:14 am AWST. When I do an advanced search on create date, set to:

- 4/11/2019 to 4/11/2019. It finds nothing.
- 3/11/2019 to 4/11/2019. It finds nothing.
- 3/11/2019 to 5/11/2019. It finds the ticket.
- 4/11/2019 to 5/11/2019. It finds the ticket.

My server is hosted in GMT +10, I've set osTicket to GMT +8

Also side note, when selecting a date the option is to click (in my case) AWST. But this doesn't do anything. Should it? It is a hyperlink.

protich:
Thanks guys for the report - I was able to reproduce and make a pull request ( #5166 ) to address the issue. It's made to `develop` branch but will get merged to `v1.14` as well.

PR #5166, opened 2019-11-13
Add Time boundaries to Between date range

protich:
Add time boundaries to between data range dates and convert results to database timezone.

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -2415,14 +2415,15 @@ class DatetimeField extends FormField {
         case 'before':
             return new Q(array("{$name}__lt" => $value));
         case 'between':
-            $tz = new DateTimeZone($cfg->getTimezone());
-            $dbtz = new DateTimeZone($cfg->getDbTimezone());
             $left = Format::parseDateTime($value['left']);
             $right = Format::parseDateTime($value['right']);
+            // TODO: allow time selection for between
             $left = $left->setTime(00, 00, 00);
             $right = $right->setTime(23, 59, 59);
-            $left->setTimezone($tz)->setTimezone($dbtz);
-            $right->setTimezone($tz)->setTimezone($dbtz);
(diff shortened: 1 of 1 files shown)


protich:
Addresses issue #5155

include/class.forms.php
```diff
@@ -2415,15 +2415,17 @@ function getSearchQ($method, $value, $name=false) {
...
-                    : $value[$side];
-            }
+            $tz = new DateTimeZone($cfg->getTimezone());
+            $dbtz = new DateTimeZone($cfg->getDbTimezone());
+            $left = Format::parseDateTime($value['left']);
+            $right = Format::parseDateTime($value['right']);
+            $left = $left->setTime(00, 00, 00);
+            $right = $right->setTime(23, 59, 59);
+            $left->setTimezone($tz)->setTimezone($dbtz);
+            $right->setTimezone($tz)->setTimezone($dbtz);
```
protich: I'm not so sure we should convert time to user's timezone. 

I added it because - between widget doesn't provide for opportunity specify timezone but I'll be interested in y'alls feedback.

JediKev: I think it's a good idea as the Date Picker is shown in the user's timezone. So if they chose between `11/11/2019` and `11/12/2019` in the date picker they would expect the system to show results in that exact timeframe in their timezone. I tested this pull and it works as expected. It pulls the correct data.

protich: The issue is mute when it comes to the agent doing Advanced Search - the date picker already registers the user's timezone as part of the entry. So we don't actually need to convert to user's timezone in this case.

My concern is when `Between` criteria is used in a global queue. The saved values will be on the queue creator's timezone - forcing conversion to the agent timezone might not be the intended results -  I added it anyhow because `Period` floats the timezone, but it's not specific tho.

PR #4680, opened 2019-01-30
PHP 7.2 Compatibility

aydreeihn:
This makes osTicket compatible with PHP versions 5.6 - 7.2

PR #5182, opened 2019-11-16
Update class.api.php, Changed Signature of parse()

mburamorde:
This resolves an php 7.2 error!

Warning:  Declaration of ApiJsonDataParser::parse($stream) should be compatible with JsonDataParser::parse($stream, $tidy = false) in /home/xxx/www.xxx.de/include/class.api.php on line 0

### DIFF
--- a/include/class.api.php
+++ b/include/class.api.php
@@ -427,8 +427,8 @@ class ApiJsonDataParser extends JsonDataParser {
 include_once "class.mailparse.php";
 class ApiEmailDataParser extends EmailDataParser {
 
-    function parse($stream) {
-        return $this->fixup(parent::parse($stream));
+    function parse($stream, $tidy = false) {
+        return $this->fixup(parent::parse($stream, $tidy = false));
     }
 
     function fixup($data) {


include/class.api.php
```diff
@@ -427,8 +427,8 @@ function fixup($current) {
 include_once "class.mailparse.php";
 class ApiEmailDataParser extends EmailDataParser {
 
-    function parse($stream) {
-        return $this->fixup(parent::parse($stream));
+    function parse($stream, $tidy = false) {
+        return $this->fixup(parent::parse($stream, $tidy = false));
```
JediKev, line 431: ~Can you update this to be:~
```
function parse($stream, $tidy=false) {
    return $this->fixup(parent::parse($stream, $tidy));
```

~Thank you~

**Edit:**
See comment below.

JediKev: @mburamorde

I just expanded the code block and see that you made this change to class `ApiEmailDataParser` which is not class `ApiJsonDataParser` as stated in the Warning message. When you look at class `ApiJsonDataParser` you will see that the `parse()` method already has `$tidy` declared as an argument with the default being `false`. Since `ApiJsonDataParser::parse()` is compatible with `JsonDataParser::parse()`, I'm not really sure why you are receiving this Warning unless you are using an older version of osTicket with [this pull](https://github.com/osTicket/osTicket/pull/4680) not merged yet. I will close your pull request as this Warning has already been taken care of in the referenced pull above.

**Note:**
As a precaution I looked at class `EmailDataParser` to see if it's `parse()` method had `$tidy` declared as an argument but it does not so `ApiEmailDataParser` does not need `$tidy` declared as an argument.

Cheers.

Issue #5194, opened 2019-11-19
Feature request: ability to change parent ticket's status when merging


FinThunderstorm:
Feature Request:
Ability to change parent ticket's status when merging new ticket to it. Now if I merge new ticket with old closed one, the new ticket goes to old one and the old stays closed (= not changing status). In my case we get new emails about the same thing. So it would be much more quicker if we don't need to try to search the old one from closed queue and then change it to open.

PR #5203, opened 2019-11-20
Ticket Merge Parent Status

aydreeihn:
This commit gives agents the option to change the status of the Parent ticket when doing a merge. To do this, they must choose the desired status for 'Parent Status'. If no status is chosen, the Parent Ticket will remain in its current status.

Screen Shot 2019-11-21 at 9 22 32 AM

This change was requested in Issue #5194.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -2555,6 +2555,9 @@ implements RestrictedAccess, Threadable, Searchable {
                 $child->setMergeType($options['combine']);
                 $child->setStatus(intval($options['statusId']), false, $errors, true, true); //force close status for children
 
+                if ($options['parentStatusId'])
+                    $parent->setStatus(intval($options['parentStatusId']));
+
                 if ($options['delete-child'] || $options['move-tasks']) {
                     if ($tasks = Task::objects()
                         ->filter(array('object_id' => $child->getId()))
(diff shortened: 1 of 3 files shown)


protich:
Why the requirement to check "Set Parent Status"? I was thinking if you select a new status then that indicates the change.

aydreeihn:
> Why the requirement to check "Set Parent Status"? I was thinking if you select a new status then that indicates the change.

It probably helps to mention that the Parent Status dropdown only shows up if 'Set Parent Status' is checked.

We wouldn't be able to tell for certain which Ticket is the Parent on initial load since whatever Ticket is at the top of the list is the Parent, so I wouldn't be able to pick the actual parent ticket's status out of the list either. With the checkbox, I feel like the decision to change the parent ticket status is also more deliberate than a status option already being on the page.

JediKev:
New picture please

aydreeihn:
Looking back at this, I actually went and updated the picture after I made the change. Before, I was doing this with a checkbox that would show the parent status dropdown if you clicked it rather than just having the dropdown on the modal already with the --Select-- value if you didn't actually want to change the Parent status.

aydreeihn:
@JediKev rebased and ready to merge.

PR #5204, opened 2019-11-20
Inline Edit Fields With Data Integrity

aydreeihn:
This commit adds the 'Required to close tickets' warning to inline edit fields if they are configured with the Data Integrity box checked (Require entry to close a thread)

Required Fields Without Data:
Screen Shot 2019-11-21 at 9 29 38 AM

Required Fields With Data:
Screen Shot 2019-11-20 at 3 09 15 PM

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -1332,7 +1332,6 @@ class FormField {
     }
 
     function getEditForm($source=null) {
-
         $fields = array(
                 'field' => $this,
                 'comments' => new TextareaField(array(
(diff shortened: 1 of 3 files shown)


aydreeihn:
This adds the 'Required to close ticket' as the hint of the field, but it bothers me that it doesn't add it inline. It puts it under the field.

Screen Shot 2019-11-20 at 2 30 51 PM

protich:
I think it also make sense to add the hint to the field on ticket view - if it doesn't have data.

JediKev:
@aydreeihn 

Hints are meant to be under the title. For example:
Screen Shot 2019-11-20 at 14 48 17

If you want it inline I believe you'll have to put it in the HTML instead somehow.

Cheers.

aydreeihn:
Modified. Just need to update it again to not show the icon for checkboxes for both inline and ticket-edit pages.

PR #5207, opened 2019-11-21
queues: Revert commit cedd6121

protich:
This PR reverts commit cedd6121 because it broke adhoc search queue naming (describe) shows on Recent Searches section.

The spirit of the initial change was correct but support of complex criteria wasn't needed after all due to difficulty supporting multiple organizations - something we will revisit in the future.


protich:
Wasn't partial after all - meant it to be but it's all good fully reverted as well.

PR #5216, opened 2019-11-25
Authentication for usernames with "\" not working without this fix

EIJason:
Osticket developers developed the possiblity of allowing you to logon to a mailbox and check a shared mailbox that mailbox it has permission on... but how they do this is look for the '\'in the Username Edit box and anything before it is the Auth User and After is the Shared mailbox User....

The best fix would be a new edit box  for the shared mailbox... but as that doesn't exist what I do is check the username for '\' character  then to accommodate the username I connect with "Server\Account"  I split the Values by the slashes as the original does but I check if there are three values, the third being the shared mailbox email address, the second I check for an @ sign if it exists its a shared mailbox, if not I append the first and the second with a slash to get the username I desire...

I recommend this fix to the Developers... Or putting in a separate field for the share mailbox.

### DIFF
--- a/include/class.mailfetch.php
+++ b/include/class.mailfetch.php
@@ -55,12 +55,24 @@ class MailFetcher {
             // Support Exchange shared mailbox auth
             // (eg. user@domain.com\shared@domain.com)
             $usernames = explode('\\', $this->ht['username'], 2);
-            if (count($usernames) == 2) {
-                $this->authuser = $usernames[0];
-                $this->username = $usernames[1];
-            } else {
-                $this->username = $this->ht['username'];
-            }
+           if (count($usernames) == 3) 
+           {
+                $this->authuser = $usernames[0]."\\".$usernames[1];
+                $this->username = $usernames[2];
+           } 
+	   else 
+		if (count($usernames) == 2) 
+		{
+			$this->authuser = $usernames[0];
+			if(strpos($usernames[1],'@'))
+				$this->username = $usernames[1];
+			else
+				$this->authuser = $usernames[0]."\\".$usernames[1];
+		}
+		else
+		{
+			$this->username = $this->ht['username'];
+		}
 
             if(!strcasecmp($this->ht['protocol'],'pop')) //force pop3
                 $this->ht['protocol'] = 'pop3';


JediKev:
@EIJason 

I asked on the Forum what use-case do you have a backwards slash in the name other than shared auth, but have received no response yet.

Cheers.

EIJason:
Sorry I been sick have not seen the forum yet…

We have an exchange server that is built into an external hosted desktop / Cloud environment…
I cannot logon to the Imap with just the email Address, I think this is restricted for security reasons…

And the only way to logon to this is with the Server followed by the actual user on that server…
(This cloud service is an external outsourced IT company and these are the credentials they provided for logon as below)


So below is how we logon… server within the cloud environment then the usernames…

[cid:image001.png@01D5A3A6.A4AF05D0]

Regards
Jason

JediKev:
@EIJason 

Could you please post the picture on Github? It didn't attach and I'm wondering if you are using a domain name for the server? (ie. the first part of your auth string)

Cheers.

EIJason:
image

JediKev:
@EIJason 

Ooo, I've never seen this before (for email at least). Very specific use-case but worth preserving. I'll mess around with your pull to see what I can come up with.

Cheers.

EIJason:
If I can make a suggestion? Why not Just add a separate edit box for shared mailbox this would resolve any variations on how people logon to servers ;)

On another note I also modified the the containers for Osticket from 950px to 98%, why is the screen area limited in Osticket? I also Edited the redactor To 100% for its container so it looks more like this:

image

image

SO we can use all the screen real estate...

JediKev:
@EIJason 

> If I can make a suggestion? Why not Just add a separate edit box for shared mailbox this would resolve any variations on how people logon to servers ;)

We already have many text boxes on that page and didn't want to add more to the chaos, PLUS this is how many people are used to logging in with shared auth in other applications eg. `auth_user\shared_user` (ie. using only the username box for the auth string). So it's best to keep as is. Your use case is very specific and I don't think I've seen it elsewhere before.

>On another note I also modified the the containers for Osticket from 950px to 98%, why is the screen area limited in Osticket? I also Edited the redactor To 100% for its container so it looks more like this:

This is not the thread to discuss this topic. We will have fully responsive theme coming in v2.0 so stay tuned.

Cheers.

EIJason:
Ok well my code above makes allowance for both cases...

I only changed 3 values in the two css files to allow a larger view... would you like me to add the changes as I did on github? Whats the expected release date of osticket2? It could be useful to others while the anticipated new version is in development... ;)

JediKev:
@EIJason 

>Ok well my code above makes allowance for both cases...

But it adds some other stuff I'm not too sure about. I will have to pull and test locally to determine.

>Whats the expected release date of osticket2?

No set release date and no set development date just yet. It will be called osTicket v2.0.

>would you like me to add the changes as I did on github? ... It could be useful to others while the anticipated new version is in development... ;)

You can if you want but people have already done this many...many times before here and on the Forum. There are even themes available but they are not recommended as they were not developed by us nor affiliated with us.

Cheers.

EIJason:
Yeah And I have tried some themes before but you never know if its going to work on your newer versions hence why I only make small changes now to allow osticket more screen real estate... I am anxiously awaiting your responsive version.... ;)

Issue #5215, opened 2019-11-24
links not functioning (process install completed)


balojs:
### Prerequisites

* [x] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [ ] Do you have any errors in the PHP error log, or javascript console?
* [ ] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [ ] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/osTicket/osTicket/wiki/Github-Issues-Guidelines)

### Description
links osTicket Forums and  osTicket Documentation not functioning 

### Steps to Reproduce

1. Install Osticket v1.12.5 (933bb1f) 
2. Click on https://forum.osticket.com/
3. Click on https://docs.osticket.com/

**Expected behavior:** [What you expected to happen]
   https://forum.osticket.com/
    https://docs.osticket.com/

**Actual behavior:** [What actually happened]
![image](https://user-images.githubusercontent.com/3303352/69501759-a787c980-0ecd-11ea-87f6-442cd1a61f92.png)

### Versions

Osticket v1.12.5 (933bb1f) 

Also, please include the OS and what version of the OS you're running. As well as your browser and browser version.
N/A

URL
http://localhost/osticket12/setup/install.php#

![image](https://user-images.githubusercontent.com/3303352/69501585-dc931c80-0ecb-11ea-8c9c-24bc0cfd9a3d.png)

JediKev:
@balojs 

Looks like it’s the pound symbols as the href. Will make a pull tomorrow. Thanks for the report.

Cheers.

JediKev:
@balojs 

Here you go:
- https://github.com/osTicket/osTicket/pull/5217

Thanks again for the report!

Cheers.

balojs:
Cool @JediKev 💯 thanks my friend

PR #5217, opened 2019-11-25
install: Forum and Docs Links

JediKev:
This addresses issue #5215 where the Post-Install page references the Forum and Docs site but the link doesn't work. This is due to the pound symbols being the `href`. This adds the links to the `href` attribute so the links actually link to the sites.

### DIFF
--- a/setup/inc/install-done.inc.php
+++ b/setup/inc/install-done.inc.php
@@ -31,11 +31,11 @@ $url=URL;
                 <tr>
                     <td width="50%">
                         <strong><?php echo __('osTicket Forums');?>:</strong><Br>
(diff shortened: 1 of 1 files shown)


setup/inc/install-done.inc.php
```diff
@@ -31,11 +31,11 @@
                 <tr>
                     <td width="50%">
                         <strong><?php echo __('osTicket Forums');?>:</strong><Br>
-                        <a href="#">https://forum.osticket.com/</a>
+                        <a href="https://forum.osticket.com/">https://forum.osticket.com/</a>
```
JediKev: target = _blank

JediKev:
Updated

PR #5220, opened 2019-11-25
Overide the redactor to autosize to the full width of the container (…

EIJason:
…allows the use of more screen real estate)

Override the max width of redactor limitation... so resizing your container redactor sizes accordingly....

image

### DIFF
--- a/css/redactor.css
+++ b/css/redactor.css
@@ -1041,6 +1041,7 @@
 
 .redactor-styles {
   margin: 0;
+  max-width:100% !important;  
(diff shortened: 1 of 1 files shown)


JediKev:
@EIJason

You should’ve combined the two commits into one pull but keep commits separate so it’s not taking up two pulls.

Cheers.

EIJason:
How do I do that? New to this GitHub only figuring it out bit by bit…

mburamorde (approved):

PR #5222, opened 2019-11-25
Ticket Merge/Link Defaults

aydreeihn:
This commit makes it so that the Ticket Merge and Link roles are enabled by default on fresh installs.


protich:
Why not role 2 & 3?

aydreeihn:
@protich Just added them for the other roles as well.

PR #5226, opened 2019-11-26
issue: Attachment Upload Configuration

JediKev:
This addresses an issue where we pull the `first()` Form Field configuration for attachments instead of determining if we are looking for the Ticket Attachment Configurations or the Task Attachment Configurations. For example, if you are creating a Ticket as a User and go to add an attachment if somehow your Task Details Form comes up first in the configuration lookup query we will apply the Task Attachment Configs instead of the Ticket Attachment Configs which is totally incorrect. This updates the lookup for attachment configurations on web uploads to be specific about what it’s looking for so it can get the correct configuration.

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -344,6 +344,8 @@ class Form {
 class SimpleForm extends Form {
     function __construct($fields=array(), $source=null, $options=array()) {
         parent::__construct($source, $options);
+        if (isset($options['type']))
+            $this->type = $options['type'];
         $this->setFields($fields);
     }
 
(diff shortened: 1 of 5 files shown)


include/class.forms.php
```diff
@@ -4645,6 +4645,18 @@ function($t) { return strpos($t, '/') !== false; }
                 );
             }
         }
+        $field_id = $this->field->get('id');
+        // Determine if for Ticket/Task/Custom
+        $object = DynamicForm::objects()
```
JediKev: Instead of `DynamicForm::objects()` do `$this->field->getForm()->getType()`

JediKev:
Updated

include/class.forms.php
```diff
@@ -342,8 +342,10 @@ static function instantiate() {
  *
  */
 class SimpleForm extends Form {
-    function __construct($fields=array(), $source=null, $options=array()) {
+    function __construct($fields=array(), $source=null, $options=array(), $type=null) {
```
JediKev: add type to options

JediKev:
Updated

PR #5228, opened 2019-11-26
support: IE Discontinued

JediKev:
Well folks, the time has finally come to lay IE in it's final resting place. Twas a long, hard battle with IE but in the end it was defeated and it's kingdom taken over by it's successor, Edge. Now there is a new king and IE must be banished from the lands he once ruled, forever. It's sad to see such a prolific leader fall to the ashes (not lol). Alright, let’s get serious for a moment. Microsoft is discontinuing IE at the end of January 2020 so it’s time to start showing a banner that says "osTicket no longer supports IE." This message is shown on the Installer, the Upgrader, in the Client Portal, and in the Staff Panel when using IE. This checks the User-Agent header for IE’s specific header and if detected will show the banner. This will not show on Edge, Chrome, Firefox, etc. only when using IE. We came to this conclusion after seeing sites like Github (a microsoft product) showing "no longer supported" banners for IE (a microsoft product).

#### Installer
Install

#### Client Portal
Client

#### Staff Panel
Staff

#### Upgrader
Upgrader


Demarcation:
i would question if this is required until such time as osticket impliments code base that prevents it working correctly in IE.

Untill then, maybe a warning in the install/upgrader that it may not support IE in future.

JediKev:
@Demarcation

Well there are people experiencing strange issues with IE currently so we are going to put the banner now so everyone can start making the switch. Why not start making the move early? GitHub is like that currently and it’s a microsoft product.

**Note:**
The code is open source so if would like to hold off on the warning you could remove the lines from your codebase.

Cheers.

TechCare-Aus:
@JediKev brilliant narrative mate!

JediKev:
- ~Check if installer has $ost set.~ It is not set but you can use `osTicket::is_ie()` directly without accessing `$ost`.
- ~Set this to a function in class osTicket and reuse `is_ie()`.~ Done.

JediKev:
Updated

include/client/header.inc.php
```diff
@@ -86,6 +90,14 @@
 </head>
 <body>
     <div id="container">
+        <?php
+        if($ost->getError())
+            echo sprintf('<div id="error_bar">%s</div>', $ost->getError());
```
JediKev: make classes and update css to match

PR #5230, opened 2019-11-27
allow for yyyy year notation in custom date format

Demarcation:
This allows the settings of custom dates to use the 'yyyy' year notation for a 4 digit year instead of 'y' which is slightly less intuative.

Without this patch, using the yyyy year notation in custom configurations will result in the scp dashboard reports date selector switching from 2019 to 2020 as the system reads the first two digits of the returned value.

### DIFF
--- a/include/class.report.php
+++ b/include/class.report.php
@@ -43,8 +43,8 @@ class OverviewReport {
         $format =  $format ?: $this->format;
         if ($translate) {
             $format = str_replace(
-                    array('y', 'Y', 'm'),
-                    array('yy', 'yyyy', 'mm'),
+                    array('yyyy','y', 'Y', 'm'),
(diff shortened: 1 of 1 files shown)


JediKev:
@Demarcation 

I do not see the same issues on v1.14.1. Below is what I see:
#### Configuration
Screen Shot 2019-11-27 at 09 38 14

#### Queues
Screen Shot 2019-11-27 at 09 38 56

#### Dashboard
Screen Shot 2019-11-27 at 09 40 04

#### Date Picker
Screen Shot 2019-11-27 at 09 40 14

Cheers.

Demarcation:
in the scp dashboard and in the date picker select a date, then run the report, then go back and select  a second date, the second date defaults to 2020 not 2019, as the function edited in this commit doesnt understand a yyyy format.

PR #5233, opened 2019-11-27
Increase Screen Area on client and Agent Panel

EIJason:
These changes make use of most of the screen, so no wasted areas on the side also added a bootstrap to the client header to improve the look and feel slightly...

image

image

### DIFF
--- a/include/client/header.inc.php
+++ b/include/client/header.inc.php
@@ -31,20 +31,20 @@ if ($lang) {
     <meta name="description" content="customer support platform">
     <meta name="keywords" content="osTicket, Customer support system, support ticket system">
     <meta name="viewport" content="width=device-width, initial-scale=1">
-	<link rel="stylesheet" href="<?php echo ROOT_PATH; ?>css/osticket.css" media="screen"/>
-    <link rel="stylesheet" href="<?php echo ASSETS_PATH; ?>css/theme.css" media="screen"/>
-    <link rel="stylesheet" href="<?php echo ASSETS_PATH; ?>css/print.css" media="print"/>
+	<link rel="stylesheet" href="<?php echo ROOT_PATH; ?>css/osticket.css" media="screen">
+    <link rel="stylesheet" href="<?php echo ASSETS_PATH; ?>css/theme.css" media="screen">
+    <link rel="stylesheet" href="<?php echo ASSETS_PATH; ?>css/print.css" media="print">
     <link rel="stylesheet" href="<?php echo ROOT_PATH; ?>scp/css/typeahead.css"
          media="screen" />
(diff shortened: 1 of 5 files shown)


include/client/header.inc.php
```diff
@@ -31,33 +31,36 @@
...
-    <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/jquery-ui-1.12.1.custom.min.js"></script>
-    <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/jquery-ui-timepicker-addon.js"></script>
-    <script src="<?php echo ROOT_PATH; ?>js/osticket.js"></script>
-    <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/filedrop.field.js"></script>
-    <script src="<?php echo ROOT_PATH; ?>scp/js/bootstrap-typeahead.js"></script>
-    <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/redactor.min.js"></script>
-    <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/redactor-plugins.js"></script>
-    <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/redactor-osticket.js"></script>
-    <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/select2.min.js"></script>
+    <script type="text/javascript" src="<?php echo ROOT_PATH; ?>js/jquery-3.4.0.min.js?f1e9e88"></script>
```
JediKev: What's up with `?f1e9e88` after all of the URLs..?

EIJason:
I am not sure where that came from I fixed it though?

EIJason (commented):
Ok this is my proposed changes both for the interface expansion and the mailfetch if username is in the format Server\Username

EIJason:
Hi JediKev,

I have a small request, can you add a Job title field into the agents that can be put on the response templates as the agent name can be?

I am using a “ – “ in the Surname and adding it following that, but the field isn’t long enough for some….

Thanks
Jason

JediKev:
@EIJason

You mean like adding a Custom Form Field and then using the form field variable in the response template? That’s already a thing via Admin Panel > Manage > Forms and Admin Panel > Emails > Templates.

Cheers.

EIJason:
Nope I mean When adding Agents I Add a job title to the Agents Surname following a Hyphen...

As we cannot add custom fields to the Agent Creation Dialogue correct? I checked manage forms…

What is preferable is a job title in the Create / Edit Agent Form

Then the ability to add that Agents name (as you can now) and title to a template (agents title not available yet)

Does that make sense?

Regards
Jason

EIJason:
image

image

PR #5239, opened 2019-12-02
emoji: Strip From Subject

JediKev:
This addresses an issue reported on the Forum where emojis are not stripped from the subject line which causes either a blank subject or half-rendered subject. This adds `Format::strip_emoticons` to TextboxField::parse in class forms so all Textbox Fields (including Issue Summary) is stripped of Emojis (since we can't handle them at all). This is also added to ThreadEntry::create in class thread so that the ThreadEntry Titles have Emojis stripped as well.

This addresses an issue where we forgot about Emojis in the Issue Details of Tickets created via API. This adds `Format::strip_emoticons` to class api so that Emojis are stripped from the Issue Details for Tickets created via API.

### DIFF
--- a/include/class.api.php
+++ b/include/class.api.php
@@ -396,7 +396,7 @@ class ApiJsonDataParser extends JsonDataParser {
                 $value = (bool)$value;
             } elseif ($key == "message") {
                 // Allow message specified in RFC 2397 format
-                $data = Format::parseRfc2397($value, 'utf-8');
+                $data = Format::strip_emoticons(Format::parseRfc2397($value, 'utf-8'));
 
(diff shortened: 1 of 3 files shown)


include/class.api.php
```diff
@@ -396,7 +396,7 @@ function fixup($current) {
                 $value = (bool)$value;
             } elseif ($key == "message") {
                 // Allow message specified in RFC 2397 format
-                $data = Format::parseRfc2397($value, 'utf-8');
+                $data = Format::parseRfc2397(Format::strip_emoticons($value), 'utf-8');
```
JediKev: Retest with `strip_emoticons` outside `Rfc`.

PR #5240, opened 2019-12-03
Issue: Help Topic Number Format

aydreeihn:
This commit fixes an issue where we did not clear the FLAG_CUSTOM_NUMBERS flag when setting the ticket number format back to System Default for Help Topics. We also needed to set the number format to NULL rather than '' when clearing it out.

### DIFF
--- a/include/class.topic.php
+++ b/include/class.topic.php
@@ -453,8 +453,8 @@ implements TemplateVariable, Searchable {
         $this->isactive = $vars['isactive'];
         $this->ispublic = $vars['ispublic'];
         $this->sequence_id = $vars['custom-numbers'] ? $vars['sequence_id'] : 0;
-        $this->number_format = $vars['custom-numbers'] ? $vars['number_format'] : '';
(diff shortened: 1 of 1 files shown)


include/class.topic.php
```diff
@@ -453,8 +453,8 @@ function update($vars, &$errors) {
         $this->isactive = $vars['isactive'];
         $this->ispublic = $vars['ispublic'];
         $this->sequence_id = $vars['custom-numbers'] ? $vars['sequence_id'] : 0;
-        $this->number_format = $vars['custom-numbers'] ? $vars['number_format'] : '';
-        $this->flags = $vars['custom-numbers'] ? self::FLAG_CUSTOM_NUMBERS : $this->flags;
+        $this->number_format = $vars['number_format'];
+        $vars['custom-numbers'] ? $this->setFlag(self::FLAG_CUSTOM_NUMBERS, true) : $this->setFlag(self::FLAG_CUSTOM_NUMBERS, false);
```
JediKev:
```php
$this->setFlag(self::FLAG_CUSTOM_NUMBERS, ($vars['custom-numbers']));
```

PR #5241, opened 2019-12-03
Issue: Mass Process Add Users to Orgs

aydreeihn:
This commit fixes an issue where we could not add multiple users to an organization at a time due to an issue with trying to pass an object to parseJSON rather than a string.

If any of the selected users are already part of a different organization, their organization will NOT be changed.

It also fixes the Organization typeahead to search through parts of an organization name rather than having to type out the whole thing before returning the organization.

### DIFF
--- a/include/ajax.orgs.php
+++ b/include/ajax.orgs.php
@@ -20,7 +20,7 @@ include_once(INCLUDE_DIR.'class.ticket.php');
 
 class OrgsAjaxAPI extends AjaxController {
 
-    function search($type = null) {
+    function search($type = null, $fulltext=false) {
 
         if(!isset($_REQUEST['q'])) {
             Http::response(400, 'Query argument is required');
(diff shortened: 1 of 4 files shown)


include/ajax.orgs.php
```diff
@@ -39,15 +39,23 @@ function search($type = null) {
             ->values_flat('id', 'name')
             ->limit($limit);
 
-        global $ost;
-        $orgs = $ost->searcher->find($q, $orgs);
-        $orgs->order_by(new SqlCode('__relevance__'), QuerySet::DESC)
-            ->distinct('id');
+        if ($fulltext) {
```
aydreeihn, line 42: Not sure if we even need this if statement to be honest...

JediKev:
Fixed with: https://github.com/osTicket/osTicket/pull/5437

PR #5245, opened 2019-12-04
issue: Department Disable Auto Claim

JediKev:
This addresses an issue reported on the Forum where unchecking the option `disable_auto_claim` in 1.14.x and saving will not actually uncheck the setting. This is due to a check introduced with (7cfc062) that sets the `disable_auto_claim` variable to an integer of `1` or `0`. This updates the `setFlag` method for `disable_auto_claim` to check if it's equal to `1`, if so it returns `true` otherwise it returns `false`.


scarcamo-eyg:
Same issue, changes applied and Ok, thanks

PR #5248, opened 2019-12-04
Issue: Auto Referral

aydreeihn:
This commit fixes an issue we had with Auto Referrals. If a Ticket had been transfered/responded to by 3 or more departments and a user sent in an email response to one of the transferred departments (not including the original or current department), we created an auto referral for that department which could potentially give agents to access they should not have access to.

A new column, email_id, has been added to the thread_entry_email table so that we can see which system email each thread entry has been fetched from.

This commit also adds the event (EVENT_TABLE) relationship to the ThreadEvents (THREAD_EVENT_TABLE) class.

### DIFF
--- a/include/class.thread.php
+++ b/include/class.thread.php
@@ -1319,12 +1319,14 @@ implements TemplateVariable {
 
     /* static */
     function logEmailHeaders($id, $mid, $header=false) {
+        $headerInfo = Mail_Parse::splitHeaders($header);
 
         if (!$id || !$mid)
             return false;
 
         $this->email_info = new ThreadEntryEmailInfo(array(
             'thread_entry_id' => $id,
+            'email_id' => Email::getIdByEmail($headerInfo['Delivered-To']),
             'mid' => $mid,
         ));
 
@@ -1996,6 +1998,12 @@ class ThreadEvent extends VerySimpleModel {
             'topic' => array(
                 'constraint' => array(
                     'topic_id' => 'Topic.topic_id',
+                ),
+                'null' => true,
+            ),
+            'event' => array(
+                'constraint' => array(
+                    'event_id' => 'Event.id',
                 ),
                 'null' => true,
             ),
--- a/include/upgrader/streams/core/4bd47d94-e7038ce9.patch.sql
+++ b/include/upgrader/streams/core/4bd47d94-e7038ce9.patch.sql
@@ -1,0 +1,24 @@
+/**
+ * @signature e7038ce9762843b6cfbe9fa2d98feaf3
+ * @version v1.15
+ * @title Add email_id to thread_entry_email
+ *
+ * This patch adds an email_id column to the thread_entry_email table
+ */
+ -- Create a blank temporary table with the new email_id column
+ CREATE TABLE `%TABLE_PREFIX%thread_entry_email_new` (
+  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
+  `thread_entry_id` int(11) unsigned NOT NULL,
+  `email_id` int(11) unsigned DEFAULT NULL,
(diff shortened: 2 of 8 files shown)


include/upgrader/streams/core/4bd47d94-e7038ce9.patch.sql
```diff
@@ -0,0 +1,27 @@
...
+  `id` int(11) unsigned NOT NULL AUTO_INCREMENT,
+  `thread_entry_id` int(11) unsigned NOT NULL,
+  `email_id` int(11) unsigned DEFAULT NULL,
+  `mid` varchar(255) NOT NULL,
+  `headers` text,
+  PRIMARY KEY (`id`),
+  KEY `thread_entry_id` (`thread_entry_id`),
+  KEY `mid` (`mid`),
+  KEY `email_id` (`email_Id`)
+) ENGINE=InnoDB AUTO_INCREMENT=84 DEFAULT CHARSET=utf8;
```
protich: I don't believe we need  `AUTO_INCREMENT`

include/upgrader/streams/core/4bd47d94-e7038ce9.patch.sql
```diff
@@ -0,0 +1,27 @@
...
+
+-- Add ticket_pid index to ticket table
+ALTER TABLE `%TABLE_PREFIX%ticket` ADD KEY `ticket_pid` (`ticket_pid`);
```
protich: A little worried how long this alter might take for large datasets.

aydreeihn: Should this be moved to Kevin's patch with the other indexes?

aydreeihn:
Alter table has been removed and the PR is rebased.

include/class.thread.php
```diff
@@ -1319,12 +1319,14 @@ function saveEmailInfo($vars) {
 
     /* static */
     function logEmailHeaders($id, $mid, $header=false) {
+        $headerInfo = Mail_Parse::splitHeaders($header);
```
protich, line 1322: You probably should parse header when ready to use.  We can potentially return without using it.

include/upgrader/streams/core/4bd47d94-e7038ce9.task.php
```diff
@@ -0,0 +1,154 @@
...
+         $ost->logDebug("Thread Entry Email Migration", 'Found '.$count
+             .' records to migrate', true);
+
+         if($count == 0)
+             return 0;  //Nothing else to do!!
+
+         // Get id of the next record to be inserted into the new table
+         $start = db_result(db_query("SELECT id
+         FROM ". THREAD_ENTRY_EMAIL_TABLE ."
+         WHERE id > (SELECT MAX(id) FROM ". THREAD_ENTRY_EMAIL_NEW_TABLE .")
```
protich, line 88: I would think `SELECT id ... ORDER BY id DESC LIMIT 1` would be faster - please test.

include/upgrader/streams/core/4bd47d94-e7038ce9.task.php
```diff
@@ -0,0 +1,154 @@
...
+         global $cfg, $ost;
+
+         # Since the queue is persistent - we want to make sure we get to empty
+         # before we find more events.
+         if(($qc=$this->getQueueLength()))
+             return $qc;
+
+         // Count how many rows are left
+         $sql = "SELECT COUNT(old.id)
+            FROM ". THREAD_ENTRY_EMAIL_TABLE." old
```
protich, line 69: Not sure if subquery is fast in this case. Why not LEFT JOIN?

PR #5251, opened 2019-12-05
Issue: Inline Edit Long Answer

aydreeihn:
This commit fixes an issue we had with saving an empty long answer field using inline edit. Before, the field would save as an empty line break which would then technically be set and saved, displaying nothing in the field, which prevented agents from doing another inline edit on the long answer field. This is most likely do to the redactor update that makes all the lines double spaced now.

### DIFF
--- a/include/ajax.tickets.php
+++ b/include/ajax.tickets.php
@@ -635,12 +635,16 @@ class TicketsAjaxAPI extends AjaxController {
                     case $field instanceof DepartmentField:
                         $clean = (string) Dept::lookup($field->getClean());
                         break;
+                    case $field instanceof TextareaField:
+                        $clean =  (string) $field->getClean();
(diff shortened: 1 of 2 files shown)


include/ajax.tickets.php
```diff
@@ -635,12 +635,16 @@ function editField($tid, $fid) {
                     case $field instanceof DepartmentField:
                         $clean = (string) Dept::lookup($field->getClean());
                         break;
+                    case $field instanceof TextareaField:
+                        $clean =  (string) $field->getClean();
+                        $clean = ($clean && ($clean != '<p><br /></p>')) ? $clean : '&mdash;' . __('Empty') .  '&mdash;';
```
JediKev: Instead, use `Format::strip_tags` and if null or empty then return EMPTY otherwise use clean.

PR #5258, opened 2019-12-09
Issue: Saving Ticket Cdata

aydreeihn:
This commit fixes an issue where we did not properly save some Cdata values which caused some fields to not export correctly.


aydreeihn:
@JediKev @protich I think I finally narrowed down the discrepancies between how the fields show up in the ticket queue vs the export.

protich:
Please test tickets page with a column added.

aydreeihn:
It looks like the columns display fine, but I did notice one thing just now that seems to happen even in our regular codebase. When tickets are being fetched into the helpdesk and added to the cdata table, the values are going in as blank rather than null, so when I did a search on say a custom Priority field, it showed tickets that actually had that value as well as any tickets that the system had just fetched since I logged on.

PR #5270, opened 2019-12-11
feature: Force HTTPS

JediKev:
This feature introduces a checkbox in the System Settings to force HTTPS. This is for people who either can't or don't know how to configure redirect rules via the webserver. If the setting is enabled, the system will redirect all osTicket web traffic to the HTTPS protocol automatically. In addition this adds an `HTTP_OK` option that disables the HTTPS redirect for local calls such as cron, pipe, and cli. This also updates the remote piping scripts (`automail.pl` and `automail.php`) to include support for HTTPS.

### DIFF
--- a/main.inc.php
+++ b/main.inc.php
@@ -33,6 +33,15 @@ $_SERVER['REMOTE_ADDR'] = osTicket::get_client_ip();
 if(!($ost=osTicket::start()) || !($cfg = $ost->getConfig()))
 Bootstrap::croak(__('Unable to load config info from DB.').' '.__('Get technical help!'));
 
+if ($cfg && $cfg->forceHttps()
+        && !osTicket::is_cli()
+        && !osTicket::is_https()) {
+    if ($_SERVER['REQUEST_METHOD'] !== 'GET')
+        Http::response(400, 'HTTPS Protocol Required');
+
+    Http::redirect('https://'.$_SERVER['HTTP_HOST'].$_SERVER['REQUEST_URI']);
+}
+
 //Init
 $session = $ost->getSession();
 
--- a/setup/scripts/automail.php
+++ b/setup/scripts/automail.php
@@ -6,7 +6,7 @@
     PHP script used for remote email piping...same as as the perl version.
 
     Peter Rotich <peter@osticket.com>
-    Copyright (c)  2006-2013 osTicket
+    Copyright (c)  2006-2020 osTicket
     http://www.osticket.com
 
     Released under the GNU General Public License WITHOUT ANY WARRANTY.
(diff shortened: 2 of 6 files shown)


JediKev:
~Review remote piping `automail.pl` script.~ **Done**

JediKev:
Updated

setup/scripts/automail.php
```diff
@@ -71,10 +71,13 @@ function_exists('curl_version') or die('CURL support required');
             $code = 69;
             break;
         case 500: //Server error.
-        default: //Temp (unknown) failure - retry 
+        default: //Temp (unknown) failure - retry
             $code = 75;
     }
 }
 
+if (preg_match('/HTTPS Protocol Required/', $result, $m) == 1) {
```
protich: Curious why you didn't rely on http code returned here! Asking because we might end up translating response text down the road which will break the check here of course.

main.inc.php
```diff
@@ -33,6 +33,15 @@
 if(!($ost=osTicket::start()) || !($cfg = $ost->getConfig()))
 Bootstrap::croak(__('Unable to load config info from DB.').' '.__('Get technical help!'));
 
+if ($cfg && $cfg->forceHttps()
+        && !defined('HTTP_OK')
```
protich: I actually wonder if we need to use `HTTP_OK` flag at all! It came about because of SAAS environment that required domain...etc but in general osTicket installation we can just check `osTicket::is_cli()`

setup/scripts/automail.php
JediKev: Because 400 is super generic and not related to Force HTTPS specifically. I can return a specific response code and check against that but if we return the same response code for something else the check will be inaccurate. Maybe I can set a specific response header and check that instead?

main.inc.php
JediKev: I agree with that. I can revert this to `&& !osTicket::is_cli()` and remove the added `define('HTTP_OK', TRUE);` elsewhere.

setup/scripts/automail.php
protich: As long as you're not using a code like 403, 404 etc that server might actually return in normal operations then I think you're fine in the context of osTicket specific handshake. 

We use 201 for example to indicate a ticket was created. You could use  495 for example... which nginx uses in the context of ssl certificate error - which I think is not an issue within this specific use case. 

Just be sure to document it in `class.http.php` -  as you can see 400 is already in there.

```php
class Http {

    function header_code_verbose($code) {
        switch($code):
        case 200: return '200 OK';
        case 201: return '201 Created';
        case 204: return '204 No Content';
        case 205: return '205 Reset Content';
        case 306: return '306 Switch Proxy';
        case 400: return '400 Bad Request';
        case 401: return '401 Unauthorized';
        case 403: return '403 Forbidden';
        case 404: return '404 Not Found';
        case 405: return '405 Method Not Allowed';
        case 416: return '416 Requested Range Not Satisfiable';
        case 418: return "418 I'm a teapot";
        case 422: return '422 Unprocessable Entity';
        default:  return '500 Internal Server Error';
        endswitch;
    }

```

JediKev: That’s a bet

PR #5274, opened 2019-12-13
sanitize input: remove double-quote from username.

mchobbel:
This patch prevents the ability for an agent to save a user's  name with double-quotes. Nb. when a user has double-quotes around their name (e.g. "John Doe") we run into some unexpected behavior; the email-address is hidden in that case.


JediKev:
Fixed with: https://github.com/osTicket/osTicket/pull/5395

PR #5283, opened 2019-12-17
Issue: Template Variables in Ticket Filter

aydreeihn:
This commit fixes an issue where you could not use template variables when using the 'Send an Email' filter action. All variables would just display as 'Array'.

We needed to pass the Ticket object to FA_SendEmail::apply, however, we were running the filters before the Ticket object was created. To fix this, we can add the Ticket object to vars and run the filterTicketData function again after the Ticket is created and only apply the send email filter action if the Ticket object is found.

### DIFF
--- a/include/class.filter.php
+++ b/include/class.filter.php
@@ -331,8 +331,14 @@ extends VerySimpleModel {
      * If the matches() method returns TRUE, send the initial ticket to this
      * method to apply the filter actions defined
      */
-    function apply(&$ticket, $vars) {
+    function apply(&$ticket, $vars, $postCreate=false) {
         foreach ($this->getActions() as $a) {
+            //control when certain actions should be applied
+            //if action is send email and postCreate == false, skip
+            //if action is not send email and postCreate == true, skip
+            if ((($a->type == 'email') ? !$postCreate : $postCreate))
+                continue;
+
             $a->setFilter($this);
             $a->apply($ticket, $vars);
         }
@@ -848,9 +854,9 @@ class TicketFilter {
      * should be rejected, the first filter that matches and has reject
      * ticket set is returned.
      */
-    function apply(&$ticket) {
+    function apply(&$ticket, $postCreate=false) {
         foreach ($this->getMatchingFilterList() as $filter) {
-            $filter->apply($ticket, $this->vars);
+            $filter->apply($ticket, $this->vars, $postCreate);
             if ($filter->stopOnMatch()) break;
         }
     }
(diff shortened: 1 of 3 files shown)


include/class.filter_action.php
```diff
@@ -516,40 +516,44 @@ class FA_SendEmail extends TriggerAction {
...
-        require_once PEAR_DIR . 'Mail/RFC822.php';
-        require_once PEAR_DIR . 'PEAR.php';
-
-        if (!($mails = Mail_RFC822::parseAddressList($to)) || PEAR::isError($mails))
-            return false;
-
-        // Allow %{recipient} in the body
-        foreach ($mails as $R) {
-            $recipient = sprintf('%s <%s@%s>', $R->personal, $R->mailbox, $R->host);
+        if ($ticket['ticket']) {
```
JediKev: If no `$ticket['ticket']` just return false.

include/class.ticket.php
```diff
@@ -3851,6 +3851,9 @@ static function create($vars, &$errors, $origin, $autorespond=true,
             return null;
 
         /* -------------------- POST CREATE ------------------------ */
+        $vars['ticket'] = $ticket;
```
JediKev, line 4302: Use only certain filters with actions like Send Email, etc.

aydreeihn:
It seemed like the best way to control when we run certain filter actions (before or after ticket creation) was best handled with adding the postCreate variable to the apply functions. I passed it as its own variable rather than sending it through vars because later on in the process, we start using $this->vars instead of just $vars, and it wasn't working out at the end. I'm open to suggestions though if there is a better way to make this work with a function rather than a variable. Also, it didn't seem like there were any other actions we needed to happen postCreate. Can you guys think of any we'd need to wait to run?

include/class.filter.php
```diff
@@ -278,8 +278,14 @@ function getActions() {
      * If the matches() method returns TRUE, send the initial ticket to this
      * method to apply the filter actions defined
      */
-    function apply(&$ticket, $vars) {
+    function apply(&$ticket, $vars, $postCreate=false) {
         foreach ($this->getActions() as $a) {
+            //control when certain actions should be applied
+            //if action is send email and postCreate == false, skip
+            //if action is not send email and postCreate == true, skip
+            if ((($a->type == 'email') ? !$postCreate : $postCreate))
```
protich, line 339: Did you mean - at least based on the comment above.
```php
 if ($a->type == 'email' && !$postCreate)
     continue;
```

aydreeihn: I wanted to write it in the shortest way possible. Originally, I had this:

```
if (($a->type == 'email' && !$postCreate) || ($a->type != 'email' && $postCreate))
     continue;
```

I tested the current code a lot, and both give the same results. Should I still just write it out for readability? I wrote the comment above it because I kept second guessing it.

Issue #1054, opened 2014-06-20
Show Help Topic in Ticket View


mcardosoti:
Hi

I need show Help topics in SCP ticket view, can someone help me?

![help_topic](https://cloud.githubusercontent.com/assets/7945744/3345864/519e3354-f8be-11e3-9bfe-0e3c2a7cee22.png)

PR #5291, opened 2019-12-19
Issue: DB Error #1054

aydreeihn:
Error: Unknown column 'Z1.ticket_id' in 'field list'

This commit fixes an issue with how we approach skipping counts. If a search has a keyword or if a saved queue/search has an invalid criteria saved, we will completely ignore trying to do a count. In saved queue/searches, rather than showing a number or 500, we will display '-' in its place. For the count at the bottom (Showing # - # of about #), we will just display Showing # - #.

### DIFF
--- a/include/class.pagenate.php
+++ b/include/class.pagenate.php
@@ -34,13 +34,14 @@ class PageNate {
     }
 
     function setTotal($total, $approx=false) {
-        $this->total = intval($total);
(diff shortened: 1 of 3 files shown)


JediKev (approved):
Found one discrepancy with keys but it has been mitigated and now all tests return proper results. Searches/Saved Queues with keywords show `-` as the count, if no results we do not show `X - Y of about 500`, and no db errors being thrown.

PR #5293, opened 2019-12-19
mail: Mail_Parse::getAttachments ()

JediKev:
This addresses an issue where some email attachments are not being added to the thread. This is due to Pear's `Mail_mimeDecode::_parseHeaderValue` function not parsing the headers correctly. This updates `Mail_Parse::getAttachments` to check the `content-disposition` for a filename if all else fails. This will ensure a filename is returned if the headers are not parsed correctly.


EIJason:
Ok applied both those fixes still no luck?

PR #5294, opened 2019-12-19
Issue: Remove Referrals

aydreeihn:
This commit fixes an issue where we did not check to see if we needed to remove referrals. For example, if a Ticket is referred to a Department and then the Ticket is transferred to that Department, we should remove the referral.

We will now check/remove referrals as needed for Departments, Agents, and Teams.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -2270,6 +2270,13 @@ implements RestrictedAccess, Threadable, Searchable {
         return $save ? $this->save() : true;
     }
 
+    function hasReferral($object, $type) {
+        if (($referral=$this->thread->getReferral($object->getId(), $type)))
+            return $referral;
+
+        return false;
+    }
+
     //Dept Transfer...with alert.. done by staff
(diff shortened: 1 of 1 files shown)


include/class.ticket.php
```diff
@@ -2314,6 +2314,13 @@ function transfer(TransferForm $form, &$errors, $alert=true) {
         // Log transfer event
         $this->logEvent('transferred');
 
+        if ($this->thread->referrals) {
```
protich: This is probably the **most expensive** way of implementing the check - **1)** Checking each referral (regardless of the type of object) **2)** getName() will trigger object lookup for each check!

I recommend the following;

```php

if (($referral=$this->thread->getReferral($dept->getId(), 
       ObjectModel::OBJECT_TYPE_DEPT)))
      $referral->delete();
```

`getReferral()` uses `findFirst(..)` which doesn't hit the database.

protich (changes requested):

Issue #5259, opened 2019-12-10
Advanced Search - Columns


mlipok:
### Prerequisites

* [ Y ] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [ N ] Do you have any errors in the PHP error log, or javascript console?
* [ Y ] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [ Y ] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/osTicket/osTicket/wiki/Github-Issues-Guidelines)

### Description
"Advanced Search form" have a tab "Columns"
You could add/delete tab, but the result of searching will not return results related to the selected "Columns"

[Description of the bug or feature]

### Steps to Reproduce

1. Go to "Open"
2. Click "Advanced Search"
3. Add some criteria
4. Click the tab "Columns"

**Expected behavior:** [What you expected to happen]
If you Choose "OPEN" or "CLOSED" or any other "Parent/TopLevel Queue" from "---My searches---" ComboBox, the function "Use standard columns" should be checked the column list should be possible to change in the same way, how it works when you add SubQueries in:
Admin Panel >> Settings >> Tickets >> Ticket Queues  >> Add New Queue

I mean the columns should be inherited from "Parent/TopLevel Queue" and you should be able to add new/additional columns.

Attached Screenshot is showing what is wrong.
![AdvanceSearch_Column](https://user-images.githubusercontent.com/11089482/70548506-f3e03400-1b72-11ea-8849-0e46168a3057.png)

**Actual behavior:** [What actually happened]
The result always shows a list with column:    Number, Created, Subject, From, Priority, Assignee

Never shows the expected/selected columns.
Even if you Choose "OPEN" or "CLOSED" or any other subqueries, from "---My searches---" ComboBox, the feature "Use standard columns" is always Checked.

Regardless of the above, you can add an additional column but it will not give you any effect on the result list - the columns on the list are always the same.

### Versions

Admin panel -> Dashboard -> Information which also additionally gives you information about your server.
v1.14.1 (f1e9e88)

--

JediKev:
@mlipok 

[...]

PR #5307, opened 2019-12-30
queue: Inherit Columns Option

JediKev:
This addresses part of issue #5259 where Advanced Searches are showing the Customizable Columns even though you cannot customize an Advanced Search's columns. This adds some JS to hide the `.if-not-inherited` div on initial load if `.inherit-columns` checkbox is Disabled.

This addresses an issue where saving a Parent Queue with "Use standard columns" Enabled will show the option Enabled but still show the customizable Columns instead of the Standard Columns. This addresses an issue where saving a Child Queue with "Inherit columns from the parent queue" Enabled shows the Standard Columns but does not show the option Enabled. This adds a check in `inheritColumns()` for Staff Configurations for the queue and if exists we use the Staff preferences, else we default to the Queue Flag check. This adds a new method called `staffSettings()` that returns staff settings for a queue. It takes an optional argument (a setting's name) that can return a specific setting's value.

This addresses an issue where saving a Saved Search with "Use standard columns" Enabled, shows the Customizable Columns instead of the Standard Columns and does not show the option Enabled as the INHERIT_COLUMNS flag is not set.  This adds a check when saving a SavedSearch for the `inherit-columns` setting and if set, we set the Queue's INHERIT_COLUMNS Flag.

#### Save With "Use standard columns" Checked
O Criteria

#### Save With "Inherit columns from the parent queue" Checked
I Columns

#### Shows Columns
Advanced Ticket Search

#### Save With "Use standard columns" Checked
Saved Search

### DIFF
--- a/include/class.queue.php
+++ b/include/class.queue.php
@@ -1025,7 +1025,8 @@ class CustomQueue extends VerySimpleModel {
     }
 
     function useStandardColumns() {
-        return !count($this->columns);
+        return ($this->hasFlag(self::FLAG_INHERIT_COLUMNS) ||
+                !count($this->columns));
     }
 
     function inheritExport() {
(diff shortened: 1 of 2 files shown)


mlipok:
Please also add a feature that this two following options

Queue_Inherit_Columns_Settings_Open_Closed_Tickets

https://................pl/scp/tickets.php?a=search&status=open&uid=2
https://................pl/scp/tickets.php?a=search&status=closed&uid=2

will inherit column setting accordingly from "main/top level" opened/closed tickets queue:
https://................pl/scp/tickets.php?queue=1
https://................pl/scp/tickets.php?queue=8

Issue #2870, opened 2016-01-18
1.10 Agent welcome email selection is missing


troycarpenter:
In 1.9.12, when adding a new agent there is a checkbox that says for Welcome Email when creating the agent account.  That selection seems to be missing in 1.10.  If it moved to another screen, I can't find it.

troycarpenter:
To make this work, I added the following in include/staff/staff.inc.php at line 175:
```
<?php if ($action=='create') { ?>
    <label class="checkbox">
        <input type="checkbox" name="welcome_email" id="welcome-email" checked="checked"/>
        <?php echo __('Send Welcome Email'); ?>
    </label>
<?php } ?>
```
It appears the backend code still understands how to send the email.  The first line assures the checkbox only shows up when adding an agent.  Also, this assumes you want to send the welcome emails by default as it automatically checks the box.

xdreamarun:
we have added same code at same place.
we are using osticket version 1.10
Fields is displaying while create the agent.
But no email is going to agent.

luismariveros:
I just installed osTicket Version v1.10 (901e5ea) and have the same problem... With the code of @troycarpenter it worked... I added at line 195

lzantal:
Thanks for the code @troycarpenter. This fixed my installation as well. New user to osTicket so I didn't know where to start to enable it. Whats weird is that it sent it out twice before but then it suddenly stopped. Thanks again.

akshaynikhare:
@troycarpenter  the code should be inserted at line 198 with generate better output  
![image](https://user-images.githubusercontent.com/9725336/70790040-3c3d6300-1dba-11ea-9bb3-9d6bc9a199c8.png)

as this  
![image](https://user-images.githubusercontent.com/9725336/70790071-4a8b7f00-1dba-11ea-82c2-78609fd6fde3.png)

ManZosh:
Thanks @akshaynikhare this helped big time! I'm now only wondering why this issue was opened just about 4 years ago and hasn't yet been committed to the code base after 4 versions?

JediKev:
@ManZosh 

Because this is the incorrect way of addressing the issue. The correct fix is found below and should be added to the next release once finalized:
- https://github.com/osTicket/osTicket/pull/5308

[...]

PR #5308, opened 2019-12-30
issue: New Agent Welcome Email

JediKev:
This addresses the underlying problem with issue #2870 where the Welcome Email checkbox is no longer present but even then no Welcome Email is ever sent to the Agent regardless of how they are created. This is due to an option called `welcome_email` that is not being passed to the `update()` method for Staff. When an Agent account is created with no Password set and the Authentication backend is not external we should send the Welcome Email so the Staff in question can create a Password and access the system. If no password is set and the authentication backend is external however then we should not send the email as the Staff will use their AD credentials to login. This adds an else statement to set `welcome_email = 1` if there is no password set and if there is no backend set or is Local.

### DIFF
--- a/scp/staff.php
+++ b/scp/staff.php
@@ -43,6 +43,10 @@ if($_POST){
                 foreach ($_SESSION['new-agent-passwd'] as $k=>$v)
                     if (!isset($_POST[$k]))
                         $_POST[$k] = $v;
+            } else { // If no password && no backend set or is local then send Welcome Email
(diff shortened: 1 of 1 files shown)


heifuma:
Is there an option to trigger the welcome email manually? I've created accounts for all of our staff before I've applied this patch. Even when I reset the user account no password email will be sent. I don't want to recreate all these accounts again.

heifuma:
Sorry, just found your solution on patch [#5319](https://github.com/osTicket/osTicket/pull/5319/files)

It works for me

PR #5312, opened 2020-01-02
redactor: Cancel Button

JediKev:
This addresses an issue reported on the Forum where the New Ticket Cancel buttons do not work in v1.14.1. This is due to the code using older methods utilized in v1.12. This updates the code to use the new method of accessing the plugin objects first before accessing the plugin itself.


JediKev:
@julisu 

Thank you for the report. I was able to replicate this only when selecting a Help Topic. It appears there is an issue when loading a new redactor instance in the same page as an existing redactor instance. Very strange, but we will look into this.

Cheers.

PR #5318, opened 2020-01-07
Display ticket owners telephone number and click2dial link.

Mike-Benoit:
Often time our agents receive a ticket by email, but need to call the ticket owner to get more details before they can resolve the ticket. Therefore displaying the ticket owners telephone number and including a link that uses the standard "tel:" protocol to have the browser launch a telephone app to call them is extremely helpful. 

This small self-contained PR adds that functionality as well as a slightly faster way to inline edit the ticket owners telephone number if needed.

I tested scenarios of inline editing and adding/removing a telephone extension to make sure that is stripped from the "tel:" URL.

### DIFF
--- a/include/staff/ticket-view.inc.php
+++ b/include/staff/ticket-view.inc.php
@@ -513,7 +513,20 @@ if($ticket->isOverdue())
                 <tr>
                     <th><?php echo __('Phone'); ?>:</th>
                     <td>
-                        <span id="user-<?php echo $ticket->getOwnerId(); ?>-phone"><?php echo $ticket->getPhoneNumber(); ?></span>
+                        <a href="#tickets/<?php echo $ticket->getId(); ?>/user"
+                           onclick="javascript:
+                                   saveDraft();
+                                   $.userLookup('ajax.php/tickets/<?php echo $ticket->getId(); ?>/user',
+                                   function (user) {
+                                   $('#user-'+user.id+'-name').text(user.name);
+                                   $('#user-'+user.id+'-email').text(user.email);
+                                   $('#user-'+user.id+'-phone').text(user.phone);
+                                   $('#user-'+user.id+'-phone-url').attr('href', 'tel:'+user.phone.substr(0, ( ( user.phone.indexOf('x') == -1 ) ? user.phone.length : user.phone.indexOf('x') ) ).trim());
+                                   $('select#emailreply option[value=1]').text(user.name+' <'+user.email+'>');
+                                   });
+                                   return false;
+                                   "><span id="user-<?php echo $ticket->getOwnerId(); ?>-phone"><?php echo ( ( $ticket->getPhoneNumber() != '' ) ? $ticket->getPhoneNumber() : __('None') )?></span></a>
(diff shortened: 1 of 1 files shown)


sevmonster:
I made some changes to this that we use in production:
```diff
diff --git a/include/staff/ticket-view.inc.php b/include/staff/ticket-view.inc.php
index 256ddda2..70c0bbee 100644
--- a/include/staff/ticket-view.inc.php
+++ b/include/staff/ticket-view.inc.php
@@ -387,6 +387,12 @@ if($ticket->isOverdue())
                                         $('#user-'+user.id+'-name').text(user.name);
                                         $('#user-'+user.id+'-email').text(user.email);
                                         $('#user-'+user.id+'-phone').text(user.phone);
+                                        var elem = $('#user-'+user.id+'-phone-url');
+                                        var noext = user.phone.split('x', 1)[0];
+                                        if (noext)
+                                            elem.show().attr('href', 'tel:' + noext.replace(/[-()\s]/g, ''));
+                                        else
+                                            elem.hide();
                                         $('select#emailreply option[value=1]').text(user.name+' <'+user.email+'>');
                                     });
                             return false;
@@ -511,20 +517,10 @@ if($ticket->isOverdue())
                 <tr>
                     <th><?php echo __('Phone'); ?>:</th>
                     <td>
-                        <a href="#tickets/<?php echo $ticket->getId(); ?>/user"
-                           onclick="javascript:
-                                   saveDraft();
-                                   $.userLookup('ajax.php/tickets/<?php echo $ticket->getId(); ?>/user',
-                                   function (user) {
-                                   $('#user-'+user.id+'-name').text(user.name);
-                                   $('#user-'+user.id+'-email').text(user.email);
-                                   $('#user-'+user.id+'-phone').text(user.phone);
-                                   $('#user-'+user.id+'-phone-url').attr('href', 'tel:'+user.phone.substr(0, ( ( user.phone.indexOf('x') == -1 ) ? user.phone.length : user.phone.indexOf('x') ) ).trim());
-                                   $('select#emailreply option[value=1]').text(user.name+' <'+user.email+'>');
-                                   });
-                                   return false;
-                                   "><span id="user-<?php echo $ticket->getOwnerId(); ?>-phone"><?php echo ( ( $ticket->getPhoneNumber() != '' ) ? $ticket->getPhoneNumber() : __('None') )?></span></a>
-                                    <a id="user-<?php echo $ticket->getOwnerId(); ?>-phone-url" href="tel:<?php echo trim( substr( $ticket->getPhoneNumber(), 0, ( ( strpos($ticket->getPhoneNumber(), 'x') === FALSE ) ? strlen( $ticket->getPhoneNumber() ) : strripos( $ticket->getPhoneNumber(), 'x' ) ) ) ); ?>"><i class="icon-phone"></i></a>
+                        <span id="user-<?php echo $ticket->getOwnerId(); ?>-phone"><?php echo $ticket->getPhoneNumber(); ?></span>
+                        <a id="user-<?php echo $ticket->getOwnerId(); ?>-phone-url" <?php if (!explode('x', $ticket->getPhoneNumber(), 2)[0]) echo 'style="display:none"'; ?>
+                           href="tel:<?php echo explode('x', preg_replace('/[-()\s]/', '', $ticket->getPhoneNumber()), 2)[0]; ?>"
+                        ><i class="icon-phone"></i>
                     </td>
                 </tr>
             </table>
```

- I do not think the phone number also needs to function as an edit link, that is pointlessly duplicated code. The agent can just click on the user to edit their profile if they want. If that were to be duplicated, it should at least be abstracted out into a function applied to both areas.
- The code used to generate the `tel:` link has been cleaned up. `strpos` and etc was replaced with much shorter and easier to read regex.
- The clickable phone icon is hidden automatically if there is no phone number assigned.

Potential change: I explicitly check if there is a phone number set _after_ stripping a possible extension; vanilla osTicket requires a phone number for an extension to be set in the first place, so this may not be required depending on your use case.

PR #5320, opened 2020-01-07
Update class.filter.php

### LINKED ISSUE #726: New Ticket User Auto response not sent to user (email tickets)
Current Version: V1.8.1.2.

When a user sends an email (uses email piping) to the support email, no New Ticket User Auto response is received by the user. The ticket is received ok. Other emails to the user are ok, for example when a reply is sent to the user they are received ok. 

The problem only exists if the ticket is initiated by email, as when a new ticket is initiated by the web,  a New Ticket User Auto response is received by the user ok, therefore supporting that all the appropriate autoresponse settings are ok. 

The only difference it seems between the email initiated new tickets and the web new ticket is that the web new ticket requires the user to specify a help.

Any suggestions would be most appreciated please

jezuk:
Remove X-Auto-Response-Suppress from $auto_headers
Fixes #726

Issue #726 was recently closed however the problem remains.  No further comments have been added since my last post in December 2016, I suggest the reviewing these comments and URLs linked describing a similar issue.

I have made this change on my production osTicket for several years now with no ill effects, including when out of office is enabled, there is no resulting email storm.

Credit to forum user lkrms who first posted this fix.  Thanks for your consideration.

### DIFF
--- a/include/class.filter.php
+++ b/include/class.filter.php
@@ -800,6 +800,7 @@ class TicketFilter {
             'Precedence'        => array('AUTO_REPLY', 'BULK', 'JUNK', 'LIST'),
             'X-Precedence'      => array('AUTO_REPLY', 'BULK', 'JUNK', 'LIST'),
             'X-Autoreply'       => 'YES',
+            'X-Auto-Response-Suppress' => array('AutoReply'),
             'X-Autoresponse'    => '*',
             'X-AutoReply-From'  => '*',
             'X-Autorespond'     => '*',
--- a/include/class.filter.php
+++ b/include/class.filter.php
@@ -800,7 +800,6 @@ class TicketFilter {
             'Precedence'        => array('AUTO_REPLY', 'BULK', 'JUNK', 'LIST'),
             'X-Precedence'      => array('AUTO_REPLY', 'BULK', 'JUNK', 'LIST'),
             'X-Autoreply'       => 'YES',
-            'X-Auto-Response-Suppress' => array('ALL', 'DR', 'RN', 'NRN', 'OOF', 'AutoReply'),
             'X-Autoresponse'    => '*',
             'X-AutoReply-From'  => '*',
             'X-Autorespond'     => '*',


protich:
Why on earth would you remove the auto response detection?  It's there to prevent email loops.

jezuk:
It's just specifically X-Auto-Response-Suppress that causes an issue, X-Autoreply, X-Autoresponse, X-AutoReply-From, X-Autorespond all work as expected for the reason you state.

Redmine had the same issue previously:
http://www.redmine.org/issues/19558

Another reference:
https://www.jitbit.com/maxblog/18-detecting-outlook-autoreplyout-of-office-emails-and-x-auto-response-suppress-header/

protich:
If email server adds the header `X-Auto-Response-Suppress` then I think we should honor it as a matter of standard. Happy to revisit it - but it has to be a setting (perhaps a filter) to ignore the header.

jezuk:
Ok, I'm happy to submit a more substantial pull request with this as a configurable option if that is your preference.  However I'm curious if you have a standard you are referring to, that's kind of the point, it's a misunderstood header.  There is no mention of it in https://tools.ietf.org/html/rfc3834

include/class.filter.php
```diff
@@ -800,7 +800,6 @@ static function isAutoReply($headers) {
             'Precedence'        => array('AUTO_REPLY', 'BULK', 'JUNK', 'LIST'),
             'X-Precedence'      => array('AUTO_REPLY', 'BULK', 'JUNK', 'LIST'),
             'X-Autoreply'       => 'YES',
-            'X-Auto-Response-Suppress' => array('ALL', 'DR', 'RN', 'NRN', 'OOF', 'AutoReply'),
```
JediKev, line 803: After further review we agree but we will keep the `AutoReply` option to respect suppressing auto-replies:
```php
'X-Auto-Response-Suppress' => array('AutoReply'),
```

jezuk: Great, thanks for reviewing this

jezuk:
I have made the changes suggested by @JediKev and can confirm this works as intended, resolving issue #726

Issue #5317, opened 2020-01-06
Merging tickets doesn't play nice with "Require Help Topic to Close" 


Mike-Benoit:
### Prerequisites

* [Yes] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [No] Do you have any errors in the PHP error log, or javascript console?
* [Yes] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [Yes] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

### Description

Merging tickets doesn't handle "Require Help Topic to Close"  properly if there is no help topic specified.

### Steps to Reproduce

1. Enable "Require Help Topic to Close" 
2. Create a parent ticket, and specify the help topic.
3. Create a new ticket without a help topic.
4. Merge the new ticket (without a help topic) to the parent ticket, and when doing so, set the "Child Status" dropdown box to "Resolved". 

**Expected behavior:** [What you expected to happen]
Child ticket should be resolved, or the user should be prompted that they can't change the status on the child ticket without setting a help topic before the changes are saved.

Personally since the child ticket is being merged anyways, I think it should force the ticket status to resolved regardless of the Help Topic not being specified, since the parent help topic is going to be the more useful one anyways. My understanding is the only reason to keep the child ticket around is to route new replies to it to the parent ticket.

**Actual behavior:** [What actually happened]
Child ticket is not resolved.

### Versions
v1.14.1

JediKev:
@Mike-Benoit 

This should resolve your issue:
- https://github.com/osTicket/osTicket/pull/5321

Cheers.

PR #5321, opened 2020-01-07
Issue: Close Child Ticket Without Help Topic

aydreeihn:
When merging Tickets, we automatically set children Tickets to a closed status, however, we had an issue with child Tickets that did not have a Help Topic. If the helpdesk had the 'Require Help Topic to Close' configuration, the children Tickets would not close as expected.

This fixes Issue #5317

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -1443,7 +1443,7 @@ implements RestrictedAccess, Threadable, Searchable {
     }
 
     // Ticket Status helper.
-    function setStatus($status, $comments='', &$errors=array(), $set_closing_agent=true) {
+    function setStatus($status, $comments='', &$errors=array(), $set_closing_agent=true, $force_close=false) {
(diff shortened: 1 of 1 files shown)


Mike-Benoit:
Thanks for the quick fix! I have confirmed it does resolve the issue.

include/class.ticket.php
```diff
@@ -331,7 +331,7 @@ function isCloseable() {
         } elseif (($num=$this->getNumOpenTasks())) {
             $warning = sprintf(__('%1$s has %2$d open tasks and cannot be closed'),
                     __('This ticket'), $num);
-        } elseif ($cfg->requireTopicToClose() && !$this->getTopicId()) {
+        } elseif ($cfg->requireTopicToClose() && !$this->getTopicId() && !$this->isChild()) {
```
JediKev: Check before we call `isCloseable()`.

Issue #5316, opened 2020-01-06
Inline edit of Help Topic doesn't refresh Ticket Status options.


Mike-Benoit:
### Prerequisites

* [Yes] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [No] Do you have any errors in the PHP error log, or javascript console?
* [Yes] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [Yes] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

### Description

If you have the setting enabled that requires a help topic to be specified before a ticket can be closed, and you use the inline edit to set the help topic, you still can't close the ticket until you refresh the page. 

Unfortunately this makes the newly added feature of inline help topic selection (#4985) almost useless with this setting. Which means the user can easily lose their reply when they try to set the help topic without using inline edit, since the entire page is refreshed. :(

### Steps to Reproduce

1. Turn on "Require Help Topic to Close" setting.
2. Create ticket without help topic.
3. View ticket, notice the "This ticket is missing a Help Topic and cannot be closed" message at the bottom. Also any of the "Closed/Resolved" options aren't even available to select.
4. Using inline edit set a help topic.
4. Notice the Ticket Status dropdown box at the bottom doesn't show any options to close/resolve the ticket.

**Expected behavior:** [What you expected to happen]
When the help topic is updated, it should refresh the Ticket Status dropdown box and remove the warning message about no help topic being selected.

**Actual behavior:** [What actually happened]
See above steps to reproduce. 

### Versions
v1.14.1

JediKev:
@Mike-Benoit 

Thank you for the report. We will do some testing and attempt to replicate the issue.

Cheers.

JediKev:
@Mike-Benoit 

This should address your issue:
- https://github.com/osTicket/osTicket/pull/5322

Cheers.

Mike-Benoit:
We recently pulled develop-next and this bug seems to have appeared again. I manually checked the files and the code is definitely there. Are you able to replicate this bug again?

PR #5322, opened 2020-01-07
inline: Set Help Topic Refresh Statuses

JediKev:
This addresses issue #5316 where having `Require Help Topic to Close` Enabled and setting a Help Topic via Inline Edit will not let you Close the ticket via response box unless you refresh the page. This is due to Inline Edit not refreshing the statues in the dropdown to select from. This adds a check in `include/ajax.tickets.php` that checks for if we Require HT On Close, if the ticket is open, if the staff has permission to close, and if we set a help topic. If all of this is true we gather all of the Closed statuses and return them in the response as `statuses`. In addition, this adds a check in the AJAX success to see if `statuses` were returned. If so, we loop through each, see if they already exist in the list and if not we append them as new options. After we add all of the statues we then hide the warning banner. At this point the agent can close the ticket via response box if they choose to do so.

### DIFF
--- a/include/ajax.tickets.php
+++ b/include/ajax.tickets.php
@@ -590,7 +590,7 @@ class TicketsAjaxAPI extends AjaxController {
     }
 
     function editField($tid, $fid) {
-        global $thisstaff;
+        global $cfg, $thisstaff;
 
         if (!($ticket=Ticket::lookup($tid)))
             Http::response(404, __('No such ticket'));
@@ -643,10 +643,29 @@ class TicketsAjaxAPI extends AjaxController {
                              $clean = Format::truncate($clean, 200);
                 }
 
+                // Set basic response data
+                $response = array(
+                    'value' => $clean ?: '&mdash;' . __('Empty') .  '&mdash;',
+                    'id' => $fid, 'msg' => $msg
+                );
+
+                // If we require HT to close, the ticket is open, the staff has permission
+                // to close, and we set a HT - ensure we provide all available statuses
+                if ($cfg->requireTopicToClose() && $ticket->isOpen()
+                        && $ticket->checkStaffPerm($thisstaff, Ticket::PERM_CLOSE)
+                        && ($field instanceof TopicField) && $clean) {
+                    $statuses = array();
+                    foreach (TicketStatusList::getStatuses(
+                            array('states' => array('closed'))) as $s) {
+                        if (!$s->isEnabled()) continue;
+                        $statuses[$s->getId()] = $s->getName();
+                    }
+                    if (!is_null($statuses))
+                        $response['statuses'] = $statuses;
+                }
+
                 $clean = is_array($clean) ? $clean[0] : $clean;
-                Http::response(201, $this->json_encode(['value' =>
-                            $clean ?: '&mdash;' . __('Empty') .  '&mdash;',
-                            'id' => $fid, 'msg' => $msg]));
+                Http::response(201, $this->json_encode($response));
             }
 
             $form->addErrors($errors);
--- a/scp/js/ticket.js
+++ b/scp/js/ticket.js
@@ -366,6 +366,23 @@ var ticket_onload = function($) {
                 $('#msg-txt').text(obj.msg);
                 $('div#msg_notice').show();
             }
+            // If Help Topic was set and statuses are returned
+            if (obj.statuses) {
+                var reply = $('select[name=reply_status_id]');
+                var note = $('select[name=note_status_id]');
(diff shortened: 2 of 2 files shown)


Mike-Benoit:
Thanks for the quick response! 

While this does fix the issue with the Ticket Status dropdown box when on the "Post Reply" tab, the original bug still exists on the "Internal Note" tab on the same screen.  Apparently they are two different dropdown boxes?

JediKev:
@Mike-Benoit 

Pull updated, try now.

Cheers.

Mike-Benoit:
Unfortunately since you did a force push we can't automatically apply this to our fork, so its going to take some more time to test. Not sure why a force push was required rather than just adding another commit, but it would sure make things easier if force pushes weren't used.

We've ran into this a few times with the osTicket git repo and it causes a lot of problems for any downstream developers.

JediKev:
@Mike-Benoit 

It's very messy to add another commit like that. This is a pull so it should be as clean as possible and contain only necessary commits. You should always apply patches/diffs for testing so you can revert the changes if needed.

To expand on this, imagine if I needed to change the wording of something like 30 different times, with your method I'd have 30 useless commits and 1 actual commit for adding one thing. That's not good practice. A pull request should contain the core commits and not "changed this to this because I forgot" commits. Just way too messy and unnecessary.

Also, never merge something to your fork until it's merged upstream to avoid things like this. Keep them as patches (if you need to do something else you can simply stash the changes) and then once the pull is merged to core you can either checkout (remove) the changes and merge with upstream or just stash the changes, merge with upstream, and when you pop the stash only the non-committed changes will be applied.

Cheers.

Mike-Benoit:
Even applying patches/diffs causes problems with having to revert one and apply another, and if you have merge conflicts it requires you to repeat all the conflict resolution multiple times. This is why force commits should only be used as a last resort to fix a mistake.

Personally, I've never seen a git workflow that involves regular force commits. Usually PR's will have many commits (some would argue thats actually beneficial, like Linus himself), and when merging to the master/develop branch you simply do a squash merge so they appear in that branch as a single commit. Keeps the history on the master/develop branch nice and clean, while not causing headaches for everyone.

JediKev:
@Mike-Benoit 

I don't want to keep going back and forth about this, so my last comment is:
I personally do not like that concept so it's a matter of opinion I guess. :)

Cheers.

protich:
@Mike-Benoit - We try our hardest not to force-push anything to upstream especially to master, develop or develop-next. So once something is merged upstream - nothing can be amended to avoid exact situation you're talking about.

That said, we consider personal branches temporary (proposals) until merged - so pulling from such branch to production is unwise in so many ways since during the review I might ask for rewrites etc. and we've generally allowed amending such commits.

Mike-Benoit:
> @Mike-Benoit - We try our hardest not to force-push anything to upstream especially to master, develop or develop-next. So once something is merged upstream - nothing can be amended to avoid exact situation you're talking about.
> 
> That said, we consider personal branches temporary (proposals) until merged - so pulling from such branch to production is unwise in so many ways since during the review I might ask for rewrites etc. and we've generally allowed amending such commits.

Thanks for the response @protich. I respect your choice of work flow, even though it seems counter productive to me. 

I'm sure you would agree that if a user submits a bug report, and a fix is generated, the best case scenario is to have the original submitter test the fix and confirm that it works for them. If not, or if something else is broken they can quickly report it (like I did in this situation) and additional fixes can be easily made, rinse and repeat. The sooner this all happens the more efficient the entire process is. Alternatively, waiting until a release is made, then having the user report that it doesn't fix the bug is frustrating for everyone involved. 

With that in mind, I would expect that doing everything reasonable to encourage and make it easier for users to test bug fixes would be important, especially for an open source project. 

However I'm failing to see how force pushing to what you consider a temporary branch serves any benefit since the branch will just be deleted at some point anyways. Instead it makes it much more difficult for the user to apply subsequent fixes/rewrites and instead **actively discourages them from testing said changes** as it did in this case. It turned what could have been a 5 minute or less job into a much larger and complicated one I simply don't have the time for right now. It also makes your job more difficult since now its one less person to test the fix, and increases the chance of a new issue being submitted after the fix is released or merged anyways. If @JediKev's goal in doing a force push is to keep things "clean", I don't see how submitting two separate bug reports, weeks or months apart, after the first one is already merged makes things clean. 

Anyways, @JediKev mentioned in the forums, that osTicket will be "ramping up community involvement" ( https://forum.osticket.com/d/96665-is-osticket-open-to-contributions ), so please take this as a suggestion that I believe would help with that.

PR #5327, opened 2020-01-09
Redactor revert enter key behavior back to pre-v1.14 behavior.

Mike-Benoit:
osTicket v1.14 changed the behavior of the "enter" key so it adds a new paragraph in redactor rather than the more standard adding of a new line. This one-line change reverts this behavior back.

Brought up on the forums here too:
https://forum.osticket.com/d/96617-double-spacing-on-ticket-thread/2

### DIFF
--- a/js/redactor-osticket.js
+++ b/js/redactor-osticket.js
@@ -377,6 +377,7 @@ $(function() {
                 'minHeight': selectedSize,
                 'maxWidth': el.hasClass('fullscreen') ? '950px' : false,
                 'focus': false,
+                'breakline': true,
                 'plugins': el.hasClass('no-bar')
                   ? ['imagemanager','definedlinks']
                   : ['imagemanager','table','video','definedlinks','autolock', 'fontcolor', 'fontfamily'],


ntozier:
Personally i would rather see this change include a setting to control the behavior.

TechCare-Aus:
@ntozier I can see your point and would agree. More customization is good.

But personally I would agree with @Mike-Benoit I find myself holding down shift a lot lot more nowadays.

JediKev:
@Mike-Benoit 

@ntozier is correct, this will need to be a configurable setting as most people actually do like this ability; so we can't just disable it flat out.

Cheers.

JediKev:
@Mike-Benoit 

I redid your commit below to make it an optional setting:
- https://github.com/osTicket/osTicket/pull/5330

Cheers.

dali-dbr:
Yes please make it an option. It's driving me crazy. It's not something you just switch, without giving an option for users, that use it daily, can set it back. 

Br,
Dali

ntozier:
@dali-dbr two things.  First, you are posing to a closed issue.  Second, it is not being reverted. The PR in the message that you replied to adds it as a configuration option.  All you have to do is apply that PR to your instance or wait until the next version is released.

PR #5330, opened 2020-01-09
redactor: Double Spacing Optional

JediKev:
This addresses an issue where upgraded redactor does double spacing by default (which matches most modern editors). A lot of people do not like this change however so the solution is to make double spacing optional via a setting. This adds a new Profile setting called Editor Spacing that allows you to configure Double or Single spacing for the editor. If set to Double, the text editor will double space when pressing Enter. If set to Single, the text editor will single space when pressing Enter like the old redactor version. If a staff does not have this setting configured, it defaults to Double Spacing.

### DIFF
--- a/include/ajax.config.php
+++ b/include/ajax.config.php
@@ -50,6 +50,7 @@ class ConfigAjaxAPI extends AjaxController {
               'secondary_languages' => $cfg->getSecondaryLanguages(),
               'page_size'       => $thisstaff->getPageLimit() ?: PAGE_LIMIT,
               'path'            => ROOT_PATH,
+              'editor_spacing' => $thisstaff->editorSpacing(),
         );
         return $this->json_encode($config);
     }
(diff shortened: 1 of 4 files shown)


TechCare-Aus:
@JediKev you sir are a legend

danielw0581:
Thank you @jedikev! Is there anyway a system wide flag for a default could be added at a system level? Our organization wants the default to be the single line spacing.

TechCare-Aus:
@danielw0581 - As far as I know the redactor code is system wide. 

@JediKev - I've tested this on my system and so far it has worked flawlessly. Brilliant feature update in my opinion.

JediKev:
@danielw0581 

The default is set to Double Spacing to keep things the way they currently are without making a huge change again. This prevents people from upgrading and then suddenly the editor is acting differently yet again. If you would like to change the default you can apply this pull request and change `include/class.staff.php` line `97` from:
```
'editor_spacing' => 'double',
```
to:
```
'editor_spacing' => 'single',
```

@TechCare-Aus Thank you for testing and for the kind words!

Cheers.

luke5sky:
works very well for me! thank you

JediKev:
Updated

PR #5331, opened 2020-01-09
Issue: Organization Field Change Name Type

aydreeihn:
This commit prevents Agents from being able to change the field type of the name field on the Organization form. We should be forcing it to be a Short Answer field.


protich:
We had a conversion about this PR - don't have notes on me but if I recall we needed to look into why `isChangeable()` is not working as expected in this particular case.

aydreeihn:
I think for this one, the field should have already been set up correctly based on a pull @JediKev had made a while back, and we weren't sure why.

PR #5340, opened 2020-01-16
Delete favicon.png

abbi50634:


JediKev:
@abbi50634 

Why did you make a pull request to delete the favicon? If you want to remove the favicon then you can remove it from your local files as we ship with this as a default.

Cheers.

PR #5348, opened 2020-01-24
[Feature Request] Add Signals to allow plugins to add headers

AdamDempsey:
Added two new signals so plugins can add custom headers to client + staff pages.

(Excuse the multiple commits + reverts, first pull request!)


estebandelaf:
I have the same requirement. I solved using another approach.

But the idea is having the functionality in the osticket core files.

My vote is for this approach with signals.

In my case I use a new method in plugins, more info in: https://forum.osticket.com/d/98352-use-a-plugin-to-add-javascript-code-to-head-tag-in-html

PR #5354, opened 2020-01-29
issue: User Manage Org Name

JediKev:
This addresses an issue reported on the Forum where clicking Manage Account for a User the Organization field is blank even though the User has an Organization. This is due to the system only using `$info` which is set by `$_POST`. Since we are not posting anything, just loading the popup template, there is no `$_POST` so `$info` will not be set. In addition, you cannot update the Organization using the Manage Account modal so there is no reason to use `$info` nor is there a reason to have the Organization field as an Input field. This updates the modal template to make the Organization field simple table data and to use `$user->getOrganization()->getName()` for the Organization Name.


JediKev:
Updated

PR #5357, opened 2020-01-30
plugins: Add Version Column

JediKev:
This adds a new column to the Plugins template to show the Version Number for each plugin. This adds a new method called `getVersion()` that will return either the version number from the database or the version number from the `load()` method.

In addition this addresses an issue where upon installing a plugin, no version number is added in the database. This is due to not setting the version in the SQL query when inserting the plugin table record.


JediKev:
This is a rewrite of pull [#3486](https://github.com/osTicket/osTicket/pull/3486) by @ntozier.

Issue #5372, opened 2020-02-11
Close Child Ticket Auto-refer Error


jacobshilitz:
### Prerequisites

* [ ] Can you reproduce the problem in a fresh installation of the "develop" branch? 
* [x] Do you have any errors in the PHP error log, or javascript console?
* [x] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [x] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

### Description

When closing a child ticket (in case it was reopened for some reason) the code has a bug, that it tries to refer the thread to a previously assigned or closing agent. but `$this->thread` is null because there is no thread in a child of a merged ticket.

I fix it in [my fork](https://github.com/jacobshilitz/osTicket/commit/89522de13423b57a9bbd295317b1c24ed3f9b8b7) by using `$this->getThread()->refer($refer)` instead of `$this->thread->refer($refer)`, but I think it could also happen in other places where `$this->thread` is being called, as now it's possible to have a ticket with no thread.

I'm not sure to which branch to open the PR because I don't see this commit in the develop branch at all

I couldn't "reproduce the problem in a fresh installation of the "develop" branch" because its v1.12

Thanks for developing such a good product

### Steps to Reproduce
1. Check to enable `Auto-refer on Close` in the setting
1. Merge 2 tickets
2. Reopen the Child Ticket
3. try to close the ticket

### Versions

osTicket Version | v1.14.1 (f1e9e88) —  Up to date
-- | --
Web Server Software | Apache/2.4.38 (Debian)
MySQL Version | 5.7.29
PHP Version | 7.1.33

```
Fatal error: Uncaught Error: Call to a member function refer() on null in /var/www/html/include/class.ticket.php on line 1545
( ! ) Error: Call to a member function refer() on null in /var/www/html/include/class.ticket.php on line 1545
```

jacobshilitz:
Thanks @aydreeihn, its the first time I wroth an issue, and it was nice to see it's being taken seriously

PR #5374, opened 2020-02-11
Issue: Managing Child Ticket Threads

aydreeihn:
This commit makes sure that we do ticket->getThread() rather than ticket->thread so that if we are dealing with a child ticket that doesn't have a thread, we won't get any errors.

This fixes Issue #5372

### DIFF
--- a/include/ajax.tickets.php
+++ b/include/ajax.tickets.php
@@ -565,7 +565,7 @@ class TicketsAjaxAPI extends AjaxController {
                             if ($v[0] == '-')
                                 $remove[] = substr($v, 1);
                         if (count($remove)) {
-                            $num = $ticket->thread->referrals
+                            $num = $ticket->getThread()->referrals
                                 ->filter(array('id__in' => $remove))
                                 ->delete();
                             if ($num) {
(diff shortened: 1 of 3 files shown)


jacobshilitz:
There is [2 more](https://github.com/osTicket/osTicket/blob/develop-next/include/class.collaborator.php#L64) in `class.collabrator.php` not sure if it should be edited as well, but I don't see why not

aydreeihn:
@jacobshilitz I don't think we'll ever hit collaborator->thread at a time where there isn't a thread, but I went ahead and put a check in there just to be safe. Thanks for the feedback!

include/ajax.tickets.php
```diff
@@ -565,7 +565,7 @@ function refer($tid, $target=null) {
                             if ($v[0] == '-')
                                 $remove[] = substr($v, 1);
                         if (count($remove)) {
-                            $num = $ticket->thread->referrals
+                            $num = $ticket->getThread()->referrals
```
protich, line 568: I don't think this helps at all - since getThread will return a null. Plenty of same instances in class.ticket.php.

PR #5375, opened 2020-02-12
Fix ticket cdata creation for selection fields

TrueType:


aydreeihn:
@TrueType I'm a little confused on the change. Looks like it's not saving anything for multiselect choice fields. Check out the screenshot below. The highlighted values show the different results I got where Old Way was what was saved for the old code and New Way is the new code.

Screen Shot 2020-03-26 at 1 37 31 PM

In the Ticket queue:

Screen Shot 2020-03-26 at 1 45 21 PM

PR #5380, opened 2020-02-18
issue: BooleanField Cannot Be Unchecked

JediKev:
This addresses an issue where checkboxes cannot be unchecked once saved as checked. This was introduced with a fix for another issue where it's necessary to return the default value for VisibilityConstraints. This is fine for Holiday Schedule checkboxes but for custom fields this breaks basic functionality. This updates `getValue()` for CheckboxWidget to only return the default value if `$data` (source) is not set.

This also addresses an issue where after setting a checkbox value via Inline Edit it shows the value as `0` or `1` instead of `Yes` or `No`. You have to refresh the page in order for the value to appear correctly. This is due to the raw value being returned instead of it's string equivalent. This adds a case in `TicketsAjaxAPI::editField()` for `BooleanField` that converts the raw value to the string value with `toString()`.

### DIFF
--- a/include/ajax.tickets.php
+++ b/include/ajax.tickets.php
@@ -641,6 +641,9 @@ class TicketsAjaxAPI extends AjaxController {
                         if (strlen($clean) > 200)
                              $clean = Format::truncate($clean, 200);
                         break;
+                    case $field instanceof BooleanField:
+                        $clean = $field->toString($field->getClean());
+                        break;
                     default:
                         $clean =  $field->getClean();
                         $clean = is_array($clean) ? implode($clean, ',') :
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -4623,7 +4623,7 @@ class CheckboxWidget extends Widget {
                 return $data[$this->field->get('id')];
         }
 
-        if (isset($this->value))
+        if (!$data && isset($this->value))
             return $this->value;
 
 


gitsnow:
I was able to test out the fix and it works
Thanks

include/ajax.tickets.php
```diff
@@ -641,6 +641,10 @@ function editField($tid, $fid) {
                         if (strlen($clean) > 200)
                              $clean = Format::truncate($clean, 200);
                         break;
+                    case $field instanceof BooleanField:
+                        $clean = $field->getClean();
+                        $clean = $field->toString($clean) ?: '&mdash;' . __('Empty') .  '&mdash;';
```
aydreeihn: It looks like $field->toString($clean) will always return Yes or No no matter what you pass into toString, so I don't think you need to put the 'empty' part in there. I tried manually passing in random strings, numbers, and quotes, and every time it would return either Yes or No. Not sure if we should just leave it there to be absolutely sure though. It doesn't hurt anything.

aydreeihn:
Other than that one comment, this looks good to me!

include/ajax.tickets.php
JediKev: Looking at it now, I think I can remove it because if you look at [this code snippet](https://github.com/osTicket/osTicket/blob/develop-next/include/ajax.tickets.php#L650-L654) we return "empty" if there is no `$clean`.

include/ajax.tickets.php
```diff
@@ -641,6 +641,10 @@ function editField($tid, $fid) {
                         if (strlen($clean) > 200)
                              $clean = Format::truncate($clean, 200);
                         break;
+                    case $field instanceof BooleanField:
+                        $clean = $field->getClean();
+                        $clean = $field->toString($clean);
```
protich: One-liner will probably suffice here. 
```php

$clean = $field->toString($field->getClean());

```

protich (changes requested):

include/ajax.tickets.php
JediKev: Thanks, done 👍

protich (approved):

PR #5381, opened 2020-02-18
issue: New Task Alert

JediKev:
This addresses an issue where the New Task Alert was not functioning as it should. This adds a new method called `onNewTask()` that runs after a Task is created and sends out the New Task Alert (if enabled). This will take into account the Global settings as well as the Department settings for Alerts and Notices.

This addresses an issue where the `subject` and `description` vars are missing from `Task::getVar()` so these variables are never replaced in the Task templates. This adds two cases for `Task::getVar()`; case `description` that retrieves the Task Description (if any) and case `subject` that retrieves the Task Subject. This also addresses an issue where the `description` var is missing from `Task::getVarScope()` so when an Agent uses the Variable Typeahead the `%{task.description}` option does not show.

### DIFF
--- a/include/class.task.php
+++ b/include/class.task.php
@@ -1062,6 +1062,10 @@ class Task extends TaskModel implements RestrictedAccess, Threadable {
             break;
         case 'last_update':
             return new FormattedDate($this->last_update);
+        case 'description':
+            return Format::display($this->getThread()->getVar('original') ?: '');
+        case 'subject':
+            return Format::htmlchars($this->getTitle());
         default:
             if (isset($this->_answers[$tag]))
                 // The answer object is retrieved here which will
@@ -1084,6 +1088,7 @@ class Task extends TaskModel implements RestrictedAccess, Threadable {
             'dept' => array(
                 'class' => 'Dept', 'desc' => __('Department'),
             ),
+            'description' => __('Description'),
             'due_date' => array(
                 'class' => 'FormattedDate', 'desc' => __('Due Date'),
             ),
@@ -1344,7 +1349,12 @@ class Task extends TaskModel implements RestrictedAccess, Threadable {
 
         // Create a thread + message.
         $thread = TaskThread::create($task);
-        $thread->addDescription($vars);
+        $desc = $thread->addDescription($vars);
+        // Set the ORIGINAL_MESSAGE Flag if Description is added
+        if ($desc) {
+            $desc->setFlag(ThreadEntry::FLAG_ORIGINAL_MESSAGE);
+            $desc->save();
+        }
 
 
         $task->logEvent('created', null, $thisstaff);
--- a/include/class.task.php
+++ b/include/class.task.php
@@ -1366,9 +1366,74 @@ class Task extends TaskModel implements RestrictedAccess, Threadable {
             $task->assign($form, $_errors);
         }
 
+        $task->onNewTask();
+
         Signal::send('task.created', $task);
 
         return $task;
+    }
+
+    function onNewTask($vars=array()) {
+        global $cfg, $thisstaff;
+
+        if (!$cfg->alertONNewTask() // Check if alert is enabled
+            || !($dept=$this->getDept())
+            || ($dept->isGroupMembershipEnabled() == Dept::ALERTS_DISABLED)
+            || !($email=$cfg->getAlertEmail())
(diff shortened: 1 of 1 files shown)


include/class.task.php
```diff
@@ -1062,6 +1062,13 @@ function getVar($tag) {
             break;
         case 'last_update':
             return new FormattedDate($this->last_update);
+        case 'description':
+            return Format::display($this->getThread()->getEntries()->filter(array(
+                'type' => 'M',
+                'pid' => 0,
+            ))->first()) ?: '';
```
JediKev: I think instead of:
```
return Format::display($this->getThread()->getEntries()->filter(array(
    'type' => 'M',
    'pid' => 0,
))->first()) ?: '';
```
I can set the `FLAG_ORIGINAL_MESSAGE` flag when adding the Task Description (if provided) and then select the entry with the flag set like:
```
$entry = $this->getThread()->getEntries()->filter(array(
    'type' => 'M',
    'flags__hasbit' => ThreadEntry::FLAG_ORIGINAL_MESSAGE,
    ))
    ->first();
return $entry ? Format::display($entry->getBody()) : '';
break;
```

I think this will be better than filtering all threads by `type` and `pid`. Any thoughts?

include/class.task.php
```diff
@@ -1366,11 +1374,77 @@ static function create($vars=false) {
...
+            }
+            $alert = $this->replaceVars($msg, array('recipient' => $staff));
+            $email->sendAlert($staff, $alert['subj'], $alert['body'], null, $options);
+            $sentlist[$staff->getEmail()] = 1;
+        }
+
+        // Alert Admin ONLY if not already a staff
+        if ($cfg->alertAdminONNewTask()
+                && !in_array($cfg->getAdminEmail(), $sentlist)) {
+                //&& ($dept->isGroupMembershipEnabled() != Dept::ALERTS_DISABLED)) {
```
aydreeihn: Looks like you can go ahead and get rid of this since you did the check on line 1389

JediKev: lol that was embarrassing.. thank you 🤣

JediKev:
I'm going to make a patch that is parallel with this pull. The patch will check the Task Details/Title and Task Details/Description flags to ensure:
- Title field has `FLAG_AGENT_REQUIRED`, `FLAG_AGENT_VIEW`, `FLAG_MASK_REQUIRE` and `FLAG_MASK_VIEW` flags set (to ensure Title is Visible/Required for Agents and to ensure no one can change this behavior)
- Description field has `FLAG_AGENT_VIEW` and `FLAG_MASK_VIEW` flags set (to ensure Description is Visible for Agents and to ensure no one can change this behavior)

include/class.task.php
```diff
@@ -1062,6 +1062,14 @@ function getVar($tag) {
             break;
         case 'last_update':
             return new FormattedDate($this->last_update);
+        case 'description':
+            $entry = $this->getThread()->getEntries()->filter(array(
```
protich: This duplicates what's already in `class.thread.php` 

```php
return Format::display($this->getThread()->getVar('original') ?: '');
```

include/class.task.php
```diff
@@ -1084,6 +1092,7 @@ static function getVarScope() {
             'dept' => array(
                 'class' => 'Dept', 'desc' => __('Department'),
             ),
+            'description' => 'Description',
```
protich: This needs to be translated.

include/class.task.php
```diff
@@ -1366,11 +1380,76 @@ static function create($vars=false) {
...
+            }
+            $alert = $this->replaceVars($msg, array('recipient' => $staff));
+            $email->sendAlert($staff, $alert['subj'], $alert['body'], null, $options);
+            $sentlist[$staff->getEmail()] = 1;
+        }
+
+        // Alert Admin ONLY if not already a staff
+        if ($cfg->alertAdminONNewTask()
+                && !in_array($cfg->getAdminEmail(), $sentlist)) {
+            $alert = $this->replaceVars($msg, array('recipient' => 'Admin'));
```
protich: Perhaps translate "Admin"?

protich (changes requested):

include/class.task.php
JediKev: can't believe I missed that, thank you

protich (approved):

Issue #5313, opened 2020-01-03
Can't merge tickets from ticket view


berkenpies:
We're having difficulties merging tickets.

Problems occurs when you go directly to a ticket and use the merge ticket function from the options menu.

**Steps to reproduce:**

- Click on any ticket in the "Open" ticket queue. Haven't tried it with other queues but I assume problem is most likely not related to queues.
 
- Go to "Merge tickets" (Option/Configuration icon -> "Merge tickets").
 ![ticket_merge](https://user-images.githubusercontent.com/21110769/71714483-61981c80-2e0e-11ea-8a72-e95dca2fcebf.png)
 
- Use "Select ticket" to find an other ticket you wish to merge with and click "Add a ticket". Now you have 2 tickets in your overview.
![ticket_merge_2tickets](https://user-images.githubusercontent.com/21110769/71714730-4a0d6380-2e0f-11ea-9098-1e9f7407742e.png)

- Click "Save changes".

Nothing will happen after that. Saving results in a (silent) 500 error and osTicket will redirect you to the ticket overview.

The actual error from the PHP-FPM error log:

`2020/01/03 09:11:05 [error] 29388#0: *231869 FastCGI sent in stderr: "PHP message: PHP Fatal error:  Uncaught Error: Call to a member function getMergeType() on null in <osticket_path>/include/class.ticket.php:2531
Stack trace:
#0 <osticket_path>/include/ajax.tickets.php(389): Ticket::merge(Array)
#1 <osticket_path>/include/class.dispatcher.php(145): TicketsAjaxAPI->updateMerge('<id>')
#2 <osticket_path>/include/class.dispatcher.php(38): UrlMatcher->dispatch('<id>/merge', Array)
#3 <osticket_path>/include/class.dispatcher.php(120): Dispatcher->resolve('<id>/merge', Array)
#4 <osticket_path>/include/class.dispatcher.php(38): UrlMatcher->dispatch('/tickets/<id>/...', NULL)
#5 <osticket_path>/scp/ajax.php(308): Dispatcher->resolve('/tickets/<id>/...')
#6 {main}`

Using the merge function from the (open) ticket overview seems to be working fine (yay!). With this I mean selecting tickets by using the checkboxes and the brand new merge icon:
![ticket_merge_from_overview](https://user-images.githubusercontent.com/21110769/71715120-9c9b4f80-2e10-11ea-9ce6-e8c174d424c8.png)

**Server components (if needed):**
nginx/1.17.7
[...]

Issue #5296, opened 2019-12-21
Search input should be trim()'d to avoid confusion.


Mike-Benoit:
### Description

When using the search, especially when searching for an exact ticket number, if a space is mistakenly added to the beginning or end of the search input, which is hard to tell if it has even occurred and can be common when using copy&paste, it will change the search results. 

### Steps to Reproduce

1. Search for " 123456" or "123456 " without quotes (notice the spaces)

2. Search for "123456" without quotes, assuming that ticket exists it will take you directly to the ticket itself. 

**Expected behavior:** [What you expected to happen]

Search results should be the same regardless of leading or trailing whitespace.

### Versions
v1.12 on Ubuntu 18.04

JediKev:
@Mike-Benoit 

This is not intuitive as people who want to use spaces would not ever be able to use spaces.

By this I mean using spaces for relevance searching.

Cheers.

Mike-Benoit:
I'm curious what the use case is for intentionally putting a space at the beginning or end of a search query? 

Also if they wanted to do that, they could add quotes around it still, and running trim( $search_query ) wouldn't have any effect then.

JediKev:
@Mike-Benoit 

>I'm curious what the use case is for intentionally putting a space at the beginning or end of a search query?

What I meant was that spaces should be allowed _between_ words. I interpreted your initial post wrong.

So, with latest `develop` (`1.12.x` series) I'm not able to replicate this issue. I can see the same results regardless of whitespaces at beginning or end:
#### No Whitespaces
<img width="270" alt="Screen Shot 2020-03-02 at 10 12 43" src="https://user-images.githubusercontent.com/11823401/75694728-a275ab80-5c6e-11ea-9f76-05c95e17e919.png">

#### Whitespaces At Beginning
<img width="274" alt="Screen Shot 2020-03-02 at 10 12 05" src="https://user-images.githubusercontent.com/11823401/75694743-a4d80580-5c6e-11ea-9ace-2c22a989321c.png">

#### Whitespaces At End
<img width="275" alt="Screen Shot 2020-03-02 at 10 12 31" src="https://user-images.githubusercontent.com/11823401/75694746-a6a1c900-5c6e-11ea-91dc-c14e5e2ae772.png">

[...]

PR #5386, opened 2020-02-27
issue: Duplicate Tickets in Lookup

JediKev:
This addresses issue #5313 where duplicate tickets appear in the ticket lookup results; especially when looking for a ticket to merge. This is due to a discrepancy between both queries looking for hits on the search criteria. The first query that looks for hits pushed the `number` column towards the end of the results and the second query has it as the second column in the results. When we do a `UNION ALL` these values are mapped incorrectly and it appears as there are 2 results instead of one. This adds a new method called `lookupByNumber()` that only finds results for number searches. This also changes the lookup URL for Ticket Merge from `/lookup` to `/number-lookup` so that we only search for numbers. This also reformats `lookup()` so that if the search query is a number we will pass to `lookupByNumber()` and merge the results with `UNION ALL`.

This also addresses issue #5296 where having whitespaces at the beginning or end of a ticket number lookup (eg. `<whitespaces>643906` or `643906<whitespaces>`) will not show expected results. This is due to the search query looking for `ticket.number LIKE '    643906%'` explicitly which will not find results. This updates the Number Lookup method to trim all white spaces from number criteria so the `LIKE` comparison operator works as expected and finds the appropriate results.


JediKev:
Just noticed this breaks the counts on the typeaheads for user's tickets...... 😞 Will have to figure out another way around this.

JediKev:
Updated, works in both cases now. 👍

PR #5398, opened 2020-03-06
issue: PasswordField Validation

JediKev:
This addresses an issue reported in the Plugins repo where attempting to use a password that starts with any of these characters (`= - + @`) fails with `Content cannot start with the following characters: = - + @` validation error. This is due to Formula Injection prevention for any field that has potential to be exported. I did not take PasswordFields into account as these are not exportable fields. This adds a new validator to TextboxField called `password` that calls `Validation::check_passwd` which uses `PasswordPolicy::checkPassword()` to check the password against all active policies. This also adds a constructor to class `PasswordField` so we can add the `password` validator for all password fields. In addition this adds another validator to TextboxField called `noop` that returns true to skip validation (useful for LDAP password, etc.).

This adds the ability to use error messages returned by the source validation method for more detailed errors via TextboxFields (and it's children classes). If no error is returned by the source validation method we will use the default error message set by `validateEntry()`.

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -1446,6 +1446,9 @@ class TextboxField extends FormField {
         $config = $this->getConfiguration();
         $validators = array(
             '' => '',
+            'noop' => array(
+                function($a, &$b) { return true; }
+            ),
             'formula' => array(array('Validator', 'is_formula'),
                 __('Content cannot start with the following characters: = - + @')),
             'email' =>  array(array('Validator', 'is_valid_email'),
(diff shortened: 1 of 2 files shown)


include/class.forms.php
```diff
@@ -1446,6 +1446,9 @@ function validateEntry($value) {
         $config = $this->getConfiguration();
         $validators = array(
             '' => '',
+            'noop' => array(
+                function() { return true; }
```
```php
function($a, &$b)
```

JediKev: done

Issue #5376, opened 2020-02-12
ticket_cdata table is not filled correctly for choice fields


TrueType:
Story: 
Customer complains about the CSV export of ticket data and the missing selection field values. The data was there before but got lost for the recent weeks. No clue, why the data got messy somehow along the way. The columns had 0 values instead of "key,value" glued strings.

Took me a longer time got to get into the matter. The ticket_cdata table gets deleted as soon as one changes the form records / form definition. It then gets recreated and filled as whatever place you may touch of the osTicket application. The values then get filled in correctly(copies of form_entry_values table). So the messy data just started the last time the form got updated. The ticket::create does not fill the cdata columns correctly for selection but enters the "0" values only.

PR
#5375

This may relate with #4071 - but I doubt it is fixed there. 

Comments:
- the codebase is a mess. sry to say so.
- within class.dynamic.forms.php is a class SelectionField which probably just is a copy of class ChoiceField in class.forms.php. 
- there should not be two pieces of code writing the cdata table.
- https://forum.osticket.com/d/95811-data-not-saving-correctly-to-ost-ticket-cdata

JediKev:
@TrueType 

Did you try doing a recursive search on Github before making this post? We have a fix here:
- https://github.com/osTicket/osTicket/pull/5258

Cheers.

TrueType:
@JediKev I really don't think so. Please have a look at getKeys() function. https://github.com/osTicket/osTicket/blob/develop/include/class.forms.php#L1847
This functions works on a "choice" record, which is an assoziative array in the current code. It was an ordinary array in older versions (>2years ~1.10.1) and where the "0" values came from. Now with the lastest code you may get the correct key of the option selected. But you still dont get anything worth to be glued together.

JediKev:
@TrueType 

We will have to look further into this to confirm.

Cheers.

JediKev:
@TrueType 

Re-tested with latest `develop-next` _(14:06 PM CST March 9, 2020)_ and the cdata table saves the key(s) for the choice(s) like `key1` if one choice is selected or `key1, key2` if multiple choices are selected. We then explode the keys via `, ` (with this fix: #5399) and lookup the value for each key. In the Ticket Header, if multiple choices are selected, will appear as `value1, value2` (as it should). In the Queue Column, if multiple choices are selected, will appear as `value1, value2` (as it should). Lastly in the Ticket Exports, if multiple choices are selected, will appear as `value1, value2` (as it should).
[...]

PR #5399, opened 2020-03-09
issue: Multiple Choice Export

JediKev:
This addresses an issue where saving multiple choices for a ChoiceField does not show all values in Queue Columns nor Ticket Exports. This is due to the `to_php()` method exploding on `,` while the keys are saved like `key1, key2` (with comma then space). This updates the `to_php()` method to explode on `, ` (comma and space) so we get the proper keys for the value lookups.
(Partially addresses #5376)

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -1817,7 +1817,7 @@ class ChoiceField extends FormField {
         if (is_string($value) && strpos($value, ',')) {
             $values = array();
             $choices = $this->getChoices();
-            $vals = explode(',', $value);
+            $vals = array_map('trim', explode(',', $value));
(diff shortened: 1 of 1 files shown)


include/class.forms.php
```diff
@@ -1817,7 +1817,7 @@ function to_php($value) {
         if (is_string($value) && strpos($value, ',')) {
             $values = array();
             $choices = $this->getChoices();
-            $vals = explode(',', $value);
+            $vals = explode(', ', $value);
```
```php
$vals = array_map('trim', explode(',', $value));
```

JediKev: done

PR #5401, opened 2020-03-10
lint: Uninitialized Matches

JediKev:
This resolves lint test complaints of an uninitialized variable `$matches` in `include/class.mailparse.php`.

### DIFF
--- a/include/class.mailparse.php
+++ b/include/class.mailparse.php
@@ -386,7 +386,7 @@ class Mail_Parse {
     }
 
     function getAttachments($part=null){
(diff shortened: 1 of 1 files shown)


include/class.mailparse.php
```diff
@@ -387,6 +387,7 @@ function mime_encode($text, $charset=null, $encoding='utf-8') {
 
     function getAttachments($part=null){
         $files=array();
+        $matches=array();
```
protich:
```php
$files = $matches = array();
```

PR #5403, opened 2020-03-10
Feature: Agent/Department Visibility

aydreeihn:
This commit adds a new feature that allows admins to limit the Departments or Agents that are visible in the Agent Panel based on new permissions. The permissions can be found here:

Admin Panel | Agents | Click an Agent | Permissions Tab | Miscellaneous

Screen Shot 2020-03-10 at 3 14 46 PM

These boxes will be checked by default for new agents as well as existing agents (in the case of an upgrade).

You can check out the documentation [here](https://github.com/osTicket/docs/blob/945c7d54c800f55eb297d8b3b27337f8122954c7/Features/Visibility%20Permissions.rst).


include/class.canned.php
```diff
@@ -229,11 +229,25 @@ static function getIdByTitle($title) {
...
             ->order_by('title')
             ->values_flat('canned_id', 'title');
 
+        if ($thisstaff) {
+            $departments = array();
+            //get responses for all depts
+            $departments[] = 0;
+            //get the responses for the agent's departments
+            $staffDepts = $thisstaff->getDepartmentNames(true);
+            foreach ($staffDepts as $k => $v)
```
protich: Instead of foreach look I would suggest then add the zero key thereafter.
```php
$departments = array_keys($thisstaff->getDepartmentNames(true));
```

PR #5405, opened 2020-03-10
[Feature] Canned responses search - Matcher for non-adjacent keywords

### LINKED ISSUE #5367: Search Canned answers (only exact match) (Improvement ?)
### Prerequisites

* [ ] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [ ] Do you have any errors in the PHP error log, or javascript console?
* [x] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [x] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

### Description
Hi,
When we need to find a canned answer, we need to write words in the exact order.
In our case we have a lot of canned answers, and we would like to search by using 2 or 3 words that appear in the title but not necessary in the exact order of the search (search each word separately)

Zaahed:
Resolves #5367 

This custom matcher will return results for non-adjacent keywords in the canned responses search field (ticket view). Makes searching for canned responses easier when you don't know the exact order of the words, and when you have a large list of canned responses.

### DIFF
--- a/scp/js/scp.js
+++ b/scp/js/scp.js
@@ -214,14 +214,31 @@ var scp_prep = function() {
                 return data;
             }
 
-            keywords = (params.term).split(" ");
+            var original = data.text.toUpperCase();
+            var term = params.term.toUpperCase();
+            var keywords = term.split(" ");
 
             for (var i = 0; i < keywords.length; i++) {
-                if (((data.text).toUpperCase()).indexOf((keywords[i]).toUpperCase()) == -1)
+                if (original.indexOf(keywords[i]) == -1)
                     return null;
             }
 
-            return data;
+            var termIndex = original.indexOf(term);
+            if (termIndex > -1)
+                return $.extend({'matchType': 'term', 'index': termIndex}, data);
+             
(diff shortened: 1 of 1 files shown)


JediKev:
@Zaahed 

The problem with this is that the results are not sorted properly. With this applied, exact matches are on bottom where they should be the first results. I'm sure a lot of people will be upset that the exact matches are not priority in the result set (how it is currently). If you could find a way to sort the results properly I would see this as a legitimate pull.

Cheers.

Zaahed:
Hi @JediKev 

Thanks for your feedback. I've pushed a new commit which sorts the results and shows the exact matches at the top.

JediKev:
We should do this at database level instead of in JS.

Zaahed:
I assume this means we should use an AJAX request similar to that of the lookup function used in the ticket list search bar.

However this also means that there will be a small delay (150+ ms) because of the AJAX request, whereas the JS returns the results instantly at each keystroke.

corent56:
Maybe it should be a good idea that when we search for a character like é or è, à... we can see all the answer with or without accent (e->e-é-è-ê ...)

PR #5418, opened 2020-03-17
export: User/Org Tickets

JediKev:
This addresses an issue introduced with #5207 where reverting the changes of a commit left some old code behind. This causes the User Ticket exports and Org Ticket exports to crash when clicking Export. This changes both `filter()` methods to setting `$this->_queue->config` directly as `filter()` no longer exists for AdhocSearches.


protich:
Can you please reference to commit that was reverted?

PR #5420, opened 2020-03-17
i18n: TextareaField Placeholder

JediKev:
This addresses an issue where TextareaFields do not show translated placeholders. This is due to using the placeholder from the default configuration instead of using the local placeholder. This updates the `TextareaWidget::render()` method to get the local placeholder instead of using the default from config. This also passes `$config['placeholder']` as a default in case there is no translation available.

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -4023,7 +4023,7 @@ class TextareaWidget extends Widget {
         <span style="display:inline-block;width:100%">
         <textarea <?php echo $rows." ".$cols." ".$maxlength." ".$class
                 .' '.Format::array_implode('=', ' ', $attrs)
-                .' placeholder="'.$config['placeholder'].'"'; ?>
+                .' placeholder="'.$this->field->getLocal('placeholder', $config['placeholder']).'"'; ?>
             id="<?php echo $this->id; ?>"
             name="<?php echo $this->name; ?>"><?php
                 echo Format::htmlchars($this->value);


protich:
Why not handle this at parent class level (`FormField`) - I can see `TextboxField` being an issue as well.

JediKev:
Need to move this to render instead.

JediKev:
Updated

include/class.forms.php
```diff
@@ -4023,7 +4023,7 @@ function render($options=array()) {
         <span style="display:inline-block;width:100%">
         <textarea <?php echo $rows." ".$cols." ".$maxlength." ".$class
                 .' '.Format::array_implode('=', ' ', $attrs)
-                .' placeholder="'.$config['placeholder'].'"'; ?>
+                .' placeholder="'.$this->field->getLocal('placeholder').'"'; ?>
```
protich: This assumes the placeholder is translated -  use config placeholder as fallback.

```php
$this->field->getLocal('placeholder', $config['placeholder'])
```

protich (changes requested):

include/class.forms.php
JediKev: Well, `getLocal()` should return the default if no translation is available via [this line](https://github.com/osTicket/osTicket/blob/develop/include/class.forms.php#L1331):
```
return $T != $tag ? $T : ($default ?: $this->get($subtag));
```

In the above line, we do `$default ?: $this->get($subtag)` (where `$default` is the default and `$subtag` is `placeholder` in this case) if no translation is available which should return the default English translation.

JediKev: But I get what you mean, in order to not refetch then we can pass a default

JediKev: updated

protich (approved):

PR #5421, opened 2020-03-17
issue: DynamicForm i18n Instructions Decode

JediKev:
This addresses an issue where saving translated Form Instructions with HTML formatting renders the Instructions as plaintext on the Client Portal showing all HTML, etc. This is due to saving the translated Instructions as encoded thanks to `Format::htmlchars()`. This uses `Format::htmldecode()` for `instructions` in `DynamicForm::getForm()` so that when the Form is rendered the Instructions are properly displayed.


protich:
Question here is why did we encode the content initially - let's have a broader discussion on the approach when you get a moment.

JediKev:
Updated

PR #5426, opened 2020-03-21
[Feature] Unread ticket message (column annotation)

mirecad:
**FEATURE DESCRIPTION**
Staff members needs to be notified about new messages in ticket thread. One already existing way is email notification. Other very simple solution could be column annotation in tickets overview table.
image
This green plus sign indicates, that there is new message in ticket thread, that I have not read yet. If I open this ticket in browser, this annotation disappears.

This became must have feature for staff in my companies. Our team works with quite a lot of tickets daily. They get lost in email notifications. Staff members rely on this simple and effective feature.

image

Feature can be turn off/on in queue column settings (Admin panel - Settings - Tickets - Queues - Columns - Config). Yes, I see, that the icon is the same as near "Add a annotation". I leave icon selection up to maintainers (see TODO), because I did not want to include additional icons into project.

**TODO**
- Choose correct icon for annotation ticket table (I used scp - icons - add.png)
- Choose coorect icon in column settings (I used plus-sign)
- Maybe add "Mark as unread" possibility on ticket page. Ideal place would be under settings icon dropdown menu. I can implement this, if anybody is interested.

**DATABASE CHANGES**
```
CREATE TABLE `ost_ticket_stafflastvisit` (
  `ticket_id` int(11) unsigned NOT NULL,
  `staff_id` int(11) unsigned NOT NULL,
  `lastvisit_date` datetime NOT NULL DEFAULT '2015-01-01 12:00:00',
  PRIMARY KEY (`ticket_id`,`staff_id`),
  KEY `FK_staff_id` (`staff_id`)
);
```

### DIFF
--- a/include/class.queue.php
+++ b/include/class.queue.php
@@ -704,7 +704,7 @@ class CustomQueue extends VerySimpleModel {
                 "width" => 250,
                 "bits" => QueueColumn::FLAG_SORTABLE,
                 "filter" => "link:ticket",
-                "annotations" => '[{"c":"TicketThreadCount","p":">"},{"c":"ThreadAttachmentCount","p":"a"},{"c":"OverdueFlagDecoration","p":"<"}]',
+                "annotations" => '[{"c":"TicketThreadCount","p":">"},{"c":"ThreadAttachmentCount","p":"a"},{"c":"OverdueFlagDecoration","p":"<"},{"c":"TicketUnreadDecoration","p":"<"}]',
                 "conditions" => '[{"crit":["isanswered","nset",null],"prop":{"font-weight":"bold"}}]',
                 "truncate" => 'ellipsis',
             )),
@@ -1742,6 +1742,31 @@ extends QueueColumnAnnotation {
 
     function isVisible($row) {
         return $row['isoverdue'];
+    }
+}
+
+class TicketUnreadDecoration
+extends QueueColumnAnnotation {
+    static $icon = 'plus-sign';
+    static $desc = 'Unread message icon';
+
+    function annotate($query, $name=false) {
+        global $thisstaff;
+
+        $query->values('updated');
+        return $query
+            ->annotate(array(
+                '_lastVisit' => TicketStaffLastVisit::objects()
+                    ->filter(array(
+                        'ticket_id' => new SqlField('ticket_id', 1),
+                        'staff_id' => $thisstaff->getId())) 
+                    ->values('lastvisit_date')        
+                    ));
+    }
+
+    function getDecoration($row, $text) {
+        if(strtotime($row['_lastVisit']) < strtotime($row['updated']))
+            return sprintf('<span class="Icon unreadTicket"></span>');
     }
 }
 
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -1313,6 +1313,23 @@ implements RestrictedAccess, Threadable, Searchable {
         return $this->save();
     }
 
+    //mark as read for given staff
+    function setStaffLastVisitNow($staff) {
+        $stafflastvisit = TicketStaffLastVisit::lookup(array(
+            'ticket_id'=>$this->ticket_id,
+            'staff_id'=>$staff->getId()
+        ));        
+        if(!isset($stafflastvisit)) {
+            $stafflastvisit = new TicketStaffLastVisit(array(
+                'ticket_id' => $this->ticket_id,
+                'staff_id' => $staff->getId(),
+                'lastvisit_date' => SqlFunction::NOW()
+            ));
+        }
+        $stafflastvisit->lastvisit_date = SqlFunction::NOW();
+        $stafflastvisit->save();        
+    }
+
     // Ticket Status helper.
     function setStatus($status, $comments='', &$errors=array(), $set_closing_agent=true) {
         global $thisstaff;
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -4343,3 +4343,19 @@ class TicketCData extends VerySimpleModel {
     );
 }
 TicketCData::$meta['table'] = TABLE_PREFIX . 'ticket__cdata';
+
+class TicketStaffLastVisit extends VerySimpleModel {
+    static $meta = array(
+        'pk' => array('ticket_id', 'staff_id'),
+        'select_related' => array('staff'),
+        'joins' => array(
+            'ticket' => array(
+                'constraint' => array('ticket_id' => 'Ticket.ticket_id'),
+            ),
+            'staff' => array(
+                'constraint' => array('staff_id' => 'Staff.staff_id')
+            ),
+        ),
+    );
+}
+TicketStaffLastVisit::$meta['table'] = TABLE_PREFIX . 'ticket_stafflastvisit';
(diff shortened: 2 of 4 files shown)


aydreeihn:
@mirecad Thanks so much for working on this feature! We love getting feedback on how to improve our software, and I think a lot of people will be able to make good use of this. I pulled your code and tested it out, and I can confirm that it's working well. For the ToDos, I think the icons you picked make sense. 

I made a few changes to reduce the lines of code needed. Also, the annotation was being shown any time the Ticket was updated at all which included having a new message, response, or even field edit. You can look at 'thread__lastmessage' rather than 'updated' in class.queue.php to show the annotation specifically for new User messages in a Ticket.

Please see the diff below:

```
diff --git a/include/class.queue.php b/include/class.queue.php
index 0fe76d58..c9b47d3f 100644
--- a/include/class.queue.php
+++ b/include/class.queue.php
@@ -1753,19 +1753,19 @@ extends QueueColumnAnnotation {
     function annotate($query, $name=false) {
         global $thisstaff;
 
-        $query->values('updated');
+        $query->values('thread__lastmessage');
         return $query
             ->annotate(array(
                 '_lastVisit' => TicketStaffLastVisit::objects()
                     ->filter(array(
                         'ticket_id' => new SqlField('ticket_id', 1),
-                        'staff_id' => $thisstaff->getId())) 
-                    ->values('lastvisit_date')        
+                        'staff_id' => $thisstaff->getId()))
+                    ->values('lastvisit_date')
                     ));
     }
 
     function getDecoration($row, $text) {
-        if(strtotime($row['_lastVisit']) < strtotime($row['updated']))
+        if(strtotime($row['_lastVisit']) < strtotime($row['thread__lastmessage']))
             return sprintf('<span class="Icon unreadTicket"></span>');
     }
 }
diff --git a/include/class.ticket.php b/include/class.ticket.php
index c5368ff2..e0656a59 100644
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -1314,20 +1314,20 @@ implements RestrictedAccess, Threadable, Searchable {
     }
 
     //mark as read for given staff
-    function setStaffLastVisitNow($staff) {
-        $stafflastvisit = TicketStaffLastVisit::lookup(array(
+    function setStaffLastVisit() {
+        global $thisstaff;
+
+        $info = array(
             'ticket_id'=>$this->ticket_id,
-            'staff_id'=>$staff->getId()
-        ));        
-        if(!isset($stafflastvisit)) {
-            $stafflastvisit = new TicketStaffLastVisit(array(
-                'ticket_id' => $this->ticket_id,
-                'staff_id' => $staff->getId(),
-                'lastvisit_date' => SqlFunction::NOW()
-            ));
-        }
+            'staff_id'=>$thisstaff->getId());
+
+        $stafflastvisit = TicketStaffLastVisit::lookup($info);
+
+        if(!$stafflastvisit)
+            $stafflastvisit = new TicketStaffLastVisit($info);
+
         $stafflastvisit->lastvisit_date = SqlFunction::NOW();
-        $stafflastvisit->save();        
+        $stafflastvisit->save();
     }
 
     // Ticket Status helper.
diff --git a/include/staff/ticket-view.inc.php b/include/staff/ticket-view.inc.php
index d345a192..62222a16 100644
--- a/include/staff/ticket-view.inc.php
+++ b/include/staff/ticket-view.inc.php
@@ -23,7 +23,7 @@ $id    = $ticket->getId();    //Ticket ID.
 $isManager = $dept->isManager($thisstaff); //Check if Agent is Manager
 $canRelease = ($isManager || $role->hasPerm(Ticket::PERM_RELEASE)); //Check if Agent can release tickets
 $canMarkAnswered = ($isManager || $role->hasPerm(Ticket::PERM_MARKANSWERED)); //Check if Agent can mark as answered/unanswered
-$ticket->setStaffLastVisitNow($thisstaff); //reset last visit time for read/unread mod
+$ticket->setStaffLastVisit(); //reset last visit time for read/unread mod
 
 //Useful warnings and errors the user might want to know!
 if ($ticket->isClosed() && !$ticket->isReopenable())
```

I noticed that you added a new table for this feature. Since this would require a database patch, we would need you to rebase your branch and make the pull request for develop-next rather than develop so that the upgrader would work correctly.

Also, I noticed that the new table would be written to regardless of if the annotation is being used or not. While the data stored is pretty minimal, this table could potentially become really large depending on the help desk's size. On the develop-next branch, we started using flags for Tickets, and I was thinking maybe we could think about setting a flag depending on if the annotation is being used to determine if we should write to the ticket_stafflastvisit. The only thing about that though is that if someone does decide to start using this feature, every Ticket would have that icon beside it for Agents, but that would be the case anyway for people upgrading since the table would start off empty.

To clarify a little bit, when this annotation is being used, each Agent will see the annotation if they have not gone to the Ticket and there is a new message.

@protich @JediKev What do you guys think?

JediKev:
@aydreeihn 

I agree, I don't think it would be necessary to add a new table just for this annotation. There has to be a way around this via flags, etc.

Cheers.

aydreeihn:
@JediKev Well with it being implemented based on Agent, I think it does have to have a table because it's showing Tickets that the specific logged in Agent hasn't read. If we made it to where it only had the annotation if there's a new message and NO Agent has opened the Ticket, that could work. Then we could set the flag as on in the onMessage function and turn it off in ticket-view.inc.php.

JediKev:
@aydreeihn True true.

mirecad:
@aydreeihn @JediKev 

>  If we made it to where it only had the annotation if there's a new message and NO Agent has opened the Ticket, that could work.

This should be really **based on each agent**. I need information that there is something new in the ticket, that **I'm not aware of**. It should indicate "Open the ticket an look what is new there if you are interested.". This is my main goal. I don't care if "There are some news in the ticket but somebody already checked it.". (Talking from my 4 year experience as agent/leader of helpdesk team with 40k tickets).

> Also, I noticed that the new table would be written to regardless of if the annotation is being used or not.

If admins allows this feature, agents want to see it working on old tickets also. That's why we want to track this as soon as possible. Even if the feature is not used. 
How to deal with existing tickets in feature release is another question (maybe start tracking views into DB and let users know avout this feature later).

sevmonster:
This looks like an incredible addition.
> This should be really **based on each agent**. I need information that there is something new in the ticket, that **I'm not aware of**.

I completely agree with this. To put my own hat in the ring, our previous ticketing system only had a fields that showed when the ticket was last changed and also by source (agent name if they used the UI, "email system" if they sent in an email reply which is very ambiguous esp. if there are multiple people replying). This means that:
- Agents cannot see if the ticket was read or somehow already managed, just that it was updated. These updates could be irrelevant metadata changes instead of useful dialogue between agent and requestor. This makes it difficult to sort by update time.
- Agents spend more time than necessary looking over tickets, because they might have forgotten they've already read/managed it and want to be sure everyone is up-to-date. Our environment is short-staffed so this happens a lot.

Giving a simple indicator that a ticket has new pertinent information *without requiring sorting* means agents can have their tickets sorted in a way that works for them while still being kept up-to-date.

JediKev:
@mirecad 

Thank you for the pull request. We have reviewed this and while the changes you've made are great we envision this a little differently and would like to add to it. We will add this to our Feature Request list.

Cheers.

Issue #5428, opened 2020-03-23
Changing manager of a department requires more than 2G of memory


iprok:
### Prerequisites

* [ ] Can you reproduce the problem in a fresh installation of the "develop" branch?
Haven't tried.

* [x] Do you have any errors in the PHP error log, or javascript console?

`Mar 23 21:57:37 help ool help: PHP Fatal error:  Allowed memory size of 2147483648 bytes exhausted (tried to allocate 20480 bytes) in /var/www/help/include/class.orm.
php on line 3478`

* [x] Did you check the [osTicket forums](https://forum.osticket.com/)?
Yes

* [x] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

Yes

### Description

Changing the Manager of "Archive Requests" department takes to much time and memory.

### Steps to Reproduce

1. Open osticket/scp/departments.php?id=15 (id=15 is "Archive Requests" in our system)

2. Change "Manager" to any other. I've tried changing from current value to None.

3. Save changes.

**Expected behavior:** [What you expected to happen]

Changes are applied quickly and correctly.

**Actual behavior:** [What actually happened]

Waiting for a long time (more than 20 seconds) then 500 error with "Allowed memory size exhausted error in logs"

### Versions

osTicker 1.14.1
Debian Stretch
PHP 7.0.33-0+deb9u7

I've tried memory_limit setting from 128M till 2GB without any success. We have approx 7000 tickets in our system.

JediKev:
@iprok 

That shouldn't happen. I cannot replicate the same issues with v1.14.1 and PHP 7.3. Take a look at your error logs (Apache Error Logs, PHP Error Logs, MySQL Error Logs, osTicket System Logs, etc.) to see if you can find any related errors.

Cheers.

iprok:
Hello! I've already mentioned the message from logs that is related to the situation:
`Mar 23 21:57:37 help ool help: PHP Fatal error: Allowed memory size of 2147483648 bytes exhausted (tried to allocate 20480 bytes) in /var/www/help/include/class.orm. php on line 3478`

JediKev:
@iprok 

That's not a full stack trace so it's really not that helpful. All that says is the memory was exhausted whilst running the `_cast()` method in class ORM, but this does not tell me anything more.

Cheers.

iprok:
Here is stack trace (from php slow log):
```
[...]

PR #5429, opened 2020-03-23
issue: Department Parent

JediKev:
This addresses issue #5428 where you can set a Department's Parent to itself, which is not supposed to happen. This adds a check before displaying available Departments to remove the current Department from the list. If you're adding a new Department this will obviously show all Departments.


protich:
Why would someone set a parent to itself is beyond me... That said, we might have to think about nested departments deeper than just the parent. In other words, a child department cannot have a parent that's in its path, otherwise we will encounter the same issue.  

We can use `getFullPath()` to check such cases.

JediKev:
Updated

PR #5431, opened 2020-03-25
Audit Plugin Modal Fix

aydreeihn:
This commit removes a line of code that caused the Ticket Audit modal to close any time you tried to go to the next page of audits or sort the audits by timestamp. The code was put in place 6 years ago for file upload fields, but they still work fine everywhere in the codebase without that line now.


aydreeihn:
It looks like we always fade the background when we have a modal now, so this change will be alright.

PR #5437, opened 2020-03-30
issue: Mass Process Add Users To Organization

JediKev:
This addresses an issue where checking the box next to a few Users' names then selecting the Action to Add To Organization does nothing. This is due to a pesky jQuery issue where `parseJSON()` will crap out if you pass in JSON (go figure). You MUST provide a string otherwise it will silently fail. So, this adds an if statement that checks if the content is a string and if so we will use `$.parseJSON()` otherwise we will not parse it.

### DIFF
--- a/include/staff/users.inc.php
+++ b/include/staff/users.inc.php
@@ -288,8 +288,9 @@ $(function() {
             $.dialog('ajax.php/orgs/lookup/form', 201, function(xhr, json) {
               var $form = $('form#users-list');
               try {
-                  var json = $.parseJSON(json),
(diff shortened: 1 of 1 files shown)


protich:
The real issue here is; we might get a string or a json object. 

I understand the fix but it's akin to turning json into a string then back to json. What about if you get a string back... 

I'll recommend the following;

```js

   if ($.type(json) == 'string')
       json = $.parseJSON(json);

```

We'll have to address inconsistent api in the future to remove the guesswork.

JediKev:
Updated

PR #5438, opened 2020-03-30
issue: Confirm Popup Promise

JediKev:
This addresses an issue where clicking Cancel on some popups (eg. Organization Deletion popup, User Deletion popup, etc.) will perform the action instead of cancelling. This is because the `then()` and `done()` callbacks used on the custom `$.confirm()` method do not respect the returned promise. This adds a check to each `$.confirm()` usage to see if the promise is equal to `FALSE`, if so we return false. If the promise is not equal to `FALSE` then we continue with the callback.

### DIFF
--- a/include/staff/orgs.inc.php
+++ b/include/staff/orgs.inc.php
@@ -219,7 +219,11 @@ $(function() {
             $form.find('#selected-count').val(ids.length);
             $form.submit();
           };
-          $.confirm(__('You sure?')).then(submit);
+          $.confirm(__('You sure?')).then(function(promise) {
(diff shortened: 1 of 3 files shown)


protich:
The issue here was jQuery changed how `then` works as of version 1.8. It used to be just a basic callback function to call on success.

JediKev:
@protich I thought so because you have `then()`, `done()`, and `fail()` which I assumed would work like AJAX where `done()` is always, `then()` is success, and `fail()` is fail but I guess they wanted to be different lol

PR #5448, opened 2020-04-08
Add WhatsApp and Chat

Crock-Linux:
Adding other sources to the ticket
and in the database - table - ticket
include in the source field: 'Web', 'Email', 'WhatsApp', 'Chat', 'Phone', 'API', 'Other'


protich:
Thanks for the PR.

We would prefer to have ticket sources fully implemented as a dynamic list, just like Ticket Statuses. Let me know if you want some pointers - otherwise you can follow how ticket statuses are managed.

PR #5453, opened 2020-04-09
Allow external UserAuthenticationBackend to provide a way to set an error while running the triggerAuth.

pitbulk:
Right now there is no way to notify the login page about an error that happened during the triggerAuth process.

This simple patch allows the UserAuthenticationBackend to return an AccessDenied exception that will be identified and its message added in the error var, that way, the error is prompted to the user.

### DIFF
--- a/login.php
+++ b/login.php
@@ -108,8 +108,12 @@ elseif (isset($_GET['do'])) {
     switch($_GET['do']) {
     case 'ext':
         // Lookup external backend
-        if ($bk = UserAuthenticationBackend::getBackend($_GET['bk']))
-            $bk->triggerAuth();
+        if ($bk = UserAuthenticationBackend::getBackend($_GET['bk'])) {
+            $result = $bk->triggerAuth();
+            if ($result instanceof AccessDenied) {
+                $errors['err'] = $result->getMessage();
+            }
+        }
     }
 }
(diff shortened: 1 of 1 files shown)


protich:
Just a thought - since `AccessDenied` is an exception why not throw it instead?

pitbulk:
Hi @protich,

The login.php file does not catch Exceptions/AccessDenied to store it in the error message var, so raising the exception is not an option.

I want the error message to be prompted in the login view to notify properly the user, external authentication are more complex than the simple validation of username/password, so provide to the user the right info is so important.

If you see, the idea of be able to retrieve AccessDenied from UserAuthenticationBackend methods isn't new

[Here](https://github.com/osTicket/osTicket/blob/43bf186db9e85c8ec218158ea4c0bc3e90bbe3eb/view.php#L36) you can see this TODO
```
    // TODO: Consider receiving an AccessDenied object
    $user =  UserAuthenticationBackend::processSignOn($errors, false);
```

And the AuthenticationBackend  class defined at class.auth.php, already is [using the AccessDenied comparation](https://github.com/osTicket/osTicket/blob/2cd7af46f218de0a6768a1f285f322ca1a18abae/include/class.auth.php#L300).

but the triggerAuth method does not pass the $error var as a reference, that we see on the process and processSignOn methods.

protich:
Far enough - I will merge.

PR #5460, opened 2020-04-17
revert: Agent Added Agent Collabs

JediKev:
This adds a check for `$thisstaff` in `Thread::addCollaborator()`. If we have a staff object we will enable Agent Collabs, if we do not have a staff object we will disable Agent Collabs. This will ensure all Agent added Agent Collabs are Enabled and any User added Agent Collabs are Disabled. This also updates the Help Tip for the setting to remove the `Web created tickets`part (as users cannot add collabs via web) and to clarify that this only affects User created tickets.

### DIFF
--- a/include/class.thread.php
+++ b/include/class.thread.php
@@ -168,7 +168,7 @@ implements Searchable {
     }
 
     function addCollaborator($user, $vars, &$errors, $event=true) {
-        global $cfg;
+        global $cfg, $thisstaff;
 
         if (!$user)
             return null;
@@ -183,8 +183,8 @@ implements Searchable {
             return null;
 
         $c->active = true;
-        // Disable Agent Collabs (if configured)
-        if ($this->object_type === 'T'
+        // Disable Agent Collabs (if configured) for User created tickets
+        if (!$thisstaff && $this->object_type === 'T'
                 && $cfg->disableAgentCollaborators()
                 && Staff::lookup($user->getDefaultEmailAddress()))
             $c->active = false;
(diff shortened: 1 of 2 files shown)


JediKev:
~Need to move the `$thisstaff` check to `Thread::addCollaborator()` to prevent the staff lookup altogether.~
Done

JediKev:
Updated
- Moved the `$thisstaff` check to `Thread::addCollaborator()` to avoid the `Staff::lookup()` altogether if Staff added Collab
- Updated the Help Tip to further clarify this is ONLY User created tickets
- Updated the Help Tip to remove the `Web created tickets` part as Users cannot add collabs via Web; only Agents can but since we are reverting this for Agent added collabs I'm removing this wording

include/ajax.thread.php
```diff
@@ -108,7 +108,7 @@ function addCollaborator($tid, $type=null, $uid=0) {
                                 $vars, $errors))) {
                     $info = array('msg' => sprintf(__('%s added as a collaborator'),
                                 Format::htmlchars($c->getName())));
-                    $c->setCc($c->active);
+                    $c->setCc();
```
protich: Wouldn't this change make `$c->active = false;` unnecessary (set in `class.thread.php` line 190).

JediKev: Well instead of passing the property I was leaving it empty because 1.) $active defaults to true in setCc and 2.) we are reverting this feature for agent added collabs so I removed it altogether so it always enables the collabs that agents add via the modal. With this being said I have the $thisstaff check in Thread::addCollab so I can pass it without worries because the property will always be set to true when agents are adding collabs. I can update this momentarily.

JediKev: Updated: Leaving `$c->active` as this will always be set to `true` for Agent added collabs

PR #5464, opened 2020-04-20
Display headers appropriate for selected language

isaacnia:
Fixed Bug: Display header appropriate for selected language


protich:
Thanks for the PR, not sure I agree with the change.

Since column headings are administratively translatable - `$C->getLocalHeading()` should be getting localized header. If it's not then that's the actual bug to fix... defaulting to the default translation there make more sense.

PR #5465, opened 2020-04-20
Issue: Filter Actions for Deleted Objects

aydreeihn:
This commit fixes an issue where an object could be deleted that was included in a filter action without properly accounting for the deletion. Now, if something is deleted and there is a Filter Action associated to that object, the Filter will be disabled and the Filter Action configuration will be updated to no longer include the invalid id. This has been done for the following objects:

- Help Topic
- Department
- Agent
- Team
- SLA
- Ticket Status
- Email
- Canned Response

### DIFF
--- a/include/class.filter.php
+++ b/include/class.filter.php
@@ -34,6 +34,7 @@ extends VerySimpleModel {
 
     const FLAG_INACTIVE_HT = 0x0001;
     const FLAG_INACTIVE_DEPT  = 0x0002;
+    const FLAG_DELETED_OBJECT  = 0x0004;
 
     static $match_types = array(
         /* @trans */ 'User Information' => array(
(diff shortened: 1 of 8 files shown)


include/class.canned.php
```diff
@@ -206,6 +206,9 @@ function delete(){
         $type = array('type' => 'deleted');
         Signal::send('object.deleted', $this, $type);
 
+        // Disable Filters
+        Filter::disableFilters($this);
```
protich: Instead of calling `disableFiltters` allover - why not listen to `object.deleted`?

aydreeihn:
**Note:** We need to merge PR #5466 before this one since we moved the object.deleted signal for Help Topics.

include/class.list.php
```diff
@@ -1458,6 +1458,9 @@ function delete() {
         if (!$this->isDeletable())
             return false;
 
+        // Disable Filters
+        Filter::disableFilters($this);
+
```
protich, line 1462: Why not use signal here as well? Also you have to make sure parent actually deletes before disabling it.

```php
if (!$this->isDeletable() || !parent::delete())
    return false;

// Send deleted signal 

return true;
```

include/class.filter.php
```diff
@@ -164,6 +165,59 @@ function useReplyToEmail() {
...
+                $FAClass = 'FA_SendEmail';
+                break;
+            case 'Canned':
+                $object_id = 'canned_id';
+                $FAClass = 'FA_AutoCannedResponse';
+        }
+        $id = $object->getId();
+        $actions = FilterAction::objects()
+            ->filter(array('type' => $FAClass::$type));
+        foreach($actions as $fa) {
```
protich, line 208: This will be expensive if there are lots actions of same type (think CM). Perhaps we can filter by configuration field - i.e `LIKE '%$object_id%'` - although that might be expensive as well. Just adding an observation here...

include/class.list.php
aydreeihn: I didn't call the signal here because we aren't currently auditing List Items (Including Ticket Statuses). We audit for lists when they are created, deleted, or the definition of the list is edited. I did make the change you noted for the code.

protich: Don't see why Audit becomes a factor here... signals don't exist because of audit plugin.

aydreeihn: It's just the way I was thinking about it, it made sense to use the signal for the times that we are auditing a delete as well as needing to disable filters because you could kill 2 birds with 1 stone, but since there's nothing to audit in this case, it seemed like it made sense to call disableFilters rather than to do it by using the signal.

protich: Please make the change...

protich (changes requested):

include/class.list.php
aydreeihn: Updated

include/class.filter.php
aydreeihn: Thinking on this one. I can do some research and play with it.

aydreeihn: Based on the time it takes the query to execute, it seems like on average, it goes slightly faster the way it is now, however, if we used ```LIKE '%$object_id%'```, we wouldn't have to use the $FAClass variable, so I think it'll just be a matter of preference here.

aydreeihn:
As I was testing this, I noticed that for mass process, when we enable a filter, we just use filter->save() to update it rather than filter->update() which would validate the filter_actions before enabling the filter. Should we change that to update instead?

protich (approved):

PR #5466, opened 2020-04-20
Issue: Class AuditEntry Not Found

aydreeihn:
This commit fixes an issue where we were trying to use a Plugin function that some helpdesks may not include. Rather than calling the class from the Plugin, the Plugin needed to have a new Signal to handle auditing certain deletions.

### DIFF
--- a/include/class.page.php
+++ b/include/class.page.php
@@ -154,9 +154,9 @@ class Page extends VerySimpleModel {
         if (!parent::delete())
             return false;
 
-        return Topic::objects()
-            ->filter(array('page_id'=>$this->getId()))
-            ->update(array('page_id'=>0));
+        $type = array('type' => 'deleted');
+        Signal::send('object.deleted', $this, $type);
+        return true;
     }
 
     function save($refetch=false) {
--- a/scp/categories.php
+++ b/scp/categories.php
@@ -109,7 +109,7 @@ if($_POST){
                                       foreach ($faqs as $key => $faq)
                                           $faq->delete();
                                   }
-                             $c->delete();
+                              $c->delete();
                         }
 
                         if (count($categories)==$count)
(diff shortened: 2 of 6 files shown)


aydreeihn:
**Note:** We will need to merge the PR for the plugin as well with this one.

scp/categories.php
```diff
@@ -124,15 +124,8 @@
...
-                                        && $data = AuditEntry::getDataById($id, 'C'))
-                                    $name = json_decode($data[2], true);
-                                else {
-                                    $name = __('NA');
-                                    $data = array('C', $id);
-                                }
-                                $type = array('type' => 'deleted');
-                                Signal::send('object.deleted', $data, $type);
+                                $data = array('C', true);
+                                Signal::send('object.audit', $id, $data);
```
protich: Why not send the signal `object.deleted`  plus the object in Category::delete() routine called in line 112 above?

protich: The advantage here is anyone (plugins) can catch such signal and do whatever with it. `object.audit` is just too specific to audit plugin.

scp/helptopics.php
```diff
@@ -171,15 +171,8 @@
...
-                                        && $data = AuditEntry::getDataById($id, 'H'))
-                                    $name = json_decode($data[2], true);
-                                else {
-                                    $name = __('NA');
-                                    $data = array('H', $id);
-                                }
-                                $type = array('type' => 'deleted');
-                                Signal::send('object.deleted', $data, $type);
+                                $data = array('H', true);
+                                Signal::send('object.audit', $id, $data);
```
protich: Same as above

protich (commented):
Let's talk about this PR when you get a moment.

include/class.page.php
```diff
@@ -153,10 +153,11 @@ function delete() {
 
         if (!parent::delete())
             return false;
-
-        return Topic::objects()
-            ->filter(array('page_id'=>$this->getId()))
-            ->update(array('page_id'=>0));
```
aydreeihn, line 159: This didn't appear to be relevant because we do not allow the Agent to delete a Page if a Topic is using it.

protich: Why the need for else block below?

aydreeihn: I went ahead and took the else out.

PR #5475, opened 2020-04-24
Misc. Fixes 

protich:
Misc. fixes for v1.14 - admittedly most might need to be back-ported to develop.

### DIFF
--- a/include/ajax.tickets.php
+++ b/include/ajax.tickets.php
@@ -64,6 +64,7 @@ class TicketsAjaxAPI extends AjaxController {
             return $this->lookup();
         }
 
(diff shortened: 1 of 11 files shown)


include/ajax.tickets.php
```diff
@@ -64,6 +64,7 @@ function lookup() {
             return $this->lookup();
         }
 
+        // TODO: Why not aggregate based on relationship?
```
protich, line 67: @aydreeihn - please review this commit. Basic optimization due to a warning I saw. We most definitely need to revisit merge feature at some point.

PR #5478, opened 2020-04-27
1.14 Misc. Fixes

aydreeihn:
- addCollaborator return null:
	- if the TicketOwner is being added as a Collaborator, we should just return null rather than throwing a specific errror message. The result would be the following message: 'Unable to add collaborator. Internal error occurred'. This ensures that children Tickets will be successfully closed upon merging without throwing a hidden error message as a result of attempting to move children participants over to the parent ticket.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -2538,10 +2538,13 @@ implements RestrictedAccess, Threadable, Searchable {
             $errors = array();
             foreach ($children as $child) {
                 if ($options['participants'] == 'all' && $collabs = $child->getCollaborators()) {
-                    foreach ($collabs as $collab)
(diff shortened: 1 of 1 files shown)


include/class.ticket.php
```diff
@@ -1212,7 +1212,7 @@ function getMissingRequiredField() {
     function addCollaborator($user, $vars, &$errors, $event=true) {
 
         if ($user && $user->getId() == $this->getOwnerId())
-            $errors['err'] = __('Ticket Owner cannot be a Collaborator');
+            return null;
```
protich: I think this is needed when adding collaborators via UI. You shouldn't be attempting to add  ticket owner as collaborator anyhow - perhaps that's where you need to do the check to avoid the error.

PR #5489, opened 2020-05-01
Issue: Language Verification

aydreeihn:
This plugin updates the line of code we have to check the DNS record for language packs to verify them as one of our official language packs.


protich:
Why was the update necessary? What changed to require the change?

aydreeihn:
Those last 3 parameters being passed to dns_get_record were just causing it to not work. I think they were being passed mainly so that the last parameter could be set to true, but I couldn't really see why that was necessary either.

PR #5495, opened 2020-05-05
Issue: Create Task File Upload

aydreeihn:
This commit fixes an issue where attachments were not being added if you tried to create a Task and populate a custom File Upload field. getClean() would return an empty array, so there was no information that could be used to create the attachment record.

### DIFF
--- a/include/ajax.tasks.php
+++ b/include/ajax.tasks.php
@@ -113,7 +113,7 @@ class TasksAjaxAPI extends AjaxController {
                 if ($desc
                         && $desc->isAttachmentsEnabled()
                         && ($attachments=$desc->getWidget()->getAttachments()))
-                    $vars['files'] = $attachments->getFiles();
+                            $vars['files'] = $attachments->getFiles();
                 $vars['staffId'] = $thisstaff->getId();
                 $vars['poster'] = $thisstaff;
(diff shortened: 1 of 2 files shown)


include/ajax.tasks.php
```diff
@@ -113,7 +113,7 @@ function add($tid=0, $vars=array()) {
                 if ($desc
                         && $desc->isAttachmentsEnabled()
                         && ($attachments=$desc->getWidget()->getAttachments()))
-                    $vars['files'] = $attachments->getFiles();
+                $vars['files'] = $attachments->getFiles();
```
protich: Looks like whitespace issue here but want to point out that the line should be tapped in since it's part of `if` statement.

include/class.dynamic_forms.php
```diff
@@ -1282,7 +1282,14 @@ function saveAnswers($isEditable=null, $refetch=false) {
 
             try {
                 $field->setForm($this);
-                $val = $field->to_database($field->getClean());
+                if (get_class($field) == 'FileUploadField') {
```
protich, line 1288: Perhaps a few comments here would be ideal to explain why the logic is needed here?

protich (changes requested):

protich (approved):

PR #5510, opened 2020-05-20
install: Embedded Domain Whitelist

JediKev:
This adds `youtube.com, dailymotion.com, vimeo.com, player.vimeo.com, web.microsoftstream.com` as the default values for the Embedded Domain Whitelist feature for new installs.

### DIFF
--- a/include/i18n/en_US/config.yaml
+++ b/include/i18n/en_US/config.yaml
@@ -86,3 +86,4 @@ core:
     log_graceperiod: 12
     client_registration: 'public'
     default_ticket_queue: 1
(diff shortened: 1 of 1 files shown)


dirigit:
You are sure you want to open those privacy leaks per default?

Have a look at laws related to data protection and privacy in California as well as european GDPR. There are ways to open those sites in compliant way but, this will be relative hard to accomplish with Osticket.

JediKev:
@dirigit

We’ve always allowed them since at _least_ the 1.10.x series. They were just hard coded before, now we are giving the option to configure and remove if you want. This is something we can put up for debate and discuss internally however (as to enable them on install or not).

Cheers.

PR #5538, opened 2020-06-05
Feature: Email Two Factor Authentication

aydreeihn:
This commit adds functionality to allow email 2FA to your helpdesk.
If enabled, an email is sent to Agents with a verification code that
must be entered to complete logging into the helpdesk.


aydreeihn:
Note: This requires a database patch. I built this on top of the last schema signature, but I can rebase it on top of the last unmerged database patch we have if we want to merge it as one single database patch. I just figured it would be easier to review it separately for now.

protich:
See #5568

PR #4008, opened 2017-10-17
oops: Fix Agent Identity Masking

JediKev:
This pull request addresses issues #2955 & #3524 where an Agent's name is
visible to User even when Agent Identity Masking is enabled. This adds a
check for `hide_staff_name` in the config table and if enabled the system
will show Staff (translatable) on all thread responses and events.

Issue #3524, opened 2016-11-22
[BUG] Agent Identity Masking no longer works [1.10]


ntozier:
I have a report that Admin panel -> Settings -> Agents -> Agent Identity Masking no longer works in 1.10.

I have confirmed this in my 1.10 installation.

Original report:
http://osticket.com/forum/discussion/89247/hide-agent-s-name-on-responses-not-working#latest

dogatemycomputer:
+1

ntozier:
ping

orangesunny:
I also have this annoying issue in my installation. :(

maurosanjo:
I also have this problem on my installation. The agent name is hidden on e-mail message, but on web view of ticket history the agent's name is always visible.

shimsim:
Has this been fixed in v1.10.1 or is it only going to be fixed in 1.11.0 ?

If not, is there any alternative fix other than the one posted on the forum discussion above ? I did try that fix but now the user ticket view looks kind of incomplete.

JediKev:
@shimsim I am actually working on this now. Will push the pull request soon, once I do I'll post back here.

shimsim:
@JediKev thats great news, we didnt realize the bug until customers started replying to agents using their names, which was a bit of a shock. its kind of a big bug for us, we like to keep agents personal details off the internet, anyway, using the code edit in the mean time, thanks for the update.

shimsim:
Great, thanks for the interim fix 👍

JediKev:
@shimsim I have pushed the fix for this here - #4008. Please make the changes, test, and report back to me. Cheers.

PR #5541, opened 2020-06-10
Honor "hide staff name" setting at page printing in client area

killerbees19:
Related: #4008 & #3524


JediKev:
@killerbees19 

Thank you for the pull request, however, I'm closing as this as a pull has already been made here:
- https://github.com/osTicket/osTicket/pull/5525

**Edit:**
In the future, refrain from calling stuff like `$cfg->hideStaffName()` in a `foreach`, it is not the best performance/practice. This means you would be checking the config for every entry (which can be a lot sometimes) when you only need to check it once.

Cheers.

killerbees19:
> however, I'm closing as this as a pull has already been made here:

Oops, sorry! I missed this one :-(

PR #5573, opened 2020-07-08
issue: Get Team Members For Alerts

JediKev:
This addresses an issue where when sending Alerts to Team Members, we grab all Members regardless of if they have the Alerts checkbox Enabled or not. This adds a new method called `getMembersForAlerts()` that will retrieve a list of all Team Members that have the Alerts checkbox Enabled. This updates all of the `getMembers()` calls to `getMembersForAlerts()` when sending Alerts.

### DIFF
--- a/include/class.team.php
+++ b/include/class.team.php
@@ -88,6 +88,17 @@ implements TemplateVariable {
         return $this->_members;
     }
 
+    function getMembersForAlerts() {
+        $alertmembers = array();
+        $members = $this->members->filter(array(
+            'flags__hasbit' => TeamMember::FLAG_ALERTS,
+        ));
+        foreach ($members as $m)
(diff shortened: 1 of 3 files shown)


include/class.team.php
```diff
@@ -88,6 +88,18 @@ function getMembers() {
         return $this->_members;
     }
 
+    function getMembersForAlerts() {
+        if (!isset($this->_members)) {
+            $this->_members = array();
```
protich: This will override `$this->_members` used in `getMembers()` -  whichever method gets called first will set the variable and subsequent calls may result in wrong members being returned.

protich (changes requested):

JediKev:
Updated

protich (approved):

PR #5575, opened 2020-07-08
issue: Track Fields Disabled By System Default Help Topic

JediKev:
This addresses an issue where Tickets that get assigned the System Default Help Topic (mainly emails) have Disabled Fields (disabled by Help Topic) appear in the Client Portal on Edit. This is due to not running through the Disabling of Fields for the Default Help Topic. This moves the section to track Disabled Fields to a new method called `trackDisabledFields()` in class Topic that can be reused. This also adds a check for the Default Help Topic and if set, we will run through the new method.

This also addresses an issue where having a Field that is Disabled by Help Topic and has a Default Value appears in the Ticket Header in the Client Portal. This is due to the system not checking for Disabled Fields per Help Topic, rather just checking for any entries with values. This adds a check to see if any Fields are Disabled by Help Topic, if so, we will exclude them from the Ticket Header in the Client Portal. This adds a new method called `getDisabledFieldsIds()` that retrieves an array of Form ID and disabled Field IDs (eg. `array(form_id => array(disabled_field_ids))`).

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -4140,25 +4140,7 @@ implements RestrictedAccess, Threadable, Searchable {
             // entries, disable and track the requested disabled fields.
             if ($vars['topicId']) {
                 if ($__topic=Topic::lookup($vars['topicId'])) {
-                    foreach ($__topic->getForms() as $index=>$topicForm) {
-                        $disabled = self::getDisabledFields($topicForm);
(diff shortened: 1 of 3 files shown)


JediKev:
~Need to rebase on @aydreeihn new changes.~ **DONE**

JediKev:
Updated

PR #5519, opened 2020-05-22
Fix connection string used for non-SSL connections

davidjosephhayes:
Fix I am using for https://github.com/osTicket/osTicket/issues/5020

Issue #5020, opened 2019-09-02
Mail fetch without SSL fails on PHP 7.2


edwinvandeven:
### Prerequisites

* [x] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [x] Do you have any errors in the PHP error log, or javascript console?
* [x] Did you check the [osTicket forums](http://osticket.com/forum/categories)?
* [x] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/osTicket/osTicket/wiki/Github-Issues-Guidelines)

### Description

Mail fetch without SSL fails on PHP 7.2

### Steps to Reproduce

1. When adding an imap or pop box that's not using SSL, settings returns an "SSL negotiation failed" error, settings don't save and mail doesn't get fetched.

**Expected behavior:** [What you expected to happen]

Settings are saved and mail is fetched.

**Actual behavior:** [What actually happened]

Settings refuse to save and mail doesn't get fetched.

**Possible fix:**

Settings are saved and mail is fetched when "/notls" is added to the server string in include/class.mailfetch.php. 

I wrote a quick and (incredibly) dirty fix in the forum, it might be useful in moving towards a permanent fix:
https://forum.osticket.com/d/95110-fetch-email-settings-question/3 

### Versions

osTicket 1.12.2
Ubuntu 18.04
php 7.2.19

JediKev:
@edwinvandeven 

Here is the pull to address this issue:
- https://github.com/osTicket/osTicket/pull/5603

Cheers.

PR #5603, opened 2020-08-07
issue: NOTLS For IMAP/POP Without SSL

JediKev:
This is a rewrite of #5519 and addresses issue #5020 where in some cases, using IMAP/POP without SSL fails because class MailFetcher is not disabling start-TLS. This adds a new if statement to check if the encryption method is SSL if not, we will add `/notls` to the server string to disable start-TLS.

### DIFF
--- a/include/class.mailfetch.php
+++ b/include/class.mailfetch.php
@@ -86,8 +86,10 @@ class MailFetcher {
             if ($this->authuser)
                 $this->srvstr .= sprintf('/authuser=%s', $this->authuser);
 
-            $this->srvstr.='/novalidate-cert}';
-
+            if (strcasecmp($this->getEncryption(), 'SSL'))
+                $this->srvstr.='/notls}';
+            else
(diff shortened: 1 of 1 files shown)


include/class.mailfetch.php
```diff
@@ -86,6 +86,9 @@ function __construct($email, $charset='UTF-8') {
             if ($this->authuser)
                 $this->srvstr .= sprintf('/authuser=%s', $this->authuser);
 
+            if (strcasecmp($this->getEncryption(), 'SSL'))
+                $this->srvstr.='/notls';
```
protich: I think we shouldn't add `novalidate-cert` when `notls` is set.

```php
if (strcasecmp($this->getEncryption(), 'SSL'))
    $this->srvstr.='/notls';
else
    $this->srvstr.='/novalidate-cert}';
```

JediKev:
Updated

PR #5617, opened 2020-08-19
oops: Local Inline Images

JediKev:
This addresses an issue caused by d98c2d0 where local inline images are stripped from the thread entries. This adds a check for the `cid` tag and if exists we know it’s a local image and we will allow it.


killerbees19:
@JediKev Isn't the `strpos()` check a little bit wrong? I think you should check for position `0`, otherwise it's possible to bypass the check with such URLs: `https://example.net/cid:image.png`

PR #5628, opened 2020-09-02
issue: Assignee Field

JediKev:
This addresses a list of issues with Custom Assignee Fields:
- The Thread Events do not show the changes
- Custom Columns do not show the current values
- Advanced Search with `includes` and `does not include` criteria does not find proper results
- Exports do not show the current values
- Template Variables do not show the proper values

This overhauls the AssigneeField class to make it like an independent Custom Field. This adds new methods such as `getChoice()`, `getKeys()`, `asVar()`, `getChanges`, and `whatChanged()` to work with AssigneeField specific data instead of relying on the ChoiceField class. This also updates methods like `getWidget()`, `display()`, `getValue()`, `to_php()`, and `to_database()` to ensure all methods work with a unified data format (JSON/Array).

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -3297,13 +3297,18 @@ class AssigneeField extends ChoiceField {
 
     function getWidget($widgetClass=false) {
         $widget = parent::getWidget($widgetClass);
-        if (is_object($widget->value))
-            $widget->value = $widget->value->getId();
+        $value = $widget->value;
+        if (is_object($value)) {
+            $id = $value->getId();
+            if ($value instanceof Staff)
+                $widget->value = 's'.$id;
+            elseif ($value instanceof Team)
+                $widget->value = 't'.$id;
+        }
         return $widget;
     }
 
     function getCriteria() {
-
         if (!isset($this->_criteria)) {
             $this->_criteria = array('available' => true);
             if (($c=parent::getCriteria()))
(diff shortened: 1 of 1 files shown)


JediKev:
Remove the statement that sets the currently assigned agent for built-in field.

JediKev:
Updated 👍

sevmonster:
Is this why the Assignee field doesn't appear to be changeable in list properties?

JediKev:
@ds6 

No, the AssigneeField is a core built-in field and cannot be changed/updated (like the Department/Help Topic Fields). You can however add a custom AssigneeField that you can configure like any Custom Field. Any Custom Assignee Fields however will not actually assign/unassign Agents/Teams, only the built-in field can do this.

Cheers.

sevmonster:
@JediKev Sorry, I am still new to osTicket and don't really follow. Let me explain what I am trying to do.
I want to set the Assignee property on list items so that I can use filters to assign tickets to specific teams based on the Item selected. Is this considered a "custom" Assignee field?
image
I noticed in the database that it does not save any changes to these fields from any source.
image
ID 72 and 73 map to the Department and Assignee fields above.

If this patch doesn't address this problem should I open an issue?

protich:
@ds6 - Interesting use case - yes that would be considered custom field. If you have a test environment then feel free to merge this PR and let us know if addresses the issue you're having. Thanks.

sevmonster:
@protich Nope, does not save whatever I set to it, and even if I set it on the database itself it does not reflect that change in the UI or when attempting to run it through a filter.

protich:
@ds6 - we will look into it. Thanks for the feedback.

sevmonster:
@protich I should add I rebased to develop head. I just tested on current PR commit and it still does not work, again even if I make changes in the database first and check the UI.
It would be great if this would work—but until then I will just use a Short Answer and map it like that 🙂

I can open an issue for this with some more testing if you want.

protich:
@ds6  - Please open an issue - we can confirm that it's a problem. Merging this PR shortly.

PR #5640, opened 2020-10-06
Return nicer default mail address

gerundt:
I noticed that *Post Reply* use a ugly default mail address. It use `Support<foo@bar.com>` (without whitespace) instead `Support <foo@bar.com>`. This PR should fix the problem.


JediKev:
Thank you for the pull request. While the change you made is great we would like to add to it. We will reference the changes here.

rmilecki:
Issue was fixed by the https://github.com/osTicket/osTicket/pull/6007

I believe this pull request can be closed now.

Issue #3422, opened 2016-10-26
Custom ROOT_PATH values are not applied


S1owbro:
At present, and under anything other than exceptional operating conditions, it is basically impossible for a custom ROOT_PATH value defined in "ost-config.php" to be applied in the execution environment.

Before the "ost-config.php" file is included by Bootstrap::loadConfig(), this global code executes:

```php
if (!defined('ROOT_PATH') && ($rp = osTicket::get_root_path(dirname(__file__))))
    define('ROOT_PATH', rtrim($rp, '/').'/');
```
"!defined('ROOT_PATH')" will always return true because "ost-config.php" hasn't yet been included, the if test will always pass except in the unlikely event that get_root_path fails, and ROOT_PATH will be defined here. Thus, when "ost-config.php" is included later on in the startup process the custom ROOT_PATH isn't set because osTicket has already defined ROOT_PATH.

legas1:
Hey, it appears this "bug" is still here. It would be awesome to be able to define ROOT_PATH inside ost-config.php at all times. 

My webhosting provider uses .htaccess to rewrite base url paths into subdomains (so test.something.eu is not like this - test.something.eu/subdom/test/...).

sknopp:
i added a PR to address this issue. Hopefully it'll be applied, as there is no chance to use osTicket within more complex setups at the moment. (in my case deployment through docker image behind an http proxy)

fjf2002:
I also ran in that issue.
Any advances here?

jfbvm:
If anyone is facing this problem, yet.

I need to add define('ROOT_PATH', '/'); to files on first lines:
client.inc.php
main.inc.php
include/ost-config.php => I added here also just in case...😅😅

In my case, "/" solved the problem because of my server configuration and proxy, but you can change it to the path you want, and it will work.

PR #5649, opened 2020-10-11
fixing ROOT_PATH config variable not being applied early enough

sknopp:
As stated in #3422 `ROOT_PATH` is currently being defined by `bootstrap.php` before the configuration file was read.
The only reason where this is needed is for correctly redirecting to the `/setup` pages during first usage.

Unfortunately because `define()` can only be used once, this leads to the `ROOT_PATH` definition in `ost-config.php` being useless all the time. With this PR another variable name is used for setup and later we fall back to it, if configuration did not define anything.

Therefore more complex webserver configurations become possible now.


trukonsult:
I ran into this as well. However, the installation process expects ROOT_PATH (or TMP_ROOT_PATH with the patch) to be set earlier. So by moving define into loadConfig() you are setting it too late. So, it's probably better to leave the define where it is and just rename ROOT_PATH to TMP_ROOT_PATH.

I'm pondering if SETUP_ROOT_PATH or INSTALL_ROOT_PATH wouldn't be a better name, but that's bikeshedding.

You need to change ROOT_PATH to TMP_ROOT_PATH in setup/inc/header.inc.php as well.

With those changes this patch works as expected for me.

fjf2002:
I also ran in that issue.
Any advances here in applying this PR?

jfbvm:
If anyone is facing this problem, yet.

I need to add define('ROOT_PATH', '/'); to files on first lines:
client.inc.php
main.inc.php
include/ost-config.php => I added here also just in case...😅😅

In my case, "/" solved the problem because of my server configuration and proxy, but you can change it to the path you want, and it will work.

JediKev:
@jfbvm 

Please stop cross posting. Post one time on one thread.

Cheers.

PR #5652, opened 2020-10-13
Fix department translations in dashboard

gerundt:
We had the problem, that in the dashboard osTicket only show one name for all departments.

I found the problem: With `dept__id` was use a wrong array key.

### DIFF
--- a/include/class.report.php
+++ b/include/class.report.php
@@ -77,7 +77,6 @@ class OverviewReport {
 
     function getPlotData() {
         list($start, $stop) = $this->getDateRange();
-        $states = array("created", "closed", "reopened", "assigned", "overdue", "transferred");
         $event_ids = Event::getIds();
 
         # Fetch all types of events over the timeframe
--- a/include/class.report.php
+++ b/include/class.report.php
@@ -212,15 +212,15 @@ class OverviewReport {
         case 'dept':
             $headers = array(__('Department'));
             $header = function($row) { return Dept::getLocalNameById($row['dept_id'], $row['dept__name']); };
-            $pk = 'dept__id';
+            $pk = 'dept_id';
             $stats = $stats
                 ->filter(array('dept_id__in' => $thisstaff->getDepts()))
-                ->values('dept__id', 'dept__name', 'dept__flags')
-                ->distinct('dept__id');
+                ->values('dept_id', 'dept__name', 'dept__flags')
+                ->distinct('dept_id');
             $times = $times
                 ->filter(array('dept_id__in' => $thisstaff->getDepts()))
-                ->values('dept__id')
-                ->distinct('dept__id');
+                ->values('dept_id')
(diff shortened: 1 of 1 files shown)


JediKev:
@gerundt 

I cannot replicate this issue using latest `develop`, Apache, PHP 7.4, and MySQL 8.0. Please post exact steps to replicate this issue.

Upon looking at the code, we actually need `dept__id`. We are pulling the data from `ost_thread_event` and that has a relation called `dept` that will grab related Departments. You can see the relation here:
- https://github.com/osTicket/osTicket/blob/develop/include/class.thread.php#L1990-L1995

So, in order for us to get the related Departments' true `id` we need `dept__id` (hence why we use `dept__name` and `dept__flags` - so we can get the true Name and Flags).

**Note:**
Also, you removed the `$states` array completely, which I have no idea why.

**TLDR;**
I don't think your changes are correct.

Cheers.

gerundt:
@JediKev

We had the problem at our productiv system. Maybe it is related to the translation system? We installed osTicket in English and use German as seconds language. The test user had German as his first langauge.

In line 214 the `Dept::getLocalNameById($row['dept_id'], $row['dept__name']);` use `dept_id` but the rest of the code use `dept__id`:

https://github.com/osTicket/osTicket/blob/develop/include/class.report.php#L214

I searched for `dept__id` in the source code and found no other using. So I switched to `dept_id` instead.

My NetBeans sayed that `$states` was unused and I found no using too. So I removed it.

gerundt:
@JediKev Back at the office I look at the code again.

To get the translation from the department, osTicket use at [line 292](https://github.com/osTicket/osTicket/blob/develop/include/class.report.php#L214) the function in the `$header` variable:

```
$rows[] = array($header($R) . $status, $R['Opened'], $R['Assigned'],
```

It get set at [line 214](https://github.com/osTicket/osTicket/blob/develop/include/class.report.php#L292):

```
$header = function($row) { return Dept::getLocalNameById($row['dept_id'], $row['dept__name']); };
```

Here an output from a `$R` variable:

```
array(9) {
  ["dept__id"]=>
  int(1)
  ["dept__name"]=>
  string(7) "Support"
  ["dept__flags"]=>
  int(4)
  ["Opened"]=>
  float(1)
  ["Assigned"]=>
  float(0)
  ["Overdue"]=>
  float(1)
  ["Closed"]=>
  float(0)
  ["Reopened"]=>
  float(0)
  ["Deleted"]=>
  float(0)
}
```

You see the array has a `dept__id` key but the function in `$header` will look for `dept_id`. So only the english name will shown.

I searched again for  `dept__id` (with two `_`) in the source code and found no other using. In this file you use also `topic_id` (with one `_`). So i thougth a `dept_id` would be more consquent. ;)

And the `$states` variable in the function `getPlotData()` looks still unused.

PR #5653, opened 2020-10-13
Fix department translations in tasks overview

gerundt:
We had the problem, that in the task overview osTicket only show one name for all departments.

Problem is the missing `dept_id` array value.

### DIFF
--- a/include/staff/tasks.inc.php
+++ b/include/staff/tasks.inc.php
@@ -161,7 +161,7 @@ $tasks->annotate(array(
 ));
 
 $tasks->values('id', 'number', 'created', 'staff_id', 'team_id',
-        'staff__firstname', 'staff__lastname', 'team__name',
+        'staff__firstname', 'staff__lastname', 'team__name', 'dept_id',
         'dept__name', 'cdata__title', 'flags', 'ticket__number', 'ticket__ticket_id');
 // Apply requested quick filter
 


JediKev:
@gerundt 

I cannot replicate this issue using latest `develop`, Apache, PHP 7.4, and MySQL 8.0. Please post exact steps to replicate this issue.

Upon looking at the code, we grab all Departments' Names from `dept__name` as Tasks have a relation called `dept` that grabs all related Departments. You can see the relation here:
- https://github.com/osTicket/osTicket/blob/develop/include/class.task.php#L25-L27

So it should be pulling all related Departments' names correctly.

Then, as far as sorting goes, we set the Department sort column to `dept__name` via:
- https://github.com/osTicket/osTicket/blob/develop/include/staff/tasks.inc.php#L55

So once you sort by the Department Column it will sort by `dept__name`.

**TLDR;**
There is no need for the changes in this pull request.

Cheers.

gerundt:
@JediKev

We had the problem at our productiv system. Maybe it is related to the translation system? We installed osTicket in English and use German as seconds language. The test user had German as his first langauge.

I think the need the `dept_id` to get the correct translation? Maybe he return always the same translation for all departments?

gerundt:
Then `$dept = Dept::getLocalById($T['dept_id'], 'name', $T['dept__name']);` makes the trouble?

https://github.com/osTicket/osTicket/blob/develop/include/staff/tasks.inc.php#L379

gerundt:
Like I said is `$dept = Dept::getLocalById($T['dept_id'], 'name', $T['dept__name']);` at [line 379](https://github.com/osTicket/osTicket/blob/develop/include/staff/tasks.inc.php#L379) the problem.

Here is a `var_dump()` from the `$T` variable:

```
array(17) {
  ["id"]=>
  int(2)
  ["number"]=>
  string(1) "2"
  ["created"]=>
  string(19) "2020-10-14 08:38:57"
  ["staff_id"]=>
  int(0)
  ["team_id"]=>
  int(0)
  ["staff__firstname"]=>
  NULL
  ["staff__lastname"]=>
  NULL
  ["team__name"]=>
  NULL
  ["dept__name"]=>
  string(7) "Support"
  ["cdata__title"]=>
  string(15) "Support Aufgabe"
  ["flags"]=>
  int(1)
  ["ticket__number"]=>
  NULL
  ["ticket__ticket_id"]=>
  NULL
  ["collab_count"]=>
  float(0)
  ["attachment_count"]=>
  float(0)
  ["thread_count"]=>
  float(1)
  ["isopen"]=>
  int(1)
}
```

There is no `dept_id` array key which use the `Dept::getLocalById()` function.

PR #5655, opened 2020-10-13
Add docker support

hiagodotme:
Docker support added with `docker-compose.yml` and `Dockerfile`.

To deploy osTicket is now simple, just get the source code and with the docker installed type the command line `docker-compose up -d` and access http://localhost/.

The docker-compose already creates a container with MySQL, where it is possible to parameterize the user and password.

Standard data for connection inside the container:

Database host: `osticket-db`
Database user: `osticket`
User password: `osticket`
Database name: `osticket`

To change this information, just edit the docker-compose.yml


Dockerfile
```diff
@@ -0,0 +1,13 @@
+FROM php:7.3-apache
+
+RUN apt-get update -y
```
oxr463:
```
RUN apt-get update -y &&\
         apt-get install -y $dependencies
```

Using the `&& \` syntax will decrease image layers.

Dockerfile
```diff
@@ -0,0 +1,13 @@
...
+RUN apt-get install -y sendmail libpng-dev zlib1g-dev libc-client-dev libkrb5-dev libicu-dev libzip-dev zip
+
+RUN docker-php-ext-install mysqli
```
oxr463: Same for all of these.

oxr463 (changes requested):

Dockerfile
hiagodotme: Thank you for the tips. I'm already sending a new pull request.

Dockerfile
hiagodotme: Thank you for the tips. I'm already sending a new pull request.

hiagodotme:
@oxr463 I sent the changes. =)

oxr463 (approved):

codestation:
IMO this doesn't actually adds docker support as the resulting image doesn't have osticket on it, just the runtime.

oxr463:
> IMO this doesn't actually adds docker support as the resulting image doesn't have osticket on it, just the runtime.

@hiagodotme we should use `COPY` in the Dockerfile to ensure that the source code is contained in the final image.

codestation:
It would require more work since one would expect that the image would be ready to use once the proper configuration is passed to the container, so no manual setup should be necessary to use it.

Right now i am using https://github.com/CampbellSoftwareSolutions/docker-osticket, but i don't like it too much because it fetch the code from a git branch so one doesn't know exactly the version that ends on the image.

hiagodotme:
@codestation I understood your point of view. Do you just want to give a docker-compose up and leave using it correct?

mitchplze:
Hello, just curious if osTicket has been Dockerized yet? Not sure if I'm missing it somewhere.

arcreigh:
> It would require more work since one would expect that the image would be ready to use once the proper configuration is passed to the container, so no manual setup should be necessary to use it.
> 
> Right now i am using https://github.com/CampbellSoftwareSolutions/docker-osticket, but i don't like it too much because it fetch the code from a git branch so one doesn't know exactly the version that ends on the image.

Looking at his dockerfile it pulls from this git repo and uses an environment variable to determine the version that you want to pull, it may be a good idea to reach out to this guy to do a pull request to get this imported into the main branch.
https://github.com/CampbellSoftwareSolutions/docker-osticket/blob/master/Dockerfile

```
RUN set -x \
    && apt-get update \
    && apt-get install -y git-core \
    && git clone -b v${OSTICKET_VERSION} --depth 1 https://github.com/osTicket/osTicket.git \
    && cd osTicket \
    && php manage.php deploy -sv /data/upload \
    # www-data is uid:gid 82:82 in php:7.0-fpm-alpine
    && chown -R 82:82 /data/upload \
    # Hide setup
    && mv /data/upload/setup /data/upload/setup_hidden \
    && chown -R root:root /data/upload/setup_hidden \
    && chmod -R go= /data/upload/setup_hidden
```

Here are lines 4 - 15 in his dockerfile
The only thing of note is an old PHP version. That should definitely be bumped to 7.4 for security reasons.

bobweston:
> It would require more work since one would expect that the image would be ready to use once the proper configuration is passed to the container, so no manual setup should be necessary to use it.
> 
> Right now i am using https://github.com/CampbellSoftwareSolutions/docker-osticket, but i don't like it too much because it fetch the code from a git branch so one doesn't know exactly the version that ends on the image.

I'm not exactly clear on what you mean. If you look in the Dockerfile on that repo, it specifies exactly what version of osTicket gets installed

https://github.com/CampbellSoftwareSolutions/docker-osticket/blob/master/Dockerfile#L3

```
FROM php:7.3-cli AS deployer
ENV OSTICKET_VERSION=1.14.3
```

My issue with it is that repo, as it is, it doesn't support php 8 and v1.16.3. The dev hasn't updated it in quite awhile and isn't responding to tickets.

I've forked that project and made the changes to support those versions. But the original repo does indeed assign the version in the ENV secton at the top

The branch point you mention is actually pointing to the version of osTicket you are installing. Go to the osticket github page and look at branches. You'll see that they're actually named after the osTicket version)

https://github.com/osTicket/osTicket/branches

Upgrading the app in docker, assuming the php version is correct, is simply a matter of bumping the value defined in that ENV constant, building the new image for that new version and restarting the container using the new image. 

In my view,, the devs here ought to consider using the campbellsoftwaresolutions project as the basis for any future effort to containerize this project. I've looked at many attempts to getting this application working in docker. The campbellsoftware solution one is closest to being ready (though as noted, it doesn't currently support php 8 in that version. My fork does. But I've only had to make some tweaks to get it working with php 8).

PR #5670, opened 2020-10-23
Fix incorrrect compare locked staff at Ticket Preview

xkvnn:
Can't show other staff locking the ticket at Ticket Preview popup,


protich:
Thanks for the PR.  Crazy that we've had the bug for 7 years.

PR #5671, opened 2020-10-23
Issue: Default Delimiter

aydreeihn:
This commit fixes an issue where we needed to make sure to put a default delimiter for Ticekt exports. If no delimiter is specified, the default delimiter will be ','

### DIFF
--- a/include/class.export.php
+++ b/include/class.export.php
@@ -603,7 +603,8 @@ class CsvExporter extends Exporter {
     }
 
     function getDelimiter() {
-        if (isset($this->options['delimiter']))
+        if (isset($this->options['delimiter'])
+                && $this->options['delimiter'])
             return $this->options['delimiter'];
         return Internationalization::getCSVDelimiter();
     }
--- a/include/staff/templates/queue-export.tmpl.php
+++ b/include/staff/templates/queue-export.tmpl.php
@@ -13,7 +13,7 @@ else
 if (isset($cache['delimiter']))
     $delimiter = $cache['delimiter'];
 else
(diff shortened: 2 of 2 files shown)


protich:
While above solves the symptom of the issue - I think the correct fix should be in class export (using only `isset` was the issue).  Also added default delimiter based on user's locale. See the diffs below. 

```diff
diff --git a/include/class.export.php b/include/class.export.php
index 13243b2f..642885c0 100644
--- a/include/class.export.php
+++ b/include/class.export.php
@@ -603,7 +603,8 @@ class CsvExporter extends Exporter {
     }
 
     function getDelimiter() {
-        if (isset($this->options['delimiter']))
+        if (isset($this->options['delimiter'])
+                && $this->options['delimiter'])
             return $this->options['delimiter'];
         return Internationalization::getCSVDelimiter();
     }
diff --git a/include/staff/templates/queue-export.tmpl.php b/include/staff/templates/queue-export.tmpl.php
index 4b6306a5..23bd486c 100644
--- a/include/staff/templates/queue-export.tmpl.php
+++ b/include/staff/templates/queue-export.tmpl.php
@@ -13,7 +13,7 @@ else
 if (isset($cache['delimiter']))
     $delimiter = $cache['delimiter'];
 else
-    $delimiter = ''; //TODO: get user's preference (browswer settings)
+    $delimiter = Internationalization::getCSVDelimiter();
 
 $fields = $queue->getExportFields(false) ?: array();
 if (isset($cache['fields']) && $fields)
```

Please note that User can input anything for delimiter - this is a feature - we don't validate it on purpose! Most common delimiters are the comma, tab, and colon but user is free to use any character they wish!

PR #5679, opened 2020-11-04
Ticket From Email

aydreeihn:
This commit ensures that Agents only have the option to choose the From Email on a Ticket based on the Department Emails they have access to. By default, the Ticket's Department Email will be the From email selected.

### DIFF
--- a/include/staff/ticket-view.inc.php
+++ b/include/staff/ticket-view.inc.php
@@ -836,9 +836,10 @@ if ($errors['err'] && isset($_POST['a'])) {
                                  $e->getId(),
                                  Format::htmlchars($e->getAddress()));
                      }
+                     $staffDepts = $thisstaff->getDepts();
                      // Optional SMTP addreses user can send email via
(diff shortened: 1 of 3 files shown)


include/staff/ticket-view.inc.php
```diff
@@ -836,9 +836,10 @@ class="icon-building icon-fixed-width"></i> <?php
                                  $e->getId(),
                                  Format::htmlchars($e->getAddress()));
                      }
+                     $staffDepts = array_keys($thisstaff->getDepartmentNames(true));
```
protich: Why not use `$thisstaff->getDepts()` ? It's cached and already in the format you need.

protich (changes requested):

include/staff/ticket-view.inc.php
aydreeihn: Updated! I had done it like that somewhere else too. I went ahead and corrected that one as well. Let me know if I should correct it separately instead though.

protich (approved):

PR #5681, opened 2020-11-05
Issue: Prevent Deleting All Topics

aydreeihn:
This commit makes sure that we always have at least 1 Public and Active Help Topic

Mass Process:
- Do not allow deletion of all Topics
- Do not allow archiving/disabling all Topics
- Allow changing Topics from Disabled to Archived and vice versa

Note: If you try to archive, disable, or delete all Topics OR all Active Topics at once, it will throw the error that at least 1 topic must be active.

Topic Update:
- Do not allow changing Topic to Private if it's the only one that's Public
- Do not allow archiving/disabling if it's the only active Topic
- Allow changing Topics from Disabled to Archived and vice versa

It also gets rid of code we had in place in the Client Portal that would put a 'General Inquiry' Topic in the list if all Topics had been deleted. It didn't have an id to get a Topic form, so it still wouldn't allow for Ticket creation.

### DIFF
--- a/scp/helptopics.php
+++ b/scp/helptopics.php
@@ -63,6 +63,10 @@ if($_POST){
             if (!$errors) {
                 $count=$_POST['ids']?count($_POST['ids']):0;
 
+                $activeTopics = Topic::getHelpTopics(false, false);
+                $allTopics = count(Topic::getAllHelpTopics());
+                $diff = array_intersect($_POST['ids'], array_keys($activeTopics));
+
                 switch(strtolower($_POST['a'])) {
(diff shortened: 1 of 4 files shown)


scp/helptopics.php
```diff
@@ -94,12 +94,17 @@
                         }
                         break;
                     case 'disable':
+                        $num=1;
```
aydreeihn: I feel like there's probably a better way to handle this, I just get weird with counters sometimes. I'm open to suggestions for improvement on the mass process bit.

PR #2901, opened 2016-02-02
Feature :: Canned Tasks

greezybacon:
This implements one of the features outlined in #2826. There it was labelled "Task Templates"
### How to Test
- Checkout the branch from my repo `feature/canned-tasks`, or [visit the branch](https://github.com/greezybacon/osTicket-1.8/tree/feature/canned-tasks) and download from github
- Install or upgrade the system normally
- "Hold on to your butts"

_Please feel free to test and report how you view the software implementing your workflow related challenges._

#### Cancellations

Tasks can be cancelled. This is a special case of PENDING, where the task is closed before it is started. Therefore the PENDING flag is never cleared. This is useful to mark tasks as unnecessary when they are included in a set. A screenshot is included below.

#### Due dates can be based on a related task

Task template due dates can be configured based on the starting date of a related task. For instance, if completing two tasks must happen within a certain period, the second task can use the start time of the first task a reference. A screenshot is included below.

#### Task Template Groups can be windowed

Task template groups can be associated with a department and/or a help topic. This will help workflows for various departments from be accidentally used or cluttering up the UI for unrelated tickets and departments.

### Glossary

I apologize of the verbiage in the pull request differ from the definitions here. That's most of my intention in posting the definitions.

**Task Template** is a skeleton of a single item in a workflow. This might also be called a _Canned Task_
**Task Template Group** is a workflow which includes one or more _Task Templates_ which describe the work items and the interdependency of those items.
**Task Set** a group of tasks which was added to a _Ticket_
**Dependency** is a task which must be completed before a certain task can be started. If task A is a _Dependency_ of task B, then task A must be completed before task B can start. Also, task B depends on task A
**Dependent** is a task on which a task depends. That is, if task B depends on task A, then task B is a _Dependent_ of task A 

### Outstanding
- [x] Installation is missing initial ticket
- [x] New ticket `Issue Summary` does not stick
- [x] Implement enabling and disabling of task templates
- [x] Implement enabling and disabling of task template groups
- [x] Add manual _start_. If a task can be started, then allow an agent to manually start the task
- [x] Polish and test the upgrader from `develop-next`
- [ ] New tasks in a set are not sorted to add to the bottom--especially if there have been deletions
- [x] Delete active task in template group seems not to work
- [x] Add support to close a task and cancel all children / descendents (ie. nothing to do)
- [ ] After adding a template to a workflow, have option to "add another"
- [ ] In task template listing, add "select all", '... none', '... toggle'
- [ ] Add "Canned Responses" to task UI
[...]

PR #5683, opened 2020-11-09
Feature/canned tasks

korobko099:


JediKev:
@korobko099 

I see you just copied someone else’s commits (#2901) and made a pull to `master`. Two things:
1.) Please don’t copy/duplicate an existing pull; we will only work with/merge the original (unless we say otherwise).
2.) We do not use `master` as a working branch. We mainly use `develop`/`develop-next` and when we cut the release we create a tag with the specific version number.

I will close this pull in favor of the original (#2901) and due to the fact that we don’t use `master` like that.

Cheers.

PR #5685, opened 2020-11-09
issue: Better URL Parsing For External Inline Images

JediKev:
This improves the parsing for external inline image URLs to better detect path extensions. This uses `parse_url()` instead of `preg_match()`/`explode()` to get the true URL path extension.

### DIFF
--- a/include/class.format.php
+++ b/include/class.format.php
@@ -424,8 +424,9 @@ class Format {
                 $m = array();
                 // Split the src URL and get the extension
                 preg_match('/src="([^"]+)"/', $match[2], $m);
-                $url = explode('.', explode('?', $m[1])[0]);
-                $ext = preg_split('/[^A-Za-z]/', end($url))[0];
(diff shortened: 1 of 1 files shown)


include/class.format.php
```diff
@@ -424,8 +424,9 @@ function($match) use ($exclude, $allowed) {
                 $m = array();
                 // Split the src URL and get the extension
                 preg_match('/src="([^"]+)"/', $match[2], $m);
-                $url = explode('.', explode('?', $m[1])[0]);
-                $ext = preg_split('/[^A-Za-z]/', end($url))[0];
+                $parts = parse_url($m[1]);
+                $path = explode('.', $parts['path']);
```
protich: You might want to account for a case where url is invalid - `parse_url` would return `false` in that case. I would also do the following.
```php
$patch = parse_url($url, PHP_URL_PATH);
```

PR #5693, opened 2020-11-12
PluginSupport: Add more Signals with documentation

t-oster:
This is a follow up of #4676

I added a few signals I need for my own plugins and added also some documentation in the signal class as comments for the constant signal-strings.

If you consider merging, please let me know, I am willing to add documentation for all other existing signal calls in order to make plugin development easier for future developers.

### DIFF
--- a/include/class.signal.php
+++ b/include/class.signal.php
@@ -25,6 +25,56 @@
  * the codebase there exists a Signal::send() for the same named signal.
  */
 class Signal {
+    
+    /**
+     * Is called during creation of the template HTML header of the staff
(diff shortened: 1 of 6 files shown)


t-oster:
Well I still suggest you document the signals with constants like in this PR.
However I managed to work around all the missing singals with a plugin like https://github.com/clonemeagain/attachment_preview which basically captures the whole HTML before sending to the browser and then provides an interface for plugins to mess with it. This way I can hijack any HTML element via plugin.

Thus I close this PR.
However If you feel like adding more Signal documentation or even include a DOM-Manipulation Signal, feel free to contact me.

PR #5694, opened 2020-11-12
issue: Draft Save

JediKev:
This addresses a small issue introduced with #5694 where Drafts do not save when creating a New Ticket via Staff/Client side. This is due to a field name mismatch in `include/ajax.draft.php`. We are expecting `message` but we get the hashed name instead. This addresses another issue where Drafts do not save when creating a New Task. We are expecting `description` but we get the hashed name instead. This adds the hashed names to the array of possible fields which allows the system to find the Draft bodies and save the Drafts successfully.

This also addresses an issue where the Task Description is missing when creating new Tasks. This is due to a field name mismatch when #5694 was applied. This gets the clean value to ensure the content is saved.

### DIFF
--- a/include/ajax.draft.php
+++ b/include/ajax.draft.php
@@ -388,8 +388,15 @@ class DraftAjaxAPI extends AjaxController {
                 return $focus;
             }
         }
+
+        // Get ThreadEntry field name.
+        $tform = TicketForm::objects()->one()->getForm();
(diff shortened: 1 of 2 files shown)


sevmonster:
@JediKev Missed this, assume related to https://github.com/osTicket/osTicket/issues/5698? Just merged to test, tickets are fine but tasks still cause errors.

Issue #5697, opened 2020-11-12
Setup fails for 1.14.3 and latest develop


sevmonster:
### Prerequisites

* [x] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [x] Do you have any errors in the PHP error log, or javascript console?
* [x] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [x] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

### Description

When going to create a new installation and database from the latest develop branch, I noticed that I receive this error:
```
PHP Fatal error:  Uncaught Error: Call to a member function getTopicSortMode() on null in C:\inetpub\test-vanilla\include\class.topic.php:390
Stack trace:
#0 C:\inetpub\test-vanilla\include\class.topic.php(447): Topic::getHelpTopics()
#1 C:\inetpub\test-vanilla\include\class.topic.php(304): Topic->update()
#2 C:\inetpub\test-vanilla\include\class.i18n.php(82): Topic::__create()
#3 C:\inetpub\test-vanilla\setup\inc\class.installer.php(174): Internationalization->loadDefaultData()
#4 C:\inetpub\test-vanilla\setup\install.php(52): Installer->install()
#5 {main}
  thrown in C:\inetpub\test-vanilla\include\class.topic.php on line 390
```
The database is created, but all tables are empty except for:
![image](https://user-images.githubusercontent.com/16788697/98997156-ad6d2500-24f9-11eb-8e26-f03d290cb2cb.png)
...and ost_config just has the blank values from `install-mysql.sql`.

I also tested this with 1.14.3 downloaded from the site, and ran into the same problem.

### Steps to Reproduce

1. Unzip 1.14.3 or run `php manage.php deploy -s` to a new folder
2. Configure host of choice and start site
3. Go through the setup procedure like normal

**Expected behavior:** Setup to complete and new site to come online

**Actual behavior:** Database is mostly empty and `ost-config.php` does not have `OSTINSTALLED` set to `TRUE` (nor are any other values updated using provided values during setup, such as database info).

### Versions

osTicket v1.14.3 _and_ 2a88796ec674f65e6eaa0037d1e01b9d885221a0.
![image](https://user-images.githubusercontent.com/16788697/98997617-616eb000-24fa-11eb-9690-34e971e17a0b.png)

sevmonster:
#5699 fixed this.

PR #5699, opened 2020-11-12
Issue: Topics on Install

aydreeihn:
This commit addresses Issue #5697. We need to make sure that we are not trying to ensure there are Help Topics before installation has completed.


sevmonster:
@aydreeihn Thanks!

PR #5705, opened 2020-11-16
Issue: Missing Events

aydreeihn:
This commit adds to the latest database task. We need to see if the event table has the login, logout, message, and note events, and if a database does not, we need to add them to be used with the audit plugin.

It also fixes a small issue where the upgrader threw an error saying the run function needed to be compatible with MigrationTask::run($max_time).

### DIFF
--- a/include/upgrader/streams/core/e7038ce9-ddbe2e76.task.php
+++ b/include/upgrader/streams/core/e7038ce9-ddbe2e76.task.php
@@ -27,7 +27,7 @@ class AddIndexMigration extends MigrationTask {
         'user:name'
     );
 
-    function run() {
+    function run($max_time) {
         global $ost;
 
         foreach ($this->indexes as $index) {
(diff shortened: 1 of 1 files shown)


include/upgrader/streams/core/e7038ce9-ddbe2e76.task.php
```diff
@@ -87,6 +87,30 @@ function run() {
...
+         // See if there are missing events that should be added to the database
+         $event_type = array('login', 'logout', 'message', 'note');
+         foreach($event_type as $eType) {
+             $sql = sprintf("SELECT * FROM `%s` WHERE name = '%s'",
+                     TABLE_PREFIX.'event', $eType);
+
+             $res=db_query($sql);
+
+             if($count = db_num_rows($res) && $count > 0)
+                return $this->error('Event already exists');
```
protich: Are you sure you want to error out (return) here? I would think `continue` will be sufficient.

PR #5710, opened 2020-11-19
Feature fix signal object created

t-oster:
If a new message is appended to a ticket (e.g. via E-Mail), normally the Signal object.created is sent. But in case no alerts or autorespond is neccessary, the method Ticket::postMessage is left early without sending the signal. Thus my plugin does not have the ability to still re-open Tickets even on auto-reply (like mailer daemon errors).

This PR makes sure the signal is sent also in this case.


t-oster:
nevermind.... the correct signal is object.edited... supid me

PR #5711, opened 2020-11-19
Restrict user searches based on perms

rodwgs:
This edit makes it so that if the agent typing into a user-search field does not have permission to search the user directory, they will not be presented a list of all users (and emails) after only a few characters. Instead, they must type a user's entire email before the user shows up as a match. This prevents users without the proper permissions from harvesting the entire user list just by searching on a few characters. We currently use this edit is our local implementation of osTicket at WGS.


rodwgs:
This is more of a suggested feature than a bug fix or anything else.

PR #5724, opened 2020-12-02
fix non-ssl SMTP servers

### LINKED ISSUE #5590: smtp server without SSL
Hello
i have an installation
64 bit Linux Mint 18.1 Serena
PHP 7.3.4
osTicket version v1.14.1 (f1e9e88)
ServerDB 10.1.38-MariaDB
mail server does not use any SSL authentication for both the SMTP and POP3 parts
I configure the SMTP server and set the correct parameters but it tells me that:
authentication failure [SMTP: STARTTLS failed (code: 220, response: Ready to start TLS)]
using an external server if using an unencrypted connection gives me the same error if instead using an encrypted connection SSL works
it appears that it no longer allows you to use an unencrypted SSL SMTP server
it seems that other users have the same problems as I could fix it
Admin panel -> Dashboard -> Information which also additionally gives you information about your server.

Also, please include the OS and what version of the OS you're running. As well as your browser and browser version.

ericLemanissier:
fixes #5590

### DIFF
--- a/include/pear/Net/SMTP.php
+++ b/include/pear/Net/SMTP.php
@@ -170,7 +170,7 @@ class Net_SMTP
         // Turn off peer name verification by default
         if (!$socket_options)
             $socket_options = array(
-                    'ssl' => array('verify_peer_name' => false)
+                    'ssl' => array('verify_peer' => false, 'verify_peer_name' => false)
                     );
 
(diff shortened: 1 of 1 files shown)


ericLemanissier:
friendly ping @JediKev

PR #5754, opened 2020-12-18
feature: Client Portal Open/Answered Queue

JediKev:
This feature adds a new Queue to the Client Portal called `Answered`. The idea is the exact opposite of the Answered Queue for Agents, meaning, "Answered" on the Client Portal are Tickets answered by the User (not Agent). This helps further separate Tickets that need immediate attention or a response from a User vs Tickets that have already been responded to by a User. This Queue acts the same as Open and Closed in regards to if there are no Answered tickets the Queue is hidden.

### DIFF
--- a/include/class.client.php
+++ b/include/class.client.php
@@ -259,20 +259,24 @@ class EndUser extends BaseAuthenticatedUser {
         return $this->_stats;
     }
 
-    function getNumTickets($forMyOrg=false, $state=false) {
+    function getNumTickets($forMyOrg=false, $state=false, $answered=null) {
         $stats = $this->getTicketStats();
         $count = 0;
         $section = $forMyOrg ? 'myorg' : 'mine';
         foreach ($stats[$section] as $row) {
+            $isanswered = ($row['isanswered'] == 1);
             if ($state && $row['status__state'] != $state)
                 continue;
+            if ($state && $state == 'open' && isset($answered))
+                if ($answered == $isanswered)
+                    continue;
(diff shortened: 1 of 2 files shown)


mlipok:
Thanks:

I have some additional concept:

As an agent I would like to have 
- Answered/NotAnswered by Agent
- Answered/NotAnswered by Client

Also as a client I would like to have
- Answered/NotAnswered by Agent
- Answered/NotAnswered by Client

Then I would be able to have new queues:

As an agent:
- Answered by Client but NotAnswered by Agent
oops agent must to answer to the clinet answers

As a client:
- Answered by Agent but NotAnswered by Client
oops client must to answer to the agent answers

Do this sounds resonabled ?

JediKev:
@mlipok 

That concept already exists and is the whole point of Answered/Unanswered. With Advanced Search/Custom Queues you have the criteria of `Answered` which can be Checked/Unchecked. Checked means it is answered by an Agent and Unchecked means Unanswered (answered by User).

As an Agent (currently):
- Answered by Agent is the Answered Queue
- Answered by User is the Open (Unanswered) Queue

As a User (with this pull):
- Answered by Agent is the Open (Unanswered) Queue
- Answered by User is the Answered Queue

Cheers.

mlipok:
Interesting .. I have to check, because I did not know that it worked like that ... I will also implement this PULL and see how it is on the other CUSTOMER side.

but not in this year..... ;)

btw.
Thanks for your's answer

PR #5773, opened 2021-01-18
view.php checks wrong parameter for client auth

gerundt:
view.php used the parameter `id` for client auth but checked the parameter `t` for numeric.

This fix problems with *LDAP Authentication and Lookup* and *HTTP Passthru Authentication* plugin.

### DIFF
--- a/view.php
+++ b/view.php
@@ -39,7 +39,7 @@ elseif (isset($_GET['auth']) || isset($_GET['t'])) {
 
 if (@$user && is_object($user) && $user->getTicketId())
     Http::redirect('tickets.php?id='.$user->getTicketId());
-elseif ($thisclient && isset($_GET['id']) && is_numeric($_GET['t']))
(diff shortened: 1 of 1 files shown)


gerundt:
Mhhh. with a cleared browser cache the SSO still don't work. The user must click one time on the `login.php` link. This script then call the `UserAuthenticationBackend::processSignOn($errors, false)` function. 

This patch helps me:
```
--- view.php.org	Mon Jan 18 10:10:06 2021
+++ view.php	Mon Jan 18 12:31:51 2021
@@ -32,7 +32,7 @@
 }
 // Try autologin the user
 // Authenticated user can be of type ticket owner or collaborator
-elseif (isset($_GET['auth']) || isset($_GET['t'])) {
+elseif (isset($_GET['auth']) || isset($_GET['t']) || isset($_GET['id'])) {
     // TODO: Consider receiving an AccessDenied object
     $user =  UserAuthenticationBackend::processSignOn($errors, false);
 }
```

Ok for you?

Issue #5780, opened 2021-01-28
Found some problems for file upload


bjoern-st:
### Prerequisites

* [x ] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [ ] Do you have any errors in the PHP error log, or javascript console?
* [ x] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [ x] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/osTicket/osTicket/wiki/Github-Issues-Guidelines)

### Description

We're dealing with software storing its project data in sqlite files. So for creating a ticket it is necessary to upload the corresponding sqlite file to the ticket, Since sqlite is not in the predefined list I added .sqlite to the additional file extension list, but stil no upload passible. Error message is "This type of file is not allowed". 
So I added compressed files to the allowed file type list, tried to upload a zip file but the same error appears. 

Debugging the code I found two Problems: 
* for uploading the zip file the given file type is **application/x-zip-compressed**, which is not listed in the allowed file types
* for uploading the sqlite file the file type is empty 

I found the problem in the file filedrop.field.js in the upload function:

```php
      if (opts.allowedfiletypes.push && opts.allowedfiletypes.length) {
        for(var fileIndex = files.length;fileIndex--;) {
          if(!files[fileIndex].type || $.inArray(files[fileIndex].type, opts.allowedfiletypes) < 0) {
            opts.error(errors[3], files[fileIndex], fileIndex);
            return false;
          }
        }
      }
```

Since the file type is empty for sqlite files and missing for the zip file the function will return at this point and will not be able to check the file extensions and pass throug. 

### Steps to Reproduce

1. Add some strange file extension for ticket attachments
2. Create a new ticket and try to upload a file with that extension

**Expected behavior:**  File should be accepted for upload.

**Actual behavior:** Error appears **This type of file is not allowed**

### Versions

at least since 1.14.3

PR #5781, opened 2021-01-28
Fix for Issue #5780

bjoern-st:
This might be more a hotfix than a regular one but it works for me.

### DIFF
--- a/js/filedrop.field.js
+++ b/js/filedrop.field.js
@@ -473,6 +473,8 @@
     // Respond to an upload
     function upload() {
       stop_loop = false;
+      var allowedType = true;
+      var allowedextension = false;      
 
       if (!files) {
         opts.error(errors[0]);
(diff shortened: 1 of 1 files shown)


ChrisTG742:
I don't think that setting the default to "true" is a good idea here. For sanity checks I'd always follow the "only if" strategy, so rather whitelisting than blacklisting. However, it's not that critical here.
What about doing it the "right way" an adding the missing mime-type to "upload/include/config/filetype.yaml"?

bjoern-st:
You're right .. The default value should of course be false .. 
What do you mean with "right way" .. Do you refer to the zip file problem or file extension problem?

ChrisTG742:
Huh? I thought that the "file extension problem #5780 " is the result of the missing mime-type for this "zip" flavour? Maybe I'm wrong, but adding the missing entries to filetype.yaml (or adding wildcard-entries) should cure both problems, shouldn't it?

bjoern-st:
Ah .. OK. Well yes, that would fix the zip file problem. 
But you have a setting to additionally allow file types to be uploaded .. for example my sqlite files ( *.sqlite ) 
But the processing of the additionally allowed file types was never reached because the function returns if the uploaded file is not in the allowed mime types .. and that is the wrong behavior ..

ChrisTG742:
Ah, okay, I finally got the point ;)
You are refering to that line, right?
`if(!files[fileIndex].type || $.inArray(files[fileIndex].type, opts.allowedfiletypes) < 0) {`
Shouldn't it be sufficient to change the "OR" to an "AND"?

PR #5792, opened 2021-02-04
Issue: Visibility Permissions

aydreeihn:
This commit fixes an issue with checking on choices for the DepartmentChoiceField and AgentSelectionField. An empty array is still considered to be set, so we needed to see if the choices were empty.


aydreeihn:
Wondering if we still need the isset check at all or if just empty will be ok?

PR #5796, opened 2021-02-08
 Add "signals" to allow plugins to modify some things in templates

informatico-madrid:
Hi, recently I had to modify the core in my project to allow some changes to the template. I have found this mode which I think might be interesting. Let's see what you think.

### DIFF
--- a/include/class.osticket.php
+++ b/include/class.osticket.php
@@ -160,6 +160,7 @@ class osTicket {
     }
 
     function getExtraHeaders() {
+	Signal::send('osticket.extraheaders', $this, $this->headers);
         return $this->headers;
     }
     function getExtraPjax() {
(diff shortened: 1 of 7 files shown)


JediKev:
@informatico-madrid 

Can you provide further clarification on what plugins would need these changes and how these signals would be used?

Cheers.

informatico-madrid:
I did This PR a lot of time ago. Remember and review it. I remember that i need This Signal in templete un orden to archive some requeriments on my old company. You can check the commits. Are self explicative

JediKev:
@informatico-madrid 

We did but not they are not very detailed. I was hoping for a detailed explanation on why each change was made and how they would be useful.

Cheers.

PR #5800, opened 2021-02-11
Show ticket status translation in Ticket View/Edit

gerundt:
Use the local name more often for ticket status in ticket view/edit.

We use now own ticket statuses with German translations. But many places in osTicket show still the English name. This PR fix at least some parts in the Ticket View/Edit.

### DIFF
--- a/include/staff/templates/status-options.tmpl.php
+++ b/include/staff/templates/status-options.tmpl.php
@@ -66,7 +66,7 @@ if (!$nextStatuses)
                 ><i class="<?php
                         echo $actions[$status->getState()]['icon'] ?: 'icon-tag';
                     ?>"></i> <?php
-                        echo __($status->getName()); ?></a>
+                        echo $status->getLocalName(); ?></a>
         </li>
     <?php
     } ?>
--- a/include/staff/ticket-open.inc.php
+++ b/include/staff/ticket-open.inc.php
@@ -423,7 +423,7 @@ print $response_form->getField('attachments')->render();
                                 $s->getId(),
                                 $selected
                                  ? 'selected="selected"' : '',
-                                __($s->getName()));
+                                $s->getLocalName());
                     }
                     ?>
                     </select>
--- a/include/staff/ticket-view.inc.php
+++ b/include/staff/ticket-view.inc.php
@@ -434,6 +434,7 @@ if($ticket->isOverdue())
                              echo sprintf('<span><a class="manage-collaborators preview"
                                     href="#thread/%d/collaborators/1"><span id="t%d-recipients"><i class="icon-group"></i> (%s)</span></a></span>',
                                     $ticket->getThreadId(),
+                                    __('Manage Collaborators'),
                                     $ticket->getThreadId(),
                                     $recipients);
                              }?>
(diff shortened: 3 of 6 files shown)


include/staff/ticket-view.inc.php
```diff
@@ -434,6 +434,7 @@ class="icon-folder-close-alt icon-fixed-width"></i> %s</a></li>',
                              echo sprintf('<span><a class="manage-collaborators preview"
                                     href="#thread/%d/collaborators/1"><span id="t%d-recipients"><i class="icon-group"></i> (%s)</span></a></span>',
                                     $ticket->getThreadId(),
+                                    __('Manage Collaborators'),
```
JediKev, line 437: This will cause issues. There are only three replacement variables (`%d`, `%d`, `%s`) yet you are passing 4 values.

include/staff/ticket-open.inc.php
```diff
@@ -423,7 +423,7 @@ class="<?php if ($cfg->isRichTextEnabled()) echo 'richtext';
                                 $s->getId(),
                                 $selected
                                  ? 'selected="selected"' : '',
-                                __($s->getName()));
```
JediKev, line 426: The purpose of `__()` is to translate the given string. Please see:
- https://github.com/osTicket/osTicket/blob/develop/include/class.translation.php#L1053-L1058

include/staff/templates/status-options.tmpl.php
```diff
@@ -66,7 +66,7 @@ class="action-dropdown anchor-right">
                 ><i class="<?php
                         echo $actions[$status->getState()]['icon'] ?: 'icon-tag';
                     ?>"></i> <?php
-                        echo __($status->getName()); ?></a>
```
JediKev, line 69: The purpose of `__()` is to translate the given string. Please see:
- https://github.com/osTicket/osTicket/blob/develop/include/class.translation.php#L1053-L1058

include/staff/ticket-view.inc.php
gerundt: Args, this line should not in the PR. 🙈 Looks like I don't disabled the change in the Github Desktop client. Will try to fix it. I am still a Git beginner.

include/staff/ticket-open.inc.php
gerundt: This maybe works with your default statuses which are in the transaltion files. But when you add own entries in the custom list *Ticket Statuses* this don't work because of the custom translation format:

https://github.com/osTicket/osTicket/blob/develop/include/class.list.php#L1271-L1273

include/staff/templates/status-options.tmpl.php
gerundt: This maybe works with your default statuses which are in the transaltion files. But when you add own entries in the custom list *Ticket Statuses* this don't work because of the custom translation format:

https://github.com/osTicket/osTicket/blob/develop/include/class.list.php#L1271-L1273

include/staff/ticket-view.inc.php
gerundt: Create the new PR #5803 without the line.

PR #5803, opened 2021-02-12
Show ticket status translation in Ticket View/Edit V2

gerundt:
Use the local name more often for ticket status in ticket view/edit.

We use custom ticket statuses with German translations. But many places in osTicket show still the English name. This PR fix at least some parts in the Ticket View/Edit.

### DIFF
--- a/include/class.list.php
+++ b/include/class.list.php
@@ -1439,12 +1439,6 @@ implements CustomListItem, TemplateVariable, Searchable {
 
         return $this->getLocalName();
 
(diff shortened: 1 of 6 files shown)


gerundt:
Any chance to get this upstream?

We have a installation with English and German translations and it is sad, that the German users see only the English words on so many places.

This would be a beginning for a better translation. 😉

protich:
@gerundt - I've refactored your PR to remove the need to explicitly request translated status names. It's not done by default... see #5885

PR #5805, opened 2021-02-16
Check CSRF token if the normal login form was used

pitbulk:
Some alternative Auth methods, like SAML, do not use the CSRF approach to validate the POST content, and instead has its own mechanism, SAML for example uses a signature validation mechanism.

The checkCSRFToken causes issues when SAML login process happens, where there is no CSRF token at all.

If we only check the CSRF when we know the normal login form was used, like for example, verifying that the POST 'userid' parameters exist, we prevent the SAML scenario to experience issues at the same time we keep the security in the other Auth mechanism.

### DIFF
--- a/scp/login.php
+++ b/scp/login.php
@@ -59,7 +59,7 @@ if ($_POST) {
     // use a different CSRF token. This will help ward off both parallel and
     // serial brute force attacks, because new tokens will have to be
     // requested for each attempt.
-    if (!$ost->checkCSRFToken()) {
+    if (isset($_POST['userid']) && !$ost->checkCSRFToken()) {
         $_SESSION['_staff']['auth']['msg'] = __('Valid CSRF Token Required');
         $redirect($_SERVER['REQUEST_URI']);
     }


protich:
@pitbulk - External authentication systems shouldn't be posting to login.php - it should register its own endpoint - e.g `/api/auth/saml` and handle it there.

pitbulk:
@protich, Thanks for the fast reply.

I agree with you that having its own endpoint is the ideal approach.

When I created the SAML extension, I followed what was done in the existing extensions like the [Oauth](https://github.com/osTicket/osTicket-plugins/blob/develop/auth-oauth), [auth-api](https://github.com/antriver/osticket-auth-api/tree/master/lib/ApiAuth) or the [cas](https://github.com/osTicket/osTicket-plugins/blob/develop/auth-cas) where no new endpoints were added.

In the past 2 similar PRs were accepted:
- https://github.com/osTicket/osTicket/pull/5453
- https://github.com/osTicket/osTicket/pull/4629

I have already customers that had to apply this change manually in order to prevent issues with the SAML extension,
So I thought it was ok to create this new PR.

If you are not ok introducing this change, I'm ok switching how the SAML extension was implemented and re-implement it with its "own endpoint", which will carry some amount of work, but I'm more concerned about the impact on customers, that will need to re-implement its SAML configuration which requires the administrator of the Identity Provider (a different admin than the one that  owns OsTicket) to register the new endpoints in their side.


The files on the plugins folder seem to not be directly accessible, so for example if I create an endpoint.php file on the saml plugin folder, http://osticket.example.com/include/plugins/auth-saml/endpoint.php is not accessible. How to make it accessible?

Do you have an example of an external authentication with its own endpoint?

protich:
@pitbulk - My comment is more of a feedback, not a rejection per se. We recently had to deploy an external auth plugin for our hosted platform - hence why I suggested it.

It's trivial to register an endpoint - in your plugin's bootstrap routine - simply register as shown below. 

```php
      Signal::connect('api', function ($dispatcher) {
            $dispatcher->append(
                url('^/auth/saml$', function () {
                    if (isset($_POST['SAMLResponse'])) {
                           // Handle authentication here.
                    }
                })
            );
        });
```

The bonus is the plugin now controls the `ACS URL`  as `https://your.domain.com/api/auth/saml` or whatever you wish the endpoint to be.

pitbulk:
Ok, thanks for the feedback, I will give it a try.

I believe that having my own endpoint will avoid any possible issue in the future caused by changes on the login view, so I'm ok with your suggestion

protich:
Closing - please reopen if it's still an issue.

PR #5815, opened 2021-02-19
Submit dialog on Enter keydown

miftygusein:
When deleting a lot of entries, being able to hit Enter on each confirmation prompt instead of moving and pointing the cursor will save time. Moreover, using Enter to confirm is a convention widespread across many operating systems.

I took the existing code you had for the Esc keydown and matched it to use the Enter key to click the confirm button. If you can think of a better way to accomplish this, then of course go for it.

### DIFF
--- a/scp/js/scp.js
+++ b/scp/js/scp.js
@@ -592,6 +592,13 @@ jQuery.fn.exists = function() { return this.length>0; };
 $.pjax.defaults.timeout = 30000;
 $(document).keydown(function(e) {
 
+   if (e.keyCode == 13 && !$('#overlay').is(':hidden')) {
+        $('div.dialog input[type=button].confirm').click();
+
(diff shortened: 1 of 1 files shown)


protich:
@mehov - We think it's indeed a good idea and plan to merge the PR. That said, we're doing a review to make sure this doesn't have unintended consequences on some dialog modals. Just wanted to say thank you!

JediKev:
@mehov 

As expected this prevents the enter key from creating a new line in text editors within popups (eg. Change Status popup). This will need to be totally reworked.

Cheers.

JediKev (commented):
After applying the changes the enter key does not create a new line in text editors within popups (eg. Change Status popup). This will need to be reworked so that existing functionality for everything in popups remains the same.

PR #5842, opened 2021-03-22
issue: Choices Field Sanitization

JediKev:
This addresses an issue where ChoicesField configuration is malformed due to sanitization. This adds a check for `$config['html']` and if it's `false` (not set) then we skip heavy sanitization.

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -4367,7 +4367,7 @@ class TextareaWidget extends Widget {
                 .' placeholder="'.$this->field->getLocal('placeholder', $config['placeholder']).'"'; ?>
             id="<?php echo $this->id; ?>"
             name="<?php echo $this->name; ?>"><?php
-                echo Format::htmlchars($this->value, true);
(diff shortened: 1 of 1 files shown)


include/class.forms.php
```diff
@@ -4367,7 +4367,7 @@ function render($options=array()) {
                 .' placeholder="'.$this->field->getLocal('placeholder', $config['placeholder']).'"'; ?>
             id="<?php echo $this->id; ?>"
             name="<?php echo $this->name; ?>"><?php
-                echo Format::htmlchars($this->value, true);
+                echo Format::htmlchars($this->value, ($config['html']) ? true : false);
```
protich: Minor change - 
```php
echo Format::htmlchars($this->value, ($config['html'])); 
```

protich (changes requested):

include/class.forms.php
JediKev: Resolved

protich (approved):

PR #5850, opened 2021-03-30
Fix #5849

### LINKED ISSUE #5849: Predefined answers are formatted in HTML, although "Enable Rich Text" is disabled
### Description

Predefined answers are formatted in HTML, although the setting "Rich Text" under"Settings" -> "System" -> "Enable Rich Text" is deactivated (check box not activated)

To describe the bug in detail:
When creating a predefined answer, the wysiwyg editor is displayed. So you cannot create a predefined answer in plain text, because the editor wraps everything in HTML tags (e.g. <p> tags). Subsequently, if you use a predefined answer for a ticket response, the provided HTML will not be rendered (which is correct), but I don't want any HTML tags at all.

---

I expected that the wysiwyg editor isn't shown when "Enable Rich Text" is disabled.

---

osTicket Version | v1.15.2
Microsoft Edge | v89.0.774.63 (64-Bit)

iseries:
Bugfix. See: #5849 Predefined answers are formatted in HTML, although "Enable Rich Text" is disabled

### DIFF
--- a/include/staff/cannedresponse.inc.php
+++ b/include/staff/cannedresponse.inc.php
@@ -1,4 +1,6 @@
 <?php
+global $cfg;
+
 if(!defined('OSTSCPINC') || !$thisstaff) die('Access Denied');
 $info=$qs = array();
 if($canned && $_REQUEST['a']!='add'){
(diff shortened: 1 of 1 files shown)


iseries:
Fix is not correctly working.

PR #5874, opened 2021-04-27
Issue: Missing null check in Staff->updatePerms()

gerundt:
If a agent has no permissions and you will give him/her the first permission, the class Staff->updatePerms() wrote a `array_key_exists() expects parameter 2 to be array, null given` warning.

Problem is, that $permissions->perms is null. With a RolePermission->exists() function is the code easier to read an don't show the warning.

### DIFF
--- a/include/class.role.php
+++ b/include/class.role.php
@@ -297,6 +297,10 @@ class RolePermission {
             $this->perms = array();
     }
 
+    function exists($perm) {
+        return @array_key_exists($perm, $this->perms);
+    }
+
     function has($perm) {
         return (bool) $this->get($perm);
     }
--- a/include/class.staff.php
+++ b/include/class.staff.php
@@ -1398,7 +1398,7 @@ implements AuthenticatedUser, EmailContact, TemplateVariable, Searchable {
         }
         $permissions = $this->getPermission();
         foreach ($vars as $k => $val) {
-             if (!array_key_exists($val, $permissions->perms)) {
+             if (!$permissions->exists($val)) {
                  $type = array('type' => 'edited', 'key' => $val);
                  Signal::send('object.edited', $this, $type);
              }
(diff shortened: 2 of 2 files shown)


JediKev:
@gerundt 

I cannot replicate the issue using v1.15.2, Apache, PHP 7.4, and MySQL 8.0. When I create an Agent with no permissions and update the Agent afterwards by adding a first permission and I dump `$permissions->perms` I get an empty array which isn't `null` and does not throw a warning.

Can you please provide the exact steps to replicate the issue as well as your environment information?

Cheers.

gerundt:
@JediKev 

We use the v1.15.2 version from Git with some modifications. Its my [MH branch](https://github.com/gerundt/osTicket/commits/MH).

Here our environment:

* Web Server Software: Microsoft-IIS/10.0
* MySQL Version: 5.6.35
* PHP Version: 7.2.11

I will look why `$permissions->perms` is nothing.

gerundt:
@JediKev

Mhh, problem seems `JsonDataParser::parse()` in the constructor from the class `RolePermission`:
https://github.com/osTicket/osTicket/blob/develop/include/class.role.php#L295

My agent (`ldap` backend) has a empty string in the `permissions` column from `ost_staff`.

When first time the function `getPermission()` in the class `Staff` is calling, it will call `new RolePermission($this->permissions)`:
https://github.com/osTicket/osTicket/blob/develop/include/class.staff.php#L787

`$this->permissions` return an empty string. Inside the constructor from the class `RolePermission` will `JsonDataParser::parse()` return `null` for an empty string.

Inside the `JsonDataParser` it will call `JsonDataParser::decode()` which use the `json_decode()` from PHP. This finally returns the `null` for an empty string. Maybe a Windows bug/problem?

include/class.role.php
```diff
@@ -297,6 +297,10 @@ function __construct($perms) {
             $this->perms = array();
     }
 
+    function exists($perm) {
+        return @array_key_exists($perm, $this->perms);
```
protich, line 301:
```php
return array_key_exists($perm, $this->perms ?: array());
```

protich (changes requested):

protich:
Thanks @gerundt for your continued help. Closing this PR since your commit is included in PR #5883.

PR #5897, opened 2021-05-22
Prevent session fixation

isaacnia:
Hi,
When authenticating a user, it doesn’t assign a new session ID, making it possible to use an existent session ID. The attack consists of obtaining a valid session ID (e.g. by connecting to the application), inducing a user to authenticate himself with that session ID, and then hijacking the user-validated session by the knowledge of the used session ID.


JediKev:
@isaacnia 

We have a pull request coming for this soon. Thank you for the pull though.

Cheers.

PR #5900, opened 2021-05-26
Issue: Deleted Forms on Help Topics

aydreeihn:
This commit fixes an issue we had with forms that have been deleted that were attached to a Help Topic. If a custom form is attached to a help topic and then the form is deleted, we put a flag on the form so that existing tickets still show the fields. However, if you are creating new tickets with those help topics, the deleted form is still showing up even though it should be hidden.

A check has now been added in the getFormsForHelpTopic method to skip deleted forms when creating new Tickets.

### DIFF
--- a/include/ajax.forms.php
+++ b/include/ajax.forms.php
@@ -29,8 +29,9 @@ class DynamicFormsAjaxAPI extends AjaxController {
         }
 
         foreach ($topic->getForms() as $form) {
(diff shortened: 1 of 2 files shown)


aydreeihn:
Should we change isDeletable to use the hasFlag function?

protich:
Yes - please.

aydreeihn:
@protich Updated!

Issue #5846, opened 2021-03-25
Forms not being assigned to tickets when help topic is assigned by email box or ticket filters


irefuse:
### Prerequisites

* [x] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [ ] Do you have any errors in the PHP error log, or javascript console?
* [x] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [x] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/osTicket/osTicket/wiki/Github-Issues-Guidelines)

### Description

When assigning a help topic using the email address configuration, the forms attached to the help topic do not get added to tickets created via email. This can also be seen when setting the help topic via Ticket Filters.

### Steps to Reproduce

1. create an email address in Admin Panel->Emails->Emails
2. set the help topic to a help topic that contains a form
3. send an email to the email box that has a the help topic assigned to it
4. force the cron to check for emails
5. ticket is generated with the proper help topic but the form is not attached to the ticket

**Expected behavior:** [What you expected to happen]
form should be attached to the generated ticket since the form is assigned to the help topic. Settings are in screenshots below
![image](https://user-images.githubusercontent.com/2688319/112532518-e7d3b200-8d76-11eb-820a-9e339003eec3.png)
![image](https://user-images.githubusercontent.com/2688319/112532678-18b3e700-8d77-11eb-92ce-1693653cf19b.png)
![image](https://user-images.githubusercontent.com/2688319/112532730-28cbc680-8d77-11eb-8900-3bfef3a7808f.png)
![image](https://user-images.githubusercontent.com/2688319/112532780-3a14d300-8d77-11eb-8c6e-b18dac4a8e37.png)

**Actual behavior:** [What actually happened]
the form does not get attached to the generated ticket

### Versions

Admin panel -> Dashboard -> Information which also additionally gives you information about your server.

Server Information
--
osTicket Version | 1.15-git (?)  Upgrade— cb6766e is available
Web Server Software | Microsoft-IIS/8.5
MySQL Version | 10.4.12
PHP Version | 7.3.22

Also, please include the OS and what version of the OS you're running. As well as your browser and browser version.
Server: Windows Server 2012 R2
Client: Windows 10 20H2; Microsoft Edge (Chromium) and Google Chrome

JediKev:
@irefuse 

[...]

PR #5906, opened 2021-06-02
Assign forms to tickets which are created by mail

gerundt:
Now should forms assigned to a ticket which created by mail. Of course are the values empty. Should fix one part of the bug #5846

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -3959,6 +3959,53 @@ implements RestrictedAccess, Threadable, Searchable {
         return db_fetch_array(db_query($sql));
     }
 
+    /**
+     * Handle the forms associate with the help topics.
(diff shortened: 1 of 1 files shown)


gerundt:
I create the new PR #6185 which has a option to enable the feature.

JVC-MSF-OCB:
very nice!

how does this work with the latest versions?

gerundt:
> very nice!
> 
> how does this work with the latest versions?

@JVC-MSF-OCB There is with #6185 a new version from the pull request, which should also work with the latest version.

JVC-MSF-OCB:
Ok cool!

johannarvaez-cmd:
Do you have this code for version 1.18?
I tried the one you shared and it didn't work.

Issue #5930, opened 2021-07-26
Setup typo or just needs to be updated


drissical:
### Prerequisites

Standard setup from zip file

### Description

Not a major or even a minor bug but just a verbiage issue. Setting up a test instance of the software and I was just putting in a simple admin password of "admin" which the tooltip shows just needs to be 5 characters, however it fails because it is not 6 characters in the error message.  I moved the tooltip down in the image so that you can see the tooltip and the error message at the same time.

### Steps to Reproduce

Setup a new install and attempt to use a 5 character password for the password.

**Expected behavior:** Albeit bad practice to use a 5 character password, I still expected it to let me do it as that is what was described. 

**Actual behavior:** Better, requiring 6 characters with an error only having 5.

### Versions
osTicket 1.15.2
Debian Buster

![image](https://user-images.githubusercontent.com/8355446/127072124-9b295f3a-2a6f-4fd8-8df8-878e74ce37fc.png)

JediKev:
@drissical 

Thank you for the report! Below is the pull to address the issue. It will soon be merged and packaged in a future release.
- https://github.com/osTicket/osTicket/pull/5931

Cheers.

PR #5931, opened 2021-07-27
issue: Setup Admin Password Heltip Verbiage

JediKev:
This addresses #5930 where the Helptip for the Admin Password field on Install is incorrect. The Helptip states the password has to be a minimum of 5 characters where in reality it requires a minimum of 6. This updates the Helptip as well as the inline string from `Must be five (5) characters or more.` to `Must be six (6) characters or more.`. In addition, this updates the phrase `(Five characters min)` to `(Six characters min)` in `include/class.validator.php` as well as updates the check for less than 5 characters.

### DIFF
--- a/include/class.validator.php
+++ b/include/class.validator.php
@@ -111,8 +111,8 @@ class Validator {
                     $this->errors[$k]=$field['error'];
                 break;
             case 'password':
-                if(strlen($this->input[$k])<5)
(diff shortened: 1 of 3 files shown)


ntozier:
Shouldn't this also include /include/class.validator.php @line 115.
see `case 'password'`

JediKev:
@ntozier Yes, thank you. (grep-ed for exact verbiage)

PR #5940, opened 2021-07-31
display clickable telephone URL

am-css:
display a clickable URL for a phone number


JediKev:
@am-css This can introduce an issue where exports will have a clickable link as well.

PR #5963, opened 2021-09-09
Use events for "Mark as (un)answered"

gerundt:
This switch the *Mark as (un)answered* thread entries from activity to event:

image

Unfortanlty I don't know how I get the new `mark-answered` and `mark-unanswered` names in the `event` table!?

I think it will be done by a SQL file in the `include\upgrader\streams\core` path:

```
/**
* @signature XXX
* @version v1.15.4
* @title Mark as (un)answered events
*
* This patch adds new events to the database to allow "mark as (un)answered" events
*
*/
-- Insert new events
INSERT INTO `%TABLE_PREFIX%event` (`id`, `name`, `description`)
VALUES
    ('','mark-answered',''),
    ('','mark-unanswered','');

-- Finished with patch
UPDATE `%TABLE_PREFIX%config`
   SET `value` = 'XXX', `updated` = NOW()
   WHERE `key` = 'schema_signature' AND `namespace` = 'core';
```

But how I know the signature ID? I hope you can help me! 😊

### DIFF
--- a/include/ajax.tickets.php
+++ b/include/ajax.tickets.php
@@ -1640,8 +1640,7 @@ class TicketsAjaxAPI extends AjaxController {
 
                 // Add success messages and log activity
                 $_SESSION['::sysmsgs']['msg'] = sprintf(__('Ticket marked as %s successfully'), $action);
-                $msg = sprintf(__('Ticket flagged as %s by %s'), $action, $thisstaff->getName());
-                $ticket->logActivity(sprintf(__('Ticket Marked %s'), ucfirst($action)), $msg);
+                $ticket->logEvent('mark-' . $action);
                 Http::response(201, $ticket->getId());
             }
         }
--- a/include/class.thread.php
+++ b/include/class.thread.php
@@ -2026,6 +2026,8 @@ class ThreadEvent extends VerySimpleModel {
     const VIEWED    = 'viewed';
     const MERGED    = 'merged';
     const UNLINKED    = 'unlinked';
+    const MARK_ANSWERED   = 'mark-answered';
+    const MARK_UNANSWERED = 'mark-unanswered';
 
     const MODE_STAFF = 1;
     const MODE_CLIENT = 2;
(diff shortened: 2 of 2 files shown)


JediKev:
@gerundt 

I will put a quote below that should help:

>After making the appropriate changes to the install script, `setup/inc/streams/core/install-mysql.php`, then create an MD5 hash of that file. This will be the new `schema_signature` of the project. The current `schema_signature` is in `include/upgrader/streams/core.sig`. You will need to create a patch from the current signature to the new one. Your patch file will be `00000000-00000000.patch.sql`, where the first zeros are the first eight chars of the current signature, and the second zeros are the first eight chars of the new signature.
>
>Your patch should make the necessary changes to the database and change the `schema_signature` in the `%config` table to match the new signature. Refer to the end of one of the other patches for an example query.
>
>You should also change the contents of `include/upgrader/streams/core.sig` to contain the MD5 hash of the install script. This change is what prompts the upgrade process.

Cheers.

JediKev:
@gerundt 

With this being said, you do know that marking as answered/unanswered already adds a system message to the thread?
Screen Shot 2021-09-09 at 09 03 32

You may want to remove this current functionality in favor of the new functionality as you wouldn't want duplicate actions. With this being said it'd have to be reviewed and discussed internally to see if this is something we'd like to do.

Cheers.

JediKev:
@gerundt 

Sorry for the spam but I also realized that you would mark this as v1.15.4 but when an upgrade is required we change the entire version so it'd actually be `v1.16` (or whatever we would decide to call it). In this case there are even more things that need to be done such as updating the versions in `bootstrap.php`, etc.

Cheers.

gerundt:
@JediKev

> With this being said, you do know that marking as answered/unanswered already adds a system message to the thread?

Yes, I know it. And they are ugly! 🙈

We use the *Mark as answered* very often and it use a lot of space in the thread. Because of that I switched it to a real event where you only need 1 line. Only when you add a comment it will still create a system message.

> With this being said it'd have to be reviewed and discussed internally to see if this is something we'd like to do.

Of couse I understand this! I have also no problem when it first get in `v1.16`. 👍

**Thank you for your cool ticket system! I use it daily on work!** 😍

gerundt:
Sad I missed the `v1.16` release. 🙁 So the next change will be first `v1.17` correct?

gerundt:
@JediKev I read that v1.17 knock on the door. Any chance to get this to the new version?

JediKev:
@gerundt 

We still need to review, make a patch, etc. So potentially but we'll have to see what we have time for this release. 👍🏻 

Cheers.

PR #5964, opened 2021-09-14
Issue: AJAX use wrong "Due Date"

gerundt:
Because the *Due Date* are converted to  DB timezone the AJAX must use this value directly. If not it update the page with a wrong time.

### DIFF
--- a/include/ajax.tasks.php
+++ b/include/ajax.tasks.php
@@ -245,6 +245,9 @@ class TasksAjaxAPI extends AjaxController {
                       );
 
                 switch (true) {
+                    case ($fid == 'duedate'):
+                        $clean = Format::datetime($task->duedate);
+                        break;
                     case $field instanceof DateTime:
                     case $field instanceof DatetimeField:
                         $clean = Format::datetime((string) $field->getClean());
--- a/include/ajax.tickets.php
+++ b/include/ajax.tickets.php
@@ -665,6 +665,9 @@ class TicketsAjaxAPI extends AjaxController {
                       );
 
                 switch (true) {
+                    case ($fid == 'duedate'):
+                        $clean = Format::datetime($ticket->getEstDueDate());
+                        break;
                     case $field instanceof DateTime:
                     case $field instanceof DatetimeField:
                         $clean = Format::datetime((string) $field->getClean());


gerundt:
Sorry, I am runned out of time yesterday! Here more details how the error happens:

When updating the due date in the ticket/task view, the date/time will correct save to the database but AJAX use the wrong timezone when updating the page. First after a reload of the page will show the correct time.

image

JediKev:
@gerundt 

So I'm not able to replicate this issue. My server timezone is UTC, my DB timezone is UTC, my osTicket System Timezone is CEST (Europe/Amsterdam), and my Agent Profile timezone is CEST (Europe/Amsterdam). I go to a Ticket/Task, click on the Due Date field, update the time to an hour or so in the future, save, the page shows the correct date/time, and refreshing the page shows the same thing.

Any other steps to replicate this issue?

Cheers.

gerundt:
@JediKev

* Our MySQL server uses `SYSTEM` as timezone. It is a German Windows 2016 server which has `UTC+01:00` as timezone.
* The timezone from osTicket is `Europe / Berlin`.
* My agent has `System Default` as time zone. It is a German Windows 10 also with `UTC+01:00`.

When I use (_without my patch_) `12` as hour, AJAX will use `10` as hour. So `-2` hours. Maybe a problem with *summer/winter* time in Germany?

I have the same problem, when I switch my agent from `System Default` to `Europe / Berlin`

Thanks for the help!

JediKev:
@gerundt 

Okay I was finally able to replicate this by forcing MySQL to `+01:00` (literal offset) and using `Europe/Berlin` as my Default System Timezone. I can see now that your patch is just putting a band-aid on the issue and fixing this surface level problem, however, the issue lies much deeper and we really need to solve the root of the issue.

With that being said, I can see that the date-picker in the popup is showing date-times in CEST (which is currently correct). Let's say I then choose a due date of `April 7, 2022 17:00 CEST`. On submit it goes through the `editField()` method which then gets the field value with `(string) $field->getClean()`. I can then see that `getClean()` returns the appropriate date (eg. `2022-04-07 17:00:00 CEST`) and is passed to `Format::datetime()` (as it's an instance of DatetimeField). In `Format::datetime()` we immediately pass it to `self::__formatDate()`. This is where I'm currently at and investigating further.

Cheers.

gerundt:
@JediKev 

If you find a better way to solve the problem, go for it! 😊

Good luck! 👍

PR #5966, opened 2021-09-22
Disable New Ticket popup window

ramrajone:
/include/staff/ticket-open.inc.php

making it optional?

<?php
// Popup user lookup on the initial page load (not post) if we don't have a
// user selected
if (!$_POST && !$user) {?>
setTimeout(function() {
$.userLookup('ajax.php/users/lookup/form', function (user) {
window.location.href = window.location.href+'&uid='+user.id;
});
}, 100);
<?php
} ?>

### DIFF
--- a/include/class.filter_action.php
+++ b/include/class.filter_action.php
@@ -320,6 +320,20 @@ class FA_RouteDepartment extends TriggerAction {
         }
     }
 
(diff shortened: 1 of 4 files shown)


JediKev:
@ramrajone I don't think you meant to open this here.

PR #5973, opened 2021-09-27
Issue: Delete Referrals

aydreeihn:
This commit fixes an issue where we were not deleting referrals when a Ticket was deleted. Now we will delete any referrals related to a thread when the thread is deleted.


protich:
@aydreeihn - I'm merging this but wanted to request that we add a core stream task, in the next release with db patches,  to delete orphaned referrals for previously deleted tickets.

PR #5975, opened 2021-09-30
FeatureRequest: #action-dropdown-stats  ( #opened / #total )

mlipok:
For description take a look here.
https://forum.osticket.com/d/99995-related-number-of-tickets

image

This information after using this patch, shows
number of (#opened / #all) tickets

Previously it was:
number of (#all) tickets

I plan to do the same with the number of tickets for the organization.
Moreover, I will add a color change if the number of open tickets is> 0

**Summary:**
The point is that the visible information "1 Open Ticket" was currently visible only after clicking on the number (4)
, and now it is immediately displayed as (1/4) so without clicking you can see how many tickets a given user has opened.
It helps agents to solve other tickets opened by the same user/client because when agents do the job for current ticket, then he already know (without additional click) that the user/client mentioned above, has also other opened tickets.

### DIFF
--- a/include/staff/ticket-view.inc.php
+++ b/include/staff/ticket-view.inc.php
@@ -397,9 +397,9 @@ if($ticket->isOverdue())
                         if ($user) { ?>
                             <a href="tickets.php?<?php echo Http::build_query(array(
                                 'status'=>'open', 'a'=>'search', 'uid'=> $user->getId()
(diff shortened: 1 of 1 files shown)


mlipok (commented):
now if any ticket is opened then the background and color is changing
if there are only closed tickets then the old format i used
I mean it will display (1 / 4) or (4)
image

mlipok:
currently this looks like:
image

PR #5977, opened 2021-10-05
issue: List Item Properties On Mouseover

JediKev:
This addresses an issue where when Inline Edit was introduced it removed the functionality for the List Item Properties popup on mouseover. This restructures the `display()` method for DynamicListItem to display the value with an info icon that on mouseover will display the associated List Item Properties.

### DIFF
--- a/include/class.list.php
+++ b/include/class.list.php
@@ -797,14 +797,13 @@ class DynamicListItem extends VerySimpleModel implements CustomListItem {
     }
 
     function display() {
-
-        return $this->getValue();
         //TODO: Allow for display mode (edit, preview or both)
-        return sprintf('<a class="preview" href="#"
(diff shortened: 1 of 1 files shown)


JVC-MSF-OCB:
Hey is it possible that with the latest upgrade the (i) icon was dropped? The hover works, but instead we have the edit icon.
image

JediKev:
@logmsf 

We are removing/reverting it soon with:
- https://github.com/osTicket/osTicket/pull/5994

However, this hasn't been merged yet so if you are using v1.15.4 then you should see it. Regardless, you can ignore this as it's going away soon anyways.

Cheers.

JVC-MSF-OCB:
@JediKev
Got you! Thanks for the note ;-)
We are indeed on v1.15.4. Any sights on the next release? (to see if we should wait for it or apply the revert)

Issue #5893, opened 2021-05-17
[Feature Request] Personal Agent Avatar


Ressy66:
Request option to add personal avatar on an agents profile page so an agent can upload an image inplace of the default if they wish.

Thanks

PR #5979, opened 2021-10-06
Personal agent avatar#5893

Koch1:
Hi. I'm starting to help the Free Software and Open Source community. Could give feedback.
I put the  option to change the avatar on the profile page, you have the Avatar Source option "Built-In/Oscar's A-Team configured. Avatars will be saved in the users database of the staff table, column avatar_data. 

issue #5893 

thank you


ntozier:
How does this handle the existing gravatar Agent avatars?
Did you add configuration options to disable gravatar (Is it global or by agent or both)?

Koch1:
In Agents Settings you have Avatar Source, in which you have another choice that is not Gravatar. The option "Built-In/Oscar's A-Team", when configured for this option, the attendants will be able to exchange icons, I think it is a global variable. If you have active in Avatar Source another Gravatar choice, you will not be able to change the images.
Maybe it would be better to put another option in the select "Avatar Source" or maybe each attendant can choose the option?
I will be available to change, evolve the idea and redo it.
Thanks

JediKev:
@Koch1 

In this case we will leave avatars to Gravatar as it doesn't require code changes.

Cheers.

Issue #5981, opened 2021-10-06
Password error message


Koch1:
### Prerequisites

* [S] Can you reproduce the problem in a fresh installation of the "develop" branch? 
* [N] Do you have any errors in the PHP error log, or javascript console? 
* [S] Did you check the [osTicket forums](https://forum.osticket.com/)? 
* [S] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported? 

### Description

I fixed the error referring to the user screen, Button Manage Account, Tab Manege Access, to change the password. It did not show error messages regarding the password, it brings generic message.  
In the code has password validation, no messages were sent to interface

PR #5982, opened 2021-10-06
Password error message #5981

Koch1:
Hi. I'm starting to help the Free Software and Open Source community. Could give feedback.

I fixed the error referring to the user screen, Button Manage Account, Tab Manege Access, to change the password. It did not show error messages regarding the password, it brings generic message.
In the code has password validation, I just sent the error message to the user interface
#5981 

thank you

### DIFF
--- a/include/ajax.users.php
+++ b/include/ajax.users.php
@@ -228,6 +228,10 @@ class UsersAjaxAPI extends AjaxController {
 
             if ($errors['err'])
                 $info['error'] = $errors['err'];
+            else if($errors['passwd1']){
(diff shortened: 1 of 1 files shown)


JediKev:
@Koch1 

After attempting to replicate your issue I think I see what's going on. It appears the errors do in-fact get returned to the front-end and they are displayed correctly (without your pull request). What's happening is that since the errors are on the second tab you don't see them initially unless you click on that tab. We will do some work to highlight the tab red if it has errors that are "hidden" so that it's more intuitive and user-friendly.

With this being said thank you for your pull as it brought this matter to light.

Cheers.

PR #5994, opened 2021-10-08
Revert "issue: List Item Properties On Mouseover"

JediKev:
This reverts commit a6a719239b1513683daa63f7b630a1aa43fefd8d as putting the preview in the display messes up exports and multi-selection.

### DIFF
--- a/include/class.list.php
+++ b/include/class.list.php
@@ -797,13 +797,14 @@ class DynamicListItem extends VerySimpleModel implements CustomListItem {
     }
 
     function display() {
+
+        return $this->getValue();
         //TODO: Allow for display mode (edit, preview or both)
-        return sprintf('%s &nbsp;<a class="preview" href="#"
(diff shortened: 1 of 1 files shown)


JVC-MSF-OCB:
We applied this revert, however I see that the (i) icon is displayed for any list item, even when they don't have additional properties making the tooltip redundant in those cases...

JediKev:
@logmsf 

Well you didn’t apply the revert correctly then. This revert only returns the value without the icon.

Cheers.

JVC-MSF-OCB:
Hey @JediKev I did indeed misunderstand hehe. Now I get it.

I was wondering if there were any plans then to re-integrate something similar where agents would be able to see these additional properties for list items?

JediKev:
@logmsf 

It’s definitely on our bug sheet to look into.

Cheers.

Issue #5851, opened 2021-03-30
Cant have Reply-To filter working


schoubi:
### Description

Hi,

I've read the previous resolved issues about Reply-To filter, but infortunately i can't make it worked...

Like others, we have a web form that use a "From: Dont Reply <webadmin@domain.com>" and "To: osticket@domain.com". The reply-to is "Reply-To: "Bill Baroud" <bill.baroud@domain.com>".

The filter have this conditions :

Rules:
   - Match All : 
   - User/Email equals nobody@domain.com
Filter action :
  - Use Reply-To Email :heavy_check_mark: 

When we use the form, the user in the ticket list is always «Dont Reply»....

I tried many Rules, even one that catch all emails, without success...

Can you give me a hand ?

### Versions

osTicket Version         |	v1.14.5
Web Server Software |   Apache/2.4.6 (CentOS) OpenSSL/1.0.2k-fips mod_fcgid/2.3.9 PHP/5.6.40 mod_wsgi/3.4 Python/2.7.5
MySQL Version          |   5.5.68
PHP Version               |    5.6.40

Email Headers :

```
Return-Path: <webadmin@domain.com>
Received: from smtp.domain.com
[...]
Date: Tue, 30 Mar 2021 14:08:38 +0200
Message-Id: <2021031208.12UC8cL1019257@domain.com>
X-Authentication-Warning: www.domain.com: apache set sender to webadmin@domain.com using -f
To: osticket@domain.com
Subject: document no 4636
X-PHP-Originating-Script: 0:system.mail.inc
MIME-Version: 1.0
Content-Type: text/plain; charset=UTF-8; format=flowed; delsp=yes
Content-Transfer-Encoding: 8Bit
X-Mailer: Drupal Webform (PHP/7.3.25)
Sender: webadmin@domain.com
From: "Dont Reply" <webadmin@domain.com>
Reply-To: "Bill Baroud" <bill.baroud@domain.com>
```

protich:
This is a known issue - I would suggest you adjust your Drupal script to use User's name in From address as shown below (for now until we release a fix).

```
From: "Bill Baroud" <webadmin@domain.com>
```

schoubi:
Thanks for your quick answer, we will try that.

This will create a user "Bill Baroud" in osticket with his proper mail address ?

ChrisTG742:
Same here. Unfortunately I can't solve this by changing the "From"-address.
[...]

PR #6010, opened 2021-10-27
Update class.mailfetch.php (fix for #5851)

ChrisTG742:
Fix for #5851: Respect "reply-to" address if present.
(See https://forum.osticket.com/d/494-email-header-reply-to-is-ignored/8 for details)


ChrisTG742:
Please ignore the scp.css file. I accidently commited to the wrong branch ;-)

PR #6012, opened 2021-10-27
Fix for #5851: Respect "reply-to" address if present.

ChrisTG742:
Fix for #5851: Respect "reply-to" address if present.
(See https://forum.osticket.com/d/494-email-header-reply-to-is-ignored/8 for details)


protich:
osTicket already supports use of Reply-To - see #5851 for reference.

PR #6014, opened 2021-10-29
issue: Improved Validation for Variable Names

JediKev:
This addresses an issue where a variable name with `/` for example throws databases errors when reindexing cdata tables. This improves the regex validation for variable names to ensure alphanumeric characters only.

### DIFF
--- a/include/class.dynamic_forms.php
+++ b/include/class.dynamic_forms.php
@@ -871,7 +871,7 @@ class DynamicFormField extends VerySimpleModel {
                 /* `variable` is used for automation. Internally it's called `name` */
                 ), "name");
         }
-        if (preg_match('/[.{}\'"`; ]/u', $this->get('name')))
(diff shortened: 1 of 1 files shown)


include/class.dynamic_forms.php
```diff
@@ -871,7 +871,7 @@ function isValid() {
                 /* `variable` is used for automation. Internally it's called `name` */
                 ), "name");
         }
-        if (preg_match('/[.{}\'"`; ]/u', $this->get('name')))
+        if ($this->get('name') && !preg_match('/^(?!\d)([[:alnum:]]|_|-)+$/', $this->get('name')))
```
protich: Unicode  support?  `/u`

JediKev: Roger roger. I’ll test it then update.

JediKev: Updated.

PR #6015, opened 2021-10-29
Fix problem with HTML elements in plain text responses

arneschuldt:
**Problem:**

In a non-rich-text setting, osTicket includes HTML elements if canned responses are added to a response. [This problem has been discussed, for instance, in a thread on the osTicket forum.](https://forum.osticket.com/d/96411-agent-responses-show-html-markup)

**Background:**

The problem came in with commit a1116036afff42b41d057f38b8c1ac3f50ddbf7b, probably during an upgrade of the Redactor library. Since this commit, a canned response loaded from the server always turns the response into a rich-text response on the client side. However, the server (rightly) expects non-rich-text content so that the response is malformatted later in this process.

**Solution:**

The solution to fix this is quite easy. The response HTML element is annotated the `.richtext` if it is meant to contain rich text. Hence, the selector for Redactor can be extended with this class. Consequently, rich text is only used if that is intended, i.e., no more malformatting with unwanted HTML elements.

### DIFF
--- a/scp/js/scp.js
+++ b/scp/js/scp.js
@@ -231,7 +231,7 @@ var scp_prep = function() {
                 success: function(canned){
                     //Canned response.
                     var box = $('#response', fObj),
-                        redactor = $R('#response');
+                        redactor = $R('#response.richtext');
                     if (canned.response) {
                         if (redactor) {
                             redactor.api('selection.restore');


JediKev:
@arneschuldt 

I was able to replicate the issue and confirm this works as expected. Your description of the problem and what this pull request solves is very well put.

The only issue is that you made this to the `develop` branch instead of the `1.14.x` branch. You will need to rebase this to `1.14.x`, force push the changes, and probably change the base branch here in Github (if it's not automatic). If you cannot do this let me know and I will do it for you (all credit, etc. remains intact). I will explain why this is needed so you and others know in the future. With osTicket, we maintain two versions of the software at a time so those who need time to upgrade to the newer version (for whatever reason) have the latest bug fixes and security patches whilst they get ready for upgrade. Since the issue in this scenario is present in both the v1.14 and v1.15 series this pull request needs to be made to the `1.14.x` branch. This way we merge it to `1.14.x` which addresses the issue in the next release of v1.14 and we then merge `1.14.x` into `develop` so the issue is addressed in the next release of v1.15.

My final note is when making pull requests try making a branch with a semi-descriptive name. Basically, anything other than `develop` or `master` or `1.14.x` (stray from making it too long though). For example, in this scenario I'd make a branch named `issue/plain-text-canned-responses` instead of `develop` (as it is now).

With all this being said, once this pull request has been rebased, I will run this by the team and if approved it will be included in the next set of releases. Thanks again.

Cheers.

arneschuldt:
@JediKev Thanks for reviewing the pull request and for your explanations.

I would like to ask you to perform the rebase for me this time. The background is: while it took some hours to trace down the problem in a running osTicket instance, I submitted the bugfix via the GitHub web interface in minutes. The problem now is that I cannot perform the requested changes as easily because I have no version checked out locally.

However, I promise to follow your instructions the next time I contribute to osTicket. Currently, I am investigating another problem that can hopefully solved also with a minimal change.

JediKev:
@arneschuldt 

Here is the rebase: #6017

Thanks again for the pull request!

Cheers.

PR #6034, opened 2021-12-07
Performance fix: join cdata table earlier

t-oster:
Hi,
we have about 250k tickets and if we load the list of open tickets, it takes more than 30s to load the page.
The culprit is the big SQL query, which joins the cdata table (containing two custom ticket fields) too late. 

In this fix/hack we just change the order of joins, so the cdata table gets joined earlier, which makes the query 10 times faster on our system.

The may be a more elegant way to ensure the table is joined earlier, so feel free to suggest another path. Without this patch, OsTicket is not usable for us with this high number of tickets.

### DIFF
--- a/include/class.orm.php
+++ b/include/class.orm.php
@@ -2626,6 +2626,12 @@ class SqlCompiler {
 
     function getJoins($queryset) {
         $sql = '';
+        //fix:move cdata table on top in the joins. 10x faster db with 250000+ tickets
+        if (array_key_exists("cdata", $this->joins)) {
+            $temp = ['cdata' => $this->joins['cdata']];
+            unset($this->joins['cdata']);
+            $this->joins = $temp + $this->joins;
+        }
         foreach ($this->joins as $path => $j) {
             if (!$j['sql'])
                 continue;


t-oster:
Hi, I just rebased it on 1.17.3 and it is still an issue. Our current Os-Ticket instance would not be workable. @JediKev please review and test with a huge number of existing tickets to confirm. I think this performance problem will affect many users in the future as soon as their ticket number grows.
This commit can be cherry-picked onto 1.17.3 without any conflicts.

davidGPA:
It also works for 1.18.1 with Mariadb 10.11.10.

Revellski:
Priceless fix, @t-oster. Thank you!

f3rrix:
Same for my 1.18.2 with ~22000 closed tickets.  Closed query was CPU bound in mysqld and timed out the ajax. With all db innodb optimizations done and doubling CPU, the fallback non-ajax still took a couple minutes to load.

With this change, about 50x faster.  Heck I could even get outrageous and turn back on top level ticket counts if I wanted.
Gold star ⭐for this , @t-oster . Difference between being able to use the system and not.

eltreno:
I agree this also fixed our issue when reading closed tickets from QTR or YEAR from 10+ minutes query to seconds.
Around 100k tickets.
We are still on osTicket (v1.17.4) with MariaDB 
Thanks @t-oster

santibespoke:
On 1.18.2, around 75K tickets. This fixed our issues on the Closed and Search tabs. Our mariadb process kept at 100%, never returning a response. Now it loads the Closed tab in 5-6 seconds. I'm in shock. Thanks @t-oster

t-oster:
Thank you for all the credit, but it was @kbastron who created the patch, I only submitted it.

PR #6038, opened 2021-12-16
issue: Malformed From Headers

JediKev:
This addresses an issue where malformed From headers cause fatal errors in osTicket. We utilize PEAR's Mail_RFC822 class, specifically the `parseAddressList()` method, to parse the address list and give us an array of recipients in the From header. When a recipient in the From header is malformed this method returns an empty array. In most cases this causes a fatal error as the Ticket now has "no user". This commit attempts to fix the malformed recipient before it gets to the `parseAddressList()` method to better ensure we can retrieve a recipient.

### DIFF
--- a/include/class.list.php
+++ b/include/class.list.php
@@ -797,13 +797,14 @@ class DynamicListItem extends VerySimpleModel implements CustomListItem {
     }
 
     function display() {
+
+        return $this->getValue();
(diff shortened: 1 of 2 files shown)


JediKev:
We can review this later. Think about if no email then reject ticket instead of fixing malformed email headers.

PR #6046, opened 2022-01-11
Fix slugify regex

davidglezz:
I think there was a typo because `a-z` is inside `A-z`. `A-z` range also match <code>[\\]^_\`</code> chars.


JediKev (approved):
Looks good! Thanks for the pull!

PR #6051, opened 2022-01-19
Auto-Close Answered Tickets

MarcBanyard:
This adds the ability to automatically close answered tickets after X amount of days of inactivity.

New Email Template created for "**Closed Tickets**"
The ability has also been added to the system to send Closed Ticket emails with the above template.

Both of the above can be set per department (**Admin Panel** > **Agents** > **Departments** > **[Department]**).

### DIFF
--- a/include/class.cron.php
+++ b/include/class.cron.php
@@ -28,6 +28,16 @@ class Cron {
     function TicketMonitor() {
         require_once(INCLUDE_DIR.'class.ticket.php');
         Ticket::checkOverdue(); //Make stale tickets overdue
+        global $cfg;
+        $depts=Dept::getDepartments();
+        $a = 0;
+        foreach($depts as $d) {
+            $b=ConfigItem::getConfigsByNamespace('dept.' . array_search($d,$depts),'autoclose_grace_period');
+            $a+=$b->ht['value']?$b->ht['value']:0;
+        }
+        if($a>0) {
+            Ticket::closeAnswered(); //Close answered tickets
+        }
         // Cleanup any expired locks
         require_once(INCLUDE_DIR.'class.lock.php');
         Lock::cleanup();
--- a/include/class.dept.php
+++ b/include/class.dept.php
@@ -447,6 +447,10 @@ implements TemplateVariable, Searchable {
         return $this->message_auto_response;
     }
 
+    function autoRespONClosedMessage() {
+        return ConfigItem::getConfigsByNamespace('dept.' . $this->getId() ,'closed_auto_response')->ht['value'];
+    }
+
     function noreplyAutoResp() {
          return $this->noreply_autoresp;
     }
@@ -490,6 +494,8 @@ implements TemplateVariable, Searchable {
         $ht['status'] = $this->getStatus();
         $ht['assignment_flag'] = $this->getAssignmentFlag();
         $ht['disable_reopen_auto_assign'] =  $this->disableReopenAutoAssign();
+        $ht['closed_auto_response'] = ConfigItem::getConfigsByNamespace('dept.' . $this->getId() ,'closed_auto_response')->ht['value'];
+        $ht['autoclose_grace_period'] = ConfigItem::getConfigsByNamespace('dept.' . $this->getId() ,'autoclose_grace_period')->ht['value']?:0;
         return $ht;
     }
 
@@ -868,6 +874,29 @@ implements TemplateVariable, Searchable {
         $this->group_membership = $vars['group_membership'];
         $this->ticket_auto_response = isset($vars['ticket_auto_response'])?$vars['ticket_auto_response']:1;
         $this->message_auto_response = isset($vars['message_auto_response'])?$vars['message_auto_response']:1;
+        if ($this->getId()) {
+            $chelper = new OsticketConfig;
+            $chelper->section = 'dept.' . $this->getId();
+            $chelper->load();
+            if (isset($chelper->config['closed_auto_response'])) {
+                $chelper->update('closed_auto_response', isset($vars['closed_auto_response']) ? $vars['closed_auto_response'] : 1);
+            } else {
+                $chelper->set('closed_auto_response', isset($vars['closed_auto_response']) ? $vars['closed_auto_response'] : 1);
+            }
+            if (isset($chelper->config['autoclose_grace_period'])) {
+                if (strlen($vars['autoclose_grace_period']) == 0) {
+                    $chelper->update('autoclose_grace_period','0');
+                } else {
+                    $chelper->update('autoclose_grace_period', isset($vars['autoclose_grace_period']) ? $vars['autoclose_grace_period'] : '0');
+                }
+            } else {
+                if (strlen($vars['autoclose_grace_period']) == 0) {
+                    $chelper->set('autoclose_grace_period','0');
+                } else {
+                    $chelper->set('autoclose_grace_period', isset($vars['autoclose_grace_period']) ? $vars['autoclose_grace_period'] : '0');
+                }
+            }
+        }
         $this->flags = $vars['flags'] ?: 0;
 
         $this->setFlag(self::FLAG_ASSIGN_MEMBERS_ONLY, isset($vars['assign_members_only']));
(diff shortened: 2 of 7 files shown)


ntozier:
There used to be two plugins that do this, I'm not sure that devs will accept this as a patch.
This might be functionality that they would want in a plugin not in core.

MarcBanyard:
There have been several people on the forums that have asked for this over the years and it is such a valuable feature that we used, but when we upgraded from an older version it would not work, so we have re-written it.

Having it as a plugin is great, but plugins are quite often not maintained and the plugins we found did not work, so we have written it as part of the core code.

It has been written so that if you want it you can use it, if not it will not break any functionality of the system, so why would you not add it to the core especially since the plugins that were written no longer work?

elekgeek:
Hope this feature gets added, thanks!

JediKev:
Thank you for the pull request but we are currently refraining from merging any new functionality and making any major changes to the current codebase while we focus on v2.0 (full codebase rewrite).

need4swede:
To confirm, the "days to close" counter only starts ticking after an initial reply from an agent, and resets after each reply? So it's only "x days of inactivity" after first reply? A bit confused about how that counter works. Thank you for your work!

MarcBanyard:
> To confirm, the "days to close" counter only starts ticking after an initial reply from an agent, and resets after each reply? So it's only "x days of inactivity" after first reply? A bit confused about how that counter works. Thank you for your work!

Its a "Close after X days of inactivity once an agent has replied".
The counter only kicks in when an agent replies to a ticket, it is also reset each time an agent replies to the ticket.

So if you set it to 4 days and a customer emailed in, the following would be true.
- The ticket would stay open indefinitely as the counter only kicks in on an agent's reply.
- If an agent has replied to a ticket and the customer replies within 4 days it cancels the timer and the ticket remains open.
- If an agent has replied to a ticket and a customer does not reply within the 4 days the ticket is automatically closed.
- Once a ticket is closed a customer can email back in and the ticket is reopened, the above starts counting again.

Hope this makes sense.

MarcBanyard:
> Thank you for the pull request but we are currently refraining from merging any new functionality and making any major changes to the current codebase while we focus on v2.0 (full codebase rewrite).

That's a pity.
I tried to ensure that there are no changes to the database and that not setting any of the options would not change any of the default behaviour of osTicket.

I know lots of people have asked for it on the forums previously and it would be great to get it merged into the core program as it does not cause any issues if it is not used.

This is the 2nd time I have re-written this into the code and the 2nd time I have requested it to be merged in.

I think osTicket is a great product and this feature would be of great benefit to be part of the core system as all of the paid for systems on the market have this as default.

need4swede:
> > To confirm, the "days to close" counter only starts ticking after an initial reply from an agent, and resets after each reply? So it's only "x days of inactivity" after first reply? A bit confused about how that counter works. Thank you for your work!
> 
> Its a "Close after X days of inactivity once an agent has replied". The counter only kicks in when an agent replies to a ticket, it is also reset each time an agent replies to the ticket.
> 
> So if you set it to 4 days and a customer emailed in, the following would be true.
> 
> * The ticket would stay open indefinitely as the counter only kicks in on an agent's reply.
> * If an agent has replied to a ticket and the customer replies within 4 days it cancels the timer and the ticket remains open.
> * If an agent has replied to a ticket and a customer does not reply within the 4 days the ticket is automatically closed.
> * Once a ticket is closed a customer can email back in and the ticket is reopened, the above starts counting again.
> 
> Hope this makes sense.

Makes perfect sense. Thank you for the detailed response.

Curious to know if an additional feature could be added, where a ticket will auto close in X days when a ticket is set to a specific status, regardless of interaction. 

### **TLDR:**

Can this be done?

```
if (ticketStatus == 'Resolved'){ 
    onReplyTimerReset = false
    autoClose = true };

else { 
    onReplyTimerReset = true
    autoClose = true };
```

Let's say you have 'Open', 'Closed' and 'Resolved' states.

When an agent chooses 'Closed', it immediately closes the ticket, and the user can't reply / reopen the ticket. Pretty straightforward.

But the 'Resolved' state basically keeps the ticket open for X amount of days, so the user can reply to the thread, maybe a "Thank you" post of some kind - but it also activates a timer to close the ticket automatically. Once the timer is up, it will automatically move the ticket from 'Resolved' to 'Closed' and the user can longer reply to the thread and has to generate a new inquiry. 

This setting wouldn't have the timer reset automatically upon reply (only if ticket status == 'Resolved') and would require the agent to manually set the ticket to 'Open' for the timer to be disabled, but instead acts as a 'Delay Closing Ticket' method. In our case, 99% of the time the user will have a "Thank you" response within 24hrs of a ticket. I want them to have the ability to post that reply, but I don't want the ticket to move out of my 'Resolved' queue and I don't want the timer to reset after one of these replies. 

Would something like this be possible as an option for this? I don't have much experience with PHP and work mainly with JS, but I would think this can be incorporated.

MarcBanyard:
> Curious to know if an additional feature could be added, where a ticket will auto close in X days when a ticket is set to a specific status, regardless of interaction.
> 
> ### **TLDR:**
> Can this be done?
> 
> ```
> if (ticketStatus == 'Resolved'){ 
>     onReplyTimerReset = false
>     autoClose = true };
> 
> else { 
>     onReplyTimerReset = true
>     autoClose = true };
> ```
> 
> Let's say you have 'Open', 'Closed' and 'Resolved' states.
> 
> But the 'Resolved' state basically keeps the ticket open for X amount of days, so the user can reply to the thread, maybe a "Thank you" post of some kind - but it also activates a timer to close the ticket automatically. Once the timer is up, it will automatically move the ticket from 'Resolved' to 'Closed' and the user can longer reply to the thread and has to generate a new inquiry.
> 
> This setting wouldn't have the timer reset automatically upon reply (only if ticket status == 'Resolved') and would require the agent to manually set the ticket to 'Open' for the timer to be disabled, but instead acts as a 'Delay Closing Ticket' method. In our case, 99% of the time the user will have a "Thank you" response within 24hrs of a ticket. I want them to have the ability to post that reply, but I don't want the ticket to move out of my 'Resolved' queue and I don't want the timer to reset after one of these replies.
> 
> Would something like this be possible as an option for this? I don't have much experience with PHP and work mainly with JS, but I would think this can be incorporated.

I Don't see why not, have a look at the files that were changed in the merge request and it should stand out, if you are able to read JS then you could have a tinker with PHP on a test site and see if that works for you.

You could duplicate the code from the files in my Merge Request and call it **Force Close Ticket After X Days**, there is a SQL query that will return the tickets that you could modify to suit your needs.


> When an agent chooses 'Closed', it immediately closes the ticket, and the user can't reply / reopen the ticket. Pretty straightforward.
> 
This is configurable by going to the **Admin Panel** > **Manage** > **Lists** > **Ticket Statuses** > **Closed** > **Item Properties** > Uncheck the **Allow tickets on this status to be reopened by end users** and click **Save**

ntozier:
You can already do something like this.

Create a closed status (let's call it: *Closed-No-Reopen* at **Manage -> Lists -> Ticket Statuses**
Under the *Item Properties* tab. set State Closed, and uncheck "Allow tickets on this status to be reopened by end users"

Run an autoclose plugin, and any tickets taht you do not want re-opened set them to the *Closed-No-Reopen*.
When an Agent closes a ticket if they don't want it to be re-openable they would set the ticket to *Closed-No-Reopen*.

New replies to tickets that are not allowed to re-open generate a new ticket.

need4swede:
> I Don't see why not, have a look at the files that were changed in the merge request and it should stand out, if you are able to read JS then you could have a tinker with PHP on a test site and see if that works for you.

Would something like this work?

"If ticket is set to 'Resolved' and has a status of 'open', automatically close ticket after X amount of days"

Screen Shot 2022-04-11 at 12 58 43

So now, regardless if an agent replies or not, as long as it's within the 'Resolved' queue, the ticket should auto-close after X days and still allow users to reply without resetting the timer.

EDIT: That didn't seem to work. Must've made an error somewhere.

kurtzelda:
The AutoClose ticket don't work on 1.16.x (think is not php 8 compliant).
I think should be very very very ... (a lot of "very") fine having this function inside the original project.

MarcBanyard:
The following Pull Request resolves the issue and works with PHP 8 and above
https://github.com/osTicket/osTicket/pull/6445

Hope this gets merged soon as it is extremely beneficial especially on busy systems by automatically closing tickets that customers have not replied to (i.e. the customers query has been resolved, so the ticket can be closed)

philbertphotos:
I like this may manually added to my build.

MarcBanyard:
> I like this may manually added to my build.

Please vote for this in #6445 as it would then be part of the core system.
Many people have asked for the ability over the years, so i've added it on several Pull Requests, but so far none have been merged in.

If many people like and need it there should be enough reason for the main devs to merge it into the core system.
Its been written to have no negateive impact to people who do not use it or do not want it, so merging it in would do no harm, but it would benefit everyone who does want and need it.

kiongchiu82:
I need this feature, will stay tuned.

asmith3006:
I would love to see this added!

Issue #6053, opened 2022-01-26
New ticket search causes DB error 1054


Peter1994:
Hi all,

Just started testing osTicket for our IT department. Everything has been working great so far except the following:

When I attempt to create a new ticket in the web interface and assign a creator to it, no results are shown:

![afbeelding](https://user-images.githubusercontent.com/1346743/151182581-b1dfb808-2f20-46c5-a4db-d1d2c6370484.png)

Instead a DB 1054 error occurs with the following output:

```
[SELECT A1.`id`,  A1.`name`,  A6.`address` FROM `ost_user` A1 LEFT JOIN `ost_user_email` A2 ON (A1.`id` = A2.`user_id`) LEFT JOIN `ost_organization` A3 ON (A1.`org_id` = A3.`id`) LEFT JOIN `ost_user_account` A4 ON (A1.`id` = A4.`user_id`) LEFT JOIN `ost_user__cdata` A5 ON (A1.`id` = A5.`user_id`) LEFT JOIN `ost_user_email` A6 ON (A1.`default_email_id` = A6.`id`) WHERE (A2.`address` LIKE '%peter%' OR A1.`name` LIKE '%peter%' OR A3.`name` LIKE '%peter%' OR A4.`username` LIKE '%peter%' OR A5.`phone` LIKE '%peter%') UNION ALL (SELECT A1.`id`,  A1.`name`,  A3.`address` FROM `ost_user` A1 LEFT JOIN `ost_user_email` A2 ON (A1.`id` = A2.`user_id`) LEFT JOIN `ost_user_email` A3 ON (A1.`default_email_id` = A3.`id`) WHERE A2.`address` IN ('Peter.<redacted>',  'peter.<redacted>',  'Peter.<redacted>',  'Peter.<redacted>',  'peter.<redacted>')) LIMIT 25] Unknown column 'A5.phone' in 'where clause'
 
 ---- Backtrace ----
 #0 (root)/include/mysqli.php(201): osTicket->logDBError('DB Error #1054',  '[SELECT A1.`id`...')
 #1 (root)/include/class.orm.php(3466): db_query('SELECT A1.`id`, ...',  true,  true)
 #2 (root)/include/class.orm.php(3522): MySqlExecutor->execute()
 #3 (root)/include/class.orm.php(2070): MySqlExecutor->getRow()
 #4 (root)/include/class.orm.php(2043): FlatArrayIterator->{closure}()
[...]

Issue #6043, opened 2022-01-03
New Ticket / User search / Unknown column 'A5.phone' in 'where clause'


mamo2021:
### Description

Error # 1054
Unknown column 'A5.phone' in 'where clause'

### Steps to Reproduce

Agent-Section:
1. Open new Ticket.
2. Enter something in Field User or Cc.

Expected: Display Searche-Results

### Versions

Admin panel -> Dashboard -> Information: 

Serverinformationen:
osTicket-Version | v1.15.4 (6bd7884)
Server-Software | Apache
MySQL-Version | 5.7.36
PHP-Version | 7.4.25

JediKev:
@mamo2021 

This is a known issue that is on our bugsheet. There are many posts on [the Forum](https://forum.osticket.com) about this but I will answer here as well.

Please add a new dummy field to the Contact Information Form (**Admin Panel > Manage > Forms > Contact Information**), save changes, delete the newly created dummy field, save changes again, and search for a User in the User Directory (**Agent Panel > Users > User Directory**). This should drop and recreate the User CDATA table with the appropriate information. The search for the User should show results, if not, wait a few minutes, refresh the page, and try again.

Cheers.

mamo2021:
That's it.
Thanks for the very fast answer!

PR #6054, opened 2022-01-27
issue: CDATA Rebuild Issue

JediKev:
This addresses issues #6043 and #6053 where installing osTicket and doing a User lookup will return a fatal error of `Unknown column 'A5.phone' in 'where clause'`. This is an indirect result of the changes we did to rebuild CDATA tables post-installation. The issue is that we are using `getInstance()` which returns a cached object. Changing this to always return a new instance via `getNewInstance()` subsequently had issues with the TicketForm not adding the Priority field. We determined this is due to object caching in the ORM so we added `ModelInstanceManager::flushCache()` before we rebuild the CDATA tables to clear the object cache and prevent the reuse of old objects.

### DIFF
--- a/include/class.dynamic_forms.php
+++ b/include/class.dynamic_forms.php
@@ -291,7 +291,15 @@ class DynamicForm extends VerySimpleModel {
         return true;
     }
 
+    static function getNewInstance() {
+        return static::objects()
+                ->one()
+                ->instanciate();
+    }
+
     static function rebuildDynamicDataViews() {
+        // Flush the object cache
+        ModelInstanceManager::flushCache();
         return self::ensureDynamicDataViews(true, true);
     }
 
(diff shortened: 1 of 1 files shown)


include/class.dynamic_forms.php
```diff
@@ -291,6 +291,11 @@ function saveTranslations($vars=false) {
         return true;
     }
 
+    static function getNewInstance() {
+        $o = static::objects()->one();
+        return $o->instanciate();
```
protich:
```php 
return static::objects()
           ->one()
           ->instanciate();
```

setup/inc/class.installer.php
```diff
@@ -318,6 +318,8 @@ function install($vars) {
             $user->setOrganization($org);
         }
 
+        // Flush the object cache
+        ModelInstanceManager::flushCache();
```
protich: Perhaps this needs to be in `DynamicForm::rebuildDynamicDataViews();` function? That way if it's called elsewhere it does the cache flush?

PR #6058, opened 2022-02-03
i18n: Missing translations for the action buttons

dsoms:
Fix missing translations for the action buttons.


JediKev (approved):
This looks good to me. Tested and working as it should. Thanks for the submission. 👍🏻

PR #6065, opened 2022-02-09
issue: Email Templates Static

JediKev:
This addresses an issue reported on the Forum where cloning a Template Set gives you a fatal error of `Non-static method EmailTemplateGroup::add() cannot be called statically`. This updates the `add()` method to static for `EmailTemplateGroup` and `EmailTemplate` classes.

### DIFF
--- a/include/class.banlist.php
+++ b/include/class.banlist.php
@@ -17,11 +17,11 @@
 require_once "class.filter.php";
 class Banlist {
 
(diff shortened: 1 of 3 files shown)


sdm216:
This issue also exists for adding an email to banlist. 
`Non-static method Banlist::add() cannot be called statically`

JediKev:
@sdm216 Added, thanks.

PR #6068, opened 2022-02-10
issue: Remaining Deprecated each()

JediKev:
This addresses an issue reported on the Forum where a couple of `each()` methods slipped through the cracks. This updates all the `each()` (minus the php_analyzer file for now) to the new PHP 8 standards.

### DIFF
--- a/include/pear/Net/DNS2.php
+++ b/include/pear/Net/DNS2.php
@@ -858,7 +858,8 @@ class Net_DNS2
             //
             if ($tcp_fallback == false) {
 
-                $ns = each($this->nameservers);
+                $ns = (empty($this->nameservers) || !is_array($this->nameservers))
+                        ? false : [key($this->nameservers), current($this->nameservers)];
                 if ($ns === false) {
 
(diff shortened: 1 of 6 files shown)


include/pear/Net/DNS2.php
```diff
@@ -858,7 +858,7 @@ protected function sendPacket(Net_DNS2_Packet $request, $use_tcp)
             //
             if ($tcp_fallback == false) {
 
-                $ns = each($this->nameservers);
+                $ns = [key($this->nameservers), current($this->nameservers)];
```
ManuelHu: I was just scrolling through this PR and saw this change that is not functionally equivalent to the previous version. The manually constructed array will never be false thus possible leading to a endless loop? At the end of the array it will be `[null,false]` which is not equivalent to false...

JediKev: Thank you for the input. We will review the change and get back to you.

medisoft (commented):
I applied the changes on my install and tested, and worked

Issue #6074, opened 2022-02-15
Error 500 when sending registration or pw reset email


medisoft:
### Prerequisites

* [yes ] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [yes] Do you have any errors in the PHP error log, or javascript console?
* [yes] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [yes] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/osTicket/osTicket/wiki/Github-Issues-Guidelines)

### Description

Many tasks under class.user.php trigger the error when ending an email

### Steps to Reproduce

1. Try to create a user. When sending the email, it will fail with 500 error (using SMTP)
2. On the agent panel, try to send the unlock email to the user, it will fail with 500 error

**Expected behavior:** [What you expected to happen]
I expect the emails to be sent

**Actual behavior:** [What actually happened]
Error 500 

### Versions

Admin panel -> Dashboard -> Information which also additionally gives you information about your server.

![image](https://user-images.githubusercontent.com/5624499/153968163-3d998eef-aba1-4a3e-90dc-dcefa2f3407d.png)

Also, please include the OS and what version of the OS you're running. As well as your browser and browser version.
[Tue Feb 15 00:05:03.386072 2022] [proxy_fcgi:error] [pid 100132] [client 2806:10be:8:3341:8252:71f3:3453:89b0:40990] AH01071: Got error 'PHP message: PHP Fatal error:  Uncaught Error: Using $this when not in object context in /var/www/tickets.domitai.com/include/class.user.php:1243\nStack trace:\n#0 /var/www/tickets.domitai.com/include/class.user.php(1215): UserAccount::sendUnlockEmail()\n#1 /var/www/tickets.domitai.com/pwreset.php(24): UserAccount->sendResetEmail()\n#2 {main}\n  thrown in /var/www/tickets.domitai.com/include/class.user.php on line 1243', referer: https://tickets.domitai.com/pwreset.php

JediKev:
@medisoft 

I can see the issue. A patch will be created and included in the next release. Thanks for the report.

Cheers.

JediKev:
@medisoft 

Here is the fix:
- https://github.com/osTicket/osTicket/pull/6075

As I mentioned this will be included in the next release.

Cheers.

PR #6075, opened 2022-02-15
issue: User sendUnlockEmail

JediKev:
This addresses issue #6074 where `sendUnlockEmail()` is static but cannot be as it uses `$this`. This updates the method to non-static and updates all calls to non-static calls.


medisoft:
Thanks. I applied the commit to my install, and it is now working :D

PR #6087, opened 2022-02-24
issue: Deprecated Required Parameter

JediKev:
This addresses a few issues reported on the Forum where some methods are throwing `Deprecated: required parameter follows optional parameter` errors. This updates the erroneous methods to match the PHP 8 standards using explicit nullable types with defaults.

### DIFF
--- a/include/class.config.php
+++ b/include/class.config.php
@@ -187,7 +187,7 @@ extends VerySimpleModel {
             ))->delete();
     }
 
-    static function getConfigsByNamespace($namespace=false, $key, $value=false) {
+    static function getConfigsByNamespace(?string $namespace=null, $key, $value=false) {
         $filter = array();
 
          $filter['key'] = $key;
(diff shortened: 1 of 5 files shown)


VincentSC:
Currently v1.16.1's crontab is broken under PHP8 due to this. I've tested the crontab-part (requires class.config.php) and it seems to fix the crontab-problem.

Rank1Mauritius:
Do you think a fix will be available before the new version is released ?

JediKev:
@Rank1Mauritius 

The issue is fixed with this pull. This will be included in the next release.

Cheers.

VincentSC:
I found more problems out of `scp` causing mostly warnings. Those are also picked up, or do you need a PR for these too?

JediKev:
@VincentSC 

Search through [my latest pulls](https://github.com/osTicket/osTicket/pulls/JediKev) (from #6071 up) and see if they've already been addressed. If not, create a new Issue Report.

Cheers.

PR #6095, opened 2022-03-14
Updated translations

pete1854:
Minor updates to swedish translations


JediKev:
@pete1854 

You would need to submit these to Imperavi directly (developers of Redactor). We simply download each latest version and update the files in our codebase. So, if we merge this the next time there is a Redactor update this will be overwritten.

Cheers.

PR #6101, opened 2022-03-17
issue: MaxLength Attribute

JediKev:
This addresses an issue reported on the Forum where Fields are not honoring the Length value. This is due to the label being set to `length` while the HTML attribute is named `maxlength`. This adds a case for `length` that sets the attribute name to `maxlength`.

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -4348,6 +4348,8 @@ class TextboxWidget extends Widget {
                     if ($v)
                         $attrs['data-translate-tag'] =  '"'.$v.'"';
                     break;
+                case 'length':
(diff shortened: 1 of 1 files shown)


include/class.forms.php
```diff
@@ -4349,7 +4349,9 @@ function render($options=array(), $extraConfig=false) {
                         $attrs['data-translate-tag'] =  '"'.$v.'"';
                     break;
                 case 'size':
+                case 'length':
                 case 'maxlength':
+                    $k = ($k === 'length') ? 'maxlength' : $k;
```
JediKev: move code to length case, move size below, set the key, and don't break

PR #6104, opened 2022-03-19
Fix malforming *text* thread entries by looking for unclosed HTML tags

rmilecki:
```
When dealing with *text* entries there is no point in checking them for
invalid HTML code (unclosed HTML tags) using Format::html_balance().
That could result in treating pure text as HTML and malforming parts
looking similar to HTML syntax.

This change *prevents* replacing fragments like:
"foo <bar@qux> ..."
with:
"foo <bar> ...</bar>"

Signed-off-by: Rafał Miłecki <rafal@milecki.pl>
```

### DIFF
--- a/include/class.thread.php
+++ b/include/class.thread.php
@@ -2840,7 +2840,7 @@ class TextThreadEntryBody extends ThreadEntryBody {
     }
 
     function getClean() {
-        return Format::htmlchars(Format::html_balance(Format::stripEmptyLines(parent::getClean())));
+        return Format::htmlchars(Format::stripEmptyLines(parent::getClean()));
     }
 
     function prepend($what) {


rmilecki:
I noticed this issue when my customer sent me a following simple **text** e-mail:
```
Dear support,

my discount code <cheap@shop> doesn't work.

Please help me.
```
### Before
Instead of showing original e-mail content osTicket displayed following malformed version which didn't allow me to fully understand my customer:

before

### After
After patching osTicket with with Pull Request change and sending identical e-mail everything works as expected:

after

JediKev:
@rmilecki 

We use `html_balance` for good reason (HTML sanitization). With your patch you’ll open yourself up to vulnerabilities like XSS, etc. which is why this pull will not be accepted.

We know the current sanitization can be frustrating at times which is why we are aiming to have better sanitization in the full code base rewrite (v2.0), so please stay tuned.

Cheers.

rmilecki:
@JediKev: please note I'm talking about **text** here. I still don't think `html_balance` / HTML sanitization makes sense in a context of **text** entries. Pure text still gets processed with `Format::htmlchars()` AKA `htmlspecialchars()`.

If user sends any **text** malicious message it will be displayed **as code** and nothing will get actually executed by a browser.

With this Pull Request applied locally I just sent following e-mail:
```
I'm EVIL. This e-mail is meant to bypass osTicket's security.

Look at this puppy:
<img src="/osticket/delete/user/5" />

Even worse attack:
<script>alert('pwned');</script>
```

It did **zero** harm:
not-so-evil

If you still think I'm wrong, can you provide an actual example how this change opens osTicket for any vulnerabilities?

rmilecki:
Just to clarify: this Pull Request **doesn't** modify `HtmlThreadEntryBody::getClean()`. That function remains unmodified and it still uses `Format::sanitize()`.

This Pull Requests affects `TextThreadEntryBody::getClean()` only.

PR #6105, opened 2022-03-20
Add basic tests checking ThreadEntryBody formatting

rmilecki:
```
Testing formatting for issues / regressions is important as it:
1. Takes care of valid HTML syntax to avoid breaking layout
2. Strips unsafe HTML code that could cause security issues

Signed-off-by: Rafał Miłecki <rafal@milecki.pl>
```

### DIFF
--- a/setup/test/tests/test.formatting.php
+++ b/setup/test/tests/test.formatting.php
@@ -1,0 +1,34 @@
+<?php
+require_once "class.test.php";
+require_once INCLUDE_DIR.'class.auth.php';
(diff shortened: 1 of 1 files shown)


rmilecki:
Those are some basic tests, they of course don't cover all `ThreadEntryBody` formatting, but should be a good start.

rmilecki:
@JediKev: could you take a look at this, please?

I think it's important to have some tests for HTML cleaning code.

PR #6106, opened 2022-03-21
Add translation for form buttons to edit.inc.php

empty88:
Add translation for form buttons to edit.inc.php


JediKev:
Please rebase this for `1.15.x` and it will be ready for merging.

empty88:
> Please rebase this for `1.15.x` and it will be ready for merging.

Sorry, but I don't know what to do. Where ist my mistake?

JediKev:
@empty88 

As you can see I switched the `base` from `develop` to `1.15.x` and it now includes a ton of extra commits rather than just your single commit. The reason this happened is because you made your `patch-4` branch from `develop` when it should be from `1.15.x`. So you need to rebase your `patch-4` branch on the `1.15.x` branch and update this pull.

If you don't know how I can close this one and make a new one.

Cheeers.

PR #6036, opened 2021-12-10
Plugins: Add Multi-Instance Support

protich:
This PR introduces multi-instance support to plugins along with other improvements.

The change provides flexibility on how plugins can be used. For example, a plugin like **Password Policies Plugin** can now have multiple instances making it possible to have different policies for Agents and Users. Authentication plugins (OAuth, SAML, LDAP) can now register different backends as needed.

Plugins list now shows number of instances - if any.

Screen Shot 2021-12-10 at 9 26 06 AM

On view the plugin show Plugin Information as well as ability to add internal notes.

Screen Shot 2021-12-10 at 9 26 37 AM

Instances tab shows list of instances along with mass actions (Enable, Disable and Delete).

Screen Shot 2021-12-10 at 9 26 55 AM

An instance can now be configured via a modal or inline - this opens up for plugin instances to be added or configured systemwide.

Screen Shot 2021-12-10 at 9 27 20 AM

PS: Goes without saying that the feature is NOT production ready yet. It requires database changes that might change before the feature is merged.

PR #6155, opened 2022-05-05
Fix: plugin config with multiple plugins

rinkp:
Due to the current static way of storing the config for all plugins, the plugin config is stored statically for all plugins. This means it is impossible to use multiple plugins with configuration. 

This pull requests fix this while still preventing multiple database calls for the same config.

### DIFF
--- a/include/class.plugin.php
+++ b/include/class.plugin.php
@@ -374,6 +374,7 @@ abstract class Plugin {
      * that inherits from PluginConfig. This is abstract and must be defined
      * by the plugin subclass.
      */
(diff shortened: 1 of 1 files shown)


JediKev:
@rinkp 

I believe you are essentially saying it's impossible to have multiple instances of the same plugin running with separate configurations, correct? If so, we are addressing this with what we call "Multi-Instance Plugins". You can see the pull here:
- https://github.com/osTicket/osTicket/pull/6036

Cheers.

PR #6163, opened 2022-05-09
Do not validate user that is blocked by admin

esamo:
Otherwise user that is blocked (but still logged in) can freely reply tickets but should not be allowed to.


JediKev:
@esamo 

Thank you for submitting this pull request. Upon investigating this turned out to be a bigger issue. We have rewritten this pull here:
- https://github.com/osTicket/osTicket/pull/6169

Cheers.

PR #6166, opened 2022-05-13
issue: Redactor Freezing

JediKev:
This addresses an issue where removing all text in Redactor in the Post Reply box freezes the entire page (using 90% or more CPU). As a quick workaround user jarnott from the Forum found that adding a `br` tag between the `draft_id` input and the Post Reply `textarea` fixes the issue. This adds the `br` tag to temporarily address the issue until Redactor addresses the issue permanently.

### DIFF
--- a/include/staff/ticket-view.inc.php
+++ b/include/staff/ticket-view.inc.php
@@ -1042,6 +1042,7 @@ if ($errors['err'] && isset($_POST['a'])) {
                         break;
                     } ?>
                     <input type="hidden" name="draft_id" value=""/>
+                    <br/>
                     <textarea name="response" id="response" cols="50"
                         data-signature-field="signature" data-dept-id="<?php echo $dept->getId(); ?>"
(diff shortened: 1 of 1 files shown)


Lineflyer:
Had the same reproducible issue with v1.15.7 (26ab7e9) using Google Chrome (no problem with Firefox) and manually applied this patch.

I can confirm this mitigates the issue of freezing (when using back key in response window) and hope it will be included into the next release.

ntozier:
This was reported to me by my agents also.
This patch fixed it on our production site 1.15.4.

mlipok:
as a reference this was discussed here:
https://forum.osticket.com/d/100957-edge-browser-freezes-when-editing-message-in-post-reply-text-box

Mike-Benoit:
Pretty sure this is actually a bug in Chrome with the spell checker, see: https://bugs.chromium.org/p/chromium/issues/detail?id=1337817

PR #6178, opened 2022-06-02
Fix: plugin management displays config edit form for different plugin

jasplin:
The getConfig method of the Plugin class incorrectly uses a static variable to cache the plugin config.
This means only the first edited plugin shows the correct edit form.

Switch to using a class member variable.

### DIFF
--- a/include/class.plugin.php
+++ b/include/class.plugin.php
@@ -375,6 +375,7 @@ abstract class Plugin {
      * by the plugin subclass.
      */
     var $config_class = null;
+    var $config = null;
     var $id;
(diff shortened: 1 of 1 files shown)


philbertphotos:
Will update the code to reflect this in a similar fashion

pablodgonzalez (approved):
I test it in 1.16.3 version.It works well

JediKev (approved):
This looks good. This wasn't a problem in older versions of PHP but since the breaking changes in PHP 8+ this is absolutely needed. 👍🏻

PR #6185, opened 2022-06-21
Add option to assign forms to tickets which are created by mail

gerundt:
This is a new version from my PR #5906 to fix one problem of the bug #5846. I know that you [don't want to allow this by default](https://forum.osticket.com/d/99863-custom-form-not-added-for-tickets-received-via-email/11). So i create a new mail option, which is off by default.


surfaceHH:
This is excactly what we need. Hopfully it will be included as soon as possible.

PR #6186, opened 2022-06-22
Update pwreset.php

gtech99:
user password resets wouldn't work because $id still contained the front letter in the subsequent lookup.


JediKev:
@gtech99 

This fix is legitimate; I have a small note for historical purposes. When looking at this the only way this will be an issue is if a ClientAccount cannot be found for the User and it goes to the `elseif` where it attempts to lookup the User by ID. However, this `elseif` is not functional as we don't actually have a `pwreset.create.php` file.

Cheers.

PR #6187, opened 2022-06-23
Update class.validator.php

parijke:
Allow email adresses as the usernames for agents


parijke:
Sorry, stumbled upon this one, did not know that

PR #6189, opened 2022-06-24
Fix truncation of messages containing encoded HTML entities which will decode in invalid HTML tags (e.g. <)

agerasch:
This commit will disable the decoding of HTML entities (e.g. `&lt;` and `&gt;`) when come along from Format::sanitize function. This will fix the problem with < and > symbols truncating the remaining message within the html_balance/strip function.

Maybe, there is a good reason for this, but I can't figure it out, so please help me :-)

### DIFF
--- a/include/class.format.php
+++ b/include/class.format.php
@@ -358,7 +358,7 @@ class Format {
     static function sanitize($text, $striptags=false, $spec=false) {
 
         //balance and neutralize unsafe tags.
-        $text = Format::safe_html($text, array('spec' => $spec));
+        $text = Format::safe_html($text, array('spec' => $spec, 'decode' => false));
 
         $text = self::localizeInlineImages($text);
 


JediKev:
@agerasch 

This is not safe, this will potentially open you up to a slew of attacks including XSS.

Cheers.

ChrisTG742:
> @agerasch
> 
> This is not safe, this will potentially open you up to a slew of attacks including XSS.
> 
> Cheers.

Are you sure it will? I rather think decoding the escaped "greater" and "less" actually makes them executable for the HTML-renderer in the browser. That's what osTicket actually does (at least for these two chars).
So lettings all specials chars encoded or escaped should be more safe then decoding it before the HTML-stream is sent to the browser, no?

moving-bits:
This might be related to a recent problem we had on support: Some user sent us an encrypted file and password (using different channels). The password started with a less sign, so something like "The password is <abced".
The email was sent in HTML format, and osTicket only displayed the string "The password is " - everything after the initial "<" of the password got truncated, even additional text lines.

Lineflyer:
I can confirm this problem with v1.16.3, not sure whether it was introduced with v1.16 or was already present in previous versions.
It however causes significant problems in communication with clients.

The problem is, that mails are truncated after a `<` sign in the clients mail text.

ChrisTG742:
Isn't that the same problem as described in PR #6104 ?

PR #6205, opened 2022-07-20
update: More Lookup Improvements

JediKev:
This adds a few more lookup improvements. This adds a new Validator called `is_alnumeric()` that will check if the provided string is in fact alphanumeric. This also updates an if statement in `TicketsAjaxAPI::lookup()` to add a check for `is_alnumeric()` to make sure we are only looking up Ticket numbers with `lookupByNumber()`.

### DIFF
--- a/include/ajax.tickets.php
+++ b/include/ajax.tickets.php
@@ -55,8 +55,9 @@ class TicketsAjaxAPI extends AjaxController {
         global $ost;
         $hits = $ost->searcher->find($q, $hits, false);
 
-        if (preg_match('/\d{2,}[^*]/', $q, $T = array())) {
-            $hits = $this->lookupByNumber($limit, $visibility, $hits);
+        if (Validator::is_alnumeric($q)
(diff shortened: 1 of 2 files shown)


include/ajax.tickets.php
```diff
@@ -55,7 +55,8 @@ function lookup() {
         global $ost;
         $hits = $ost->searcher->find($q, $hits, false);
 
-        if (preg_match('/\d{2,}[^*]/', $q, $T = array())) {
+        if (Validator::is_alnumeric($q)
+                && preg_match('/\d{2,}[^*]/', $q, $T = array())) {
             $hits = $this->lookupByNumber($limit, $visibility, $hits);
```
protich: Remove `$hits` from lookupByNumber.

protich (changes requested):

include/ajax.tickets.php
JediKev: Done

protich (approved):

PR #6215, opened 2022-08-02
Update Dockerfile to run crond

pdeffendol:
In order to fetch email regularly the cron task needs to be enabled.


pdeffendol:
Not sure what happened here.  Wrong repo.

PR #6216, opened 2022-08-03
mailparse: thread_entry_recipients to

JediKev:
This addresses an issue where some emails don't have a To header which causes a fatal error when parsing mail. This is due to assuming To is set and passing it to array_unique which expects the value to be an array. This adds a check to see if To is set and if so we call array_unique otherwise we default to an empty array.

### DIFF
--- a/include/class.mailparse.php
+++ b/include/class.mailparse.php
@@ -691,7 +691,9 @@ class EmailDataParser {
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


protich:
I think this should be made to 1.15.x branch.

JediKev:
Updated to 1.16.x 👍🏻

protich (approved):

PR #6232, opened 2022-08-23
Catch storage error when mailing ticket response

jasplin:
**tl;dr**
Unhandled error fetching inline images when replying to a ticket means email is not sent but the ticket thread is updated so that it looks like it is.

This pull request adds a try/catch around the getData() call when adding inline html images and allows the email to continue. A warning message is added to the osticket system log.

**Details**
I'm using the s3 storage plugin and I managed to mess up by setting up lifecycle rules that move s3 objects into Glacier long term storage. Hitting one of the moved objects (especially around email footer logos) causes an unhandled exception.

### DIFF
--- a/include/class.mailer.php
+++ b/include/class.mailer.php
@@ -516,11 +516,19 @@ class Mailer {
                         $file = AttachmentFile::lookup($match[1]);
                     if (!$file)
                         return $match[0];
-                    $mime->addHTMLImage($file->getData(),
-                        $file->getMimeType(), $file->getName(), false,
-                        $match[1].$domain);
-                    // Don't re-attach the image below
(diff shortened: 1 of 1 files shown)


protich (approved):

protich:
@jasplin - great find. Thank you for the PR. We now expect 10 more from you! /jk.

jasplin:
> @jasplin - great find. Thank you for the PR. We now expect 10 more from you! /jk.

thanks @protich .. in fact FYI there is already another that has been reviewed and is ready to merge -> #6178 

very likely to be more on their way :)

JediKev:
@jasplin 

Yea that’s going to be merged soon and included in the next set of releases. 👍🏻 

Cheers.

PR #6234, opened 2022-08-25
plugins: Audit enable()

JediKev:
This addresses an issue where the Audit Plugin relies on `enable()` being called in order to create the `_audit` table required for saving audit data to the database. This adds the method back temporarily until we can rewrite the plugin to fit the new standards. This also attempts to call `enable()` on install so that the table can be created. If the plugin doesn't have an `enable()` method it defaults to the parent class Plugin `enable()` which just returns `true`.

### DIFF
--- a/include/class.plugin.php
+++ b/include/class.plugin.php
@@ -366,6 +366,8 @@ class PluginManager {
         ];
         $p = Plugin::create($vars);
         if ($p->save(true)) {
+            // TODO: Trigger enable() for the plugin (for now)
+            $p->getImpl()->enable();
             static::clearCache();
             return $p;
         }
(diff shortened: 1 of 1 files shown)


include/class.plugin.php
```diff
@@ -366,6 +366,8 @@ function install($path) {
         ];
         $p = Plugin::create($vars);
         if ($p->save(true)) {
+            // Trigger enable() for the plugin
```
protich: Can you please put a TODO note here to indicate it's temporary?

JediKev: Sure thing

protich (approved):

PR #6239, opened 2022-08-26
issue: External Auth Regenerate Session

JediKev:
This addresses an issue where we are rotating the session on each login to help prevent session fixation. However, this causes an issue when using an External Authentication Backend such as OAuth2. This remove the `regenerateSession()` calls from the `login()` methods and instead we will set a time bomb that will auto regenerate the session a certain amount of time after login. This way external auth backends will still function properly and we can still regenerate the session.


JediKev:
Rebased for `develop`

protich (approved):

PR #6241, opened 2022-08-31
php: 8.1 Support

JediKev:
This addresses Warnings and Fatal Errors with PHP 8.1.x. This includes:
- Fix Deprecation Warnings (Serializable)
- Fix Deprecation Warnings (line_endings)
- Fix Upgrader Issues

### DIFF
--- a/include/class.ostsession.php
+++ b/include/class.ostsession.php
@@ -71,7 +71,7 @@ class osTicketSession {
             $this->backend = new self::$backends['db']($this->ttl);
         }
 
-        if ($this->backend instanceof SessionBackend) {
+        if (!empty($this->id) && ($this->backend instanceof SessionBackend)) {
             // Set handlers.
             session_set_save_handler(
                 array($this->backend, 'open'),
--- a/include/class.page.php
+++ b/include/class.page.php
@@ -234,10 +234,7 @@ class Page extends VerySimpleModel {
         try {
             return self::objects()->filter(array('type'=>$type))->one();
         }
-        catch (DoesNotExist $ex) {
-            return null;
-        }
-        catch (InconsistentModelException $ex) {
+        catch (DoesNotExist | InconsistentModelException $ex) {
             return null;
         }
     }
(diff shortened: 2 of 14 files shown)


JediKev:
**TODO:**
- [x] Fix remaining warnings with `(un)serialize`.
- [ ] Maybe address `strftime()` warnings?

include/class.config.php
```diff
@@ -159,7 +159,7 @@ static function items($namespace, $column='namespace') {
         try {
             count($items);
         }
-        catch (InconsistentModelException $ex) {
+        catch (InconsistentModelException|mysqli_sql_exception $ex) {
```
protich: We shouldn't be catching "naked" mysql exceptions here - it needs to go to ORM and rethrown as `InconsistentModelException` expected here.

include/class.ostsession.php
```diff
@@ -40,7 +40,7 @@ function __construct($ttl=0){
         ini_set('session.gc_maxlifetime', $ttl);
 
         // Skip db version check if version is later than 1.7
-        if (!defined('MAJOR_VERSION') && OsticketConfig::getDBVersion())
+        if (OsticketConfig::getDBVersion())
```
protich: Why this change? Removing this will results in every page load checking DB version - which results in an SQL error and unnecessary query for installation after `v1.7`.

include/class.page.php
```diff
@@ -234,10 +234,7 @@ static function lookupByType($type, $lang=false) {
         try {
             return self::objects()->filter(array('type'=>$type))->one();
         }
-        catch (DoesNotExist $ex) {
-            return null;
-        }
-        catch (InconsistentModelException $ex) {
+        catch (DoesNotExist|InconsistentModelException|mysqli_sql_exception $ex) {
```
protich: See my other comment on mysql exceptions.

protich (changes requested):

JediKev:
@protich Updated to reflect the spacing in the caught exception types (`include/class.page.php`) 👍🏻

protich (approved):

PR #6243, opened 2022-09-01
Issue: Installer YAML

aydreeihn:
This commit fixes an issue where some default data was not being inserted into the database for fresh installs:

1. Sales Department
	The Sales Department was not being created because we were trying to set an sla_id before the SLA table was created.

2. Banlist
	The default Filter Rule and Action for the System Banlist was not being created because the save function now calls the ORM save function, so it is no longer calling save_rules and save_actions for the installer. We also have to change the YAML to refer to the fields as w,h,v instead of what, how, value for them to be recognized correctly.

It also fixes some spacing issues in the Filter class.

### DIFF
--- a/include/class.filter.php
+++ b/include/class.filter.php
@@ -502,36 +502,34 @@ extends VerySimpleModel {
 
             if (is_array($rule)) {
                 if($rule["w"] || $rule["h"]) {
-                // Check for REGEX compile errors
-                if (in_array($rule["h"], array('match','not_match'))) {
-                    $wrapped = "/".$rule["v"]."/iu";
-                    if (false === @preg_match($rule["v"], ' ')
-                            && (false !== @preg_match($wrapped, ' ')))
-                        $rule["v"] = $wrapped;
+                    // Check for REGEX compile errors
+                    if (in_array($rule["h"], array('match','not_match'))) {
+                        $wrapped = "/".$rule["v"]."/iu";
+                        if (false === @preg_match($rule["v"], ' ')
+                                && (false !== @preg_match($wrapped, ' ')))
+                            $rule["v"] = $wrapped;
+                    }
+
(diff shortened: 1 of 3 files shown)


include/class.filter.php
```diff
@@ -502,36 +502,34 @@ function validate_rules($vars,&$errors) {
...
             if (is_array($rule)) {
                 if($rule["w"] || $rule["h"]) {
-                // Check for REGEX compile errors
-                if (in_array($rule["h"], array('match','not_match'))) {
-                    $wrapped = "/".$rule["v"]."/iu";
-                    if (false === @preg_match($rule["v"], ' ')
-                            && (false !== @preg_match($wrapped, ' ')))
-                        $rule["v"] = $wrapped;
-                }
+                    // Check for REGEX compile errors
```
JediKev, line 505: looks like this has some weird spacing 🤔

JediKev: nevermind it's due to the spacing changes...DUH lol

include/class.filter.php
```diff
@@ -566,6 +564,16 @@ function save($refetch=false) {
         return parent::save($refetch || $this->dirty);
     }
 
+    static function __create($vars,&$errors) {
+        $filter = new static($vars);
+        if ($filter->save()) {
+            $filter->save_rules($vars,$errors);
```
protich:
```suggestion
            $filter->save_rules($vars, $errors);
```

include/class.filter.php
```diff
@@ -566,6 +564,16 @@ function save($refetch=false) {
...
+            $filter->save_actions($filter->getId(), $vars, $errors);
+
```
protich: Kills extra empty lines.
```suggestion

```

protich (approved):

PR #6253, opened 2022-09-19
Update class.mail.php: Wrong expected status code in SMTP protocol exchange handler

ChrisTG742:
Fix for typo in function auth(): expect need to be set to 334 instead of 234 here.


ChrisTG742:
See: https://learn.microsoft.com/en-us/exchange/client-developer/legacy-protocols/how-to-authenticate-an-imap-pop-smtp-application-by-using-oauth

protich:
@ChrisTG742 - Thank you for the PR, I referenced  the same document and even have comments above the routine, but somehowe typed it wrong!

ChrisTG742:
you're welcome :)

PR #6258, opened 2022-09-21
issue: Session Cookie Refresh

JediKev:
This addresses an issue where we were not renewing the cookie expired time correctly when refreshing the session. This updates `refreshSession()` to accept a `$refreshRate` defaulting to 60 (seconds). This also updates the `if()` statement from a negative to a postive logic for better readability.

### DIFF
--- a/include/class.usersession.php
+++ b/include/class.usersession.php
@@ -160,24 +160,23 @@ trait UserSessionTrait {
     // User class
     var $class = '';
 
-    function refreshSession($force=false) {
+    function refreshSession($refreshRate=60): void {
         // If TIME_BOMB isset and less than the current time we need to regenerate
         // session to help mitigate session fixation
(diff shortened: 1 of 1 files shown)


include/class.usersession.php
```diff
@@ -160,24 +160,23 @@ trait UserSessionTrait {
...
+            $this->token = $this->getSessionToken();
+            osTicketSession::renewCookie($updated, $this->maxidletime);
+        }
     }
 
     function regenerateSession($destroy=false) {
         $this->session->regenerateSession($destroy);
         // Set cookie for the new session id.
-        $this->refreshSession(true);
+        $this->refreshSession(0);
```
JediKev: `-1`

JediKev:
`refreshSession(0)` updated to `refreshSession(-1)`. Lint tests and normal tests pass. 👍🏻

PR #6259, opened 2022-09-22
plugins: Make Plugins Upgradeable Again!

protich:
This adds ability and interface to auto-trigger plugins upgrade or downgrade when there's a version mismatch on-load.

The base plugin class simply updates the version (for now) - the downstream Plugins should implement the actual upgrade / downgrade process as needed, especially if it creates database tables or needs to change stored config info to match the new version format.

### DIFF
--- a/include/class.plugin.php
+++ b/include/class.plugin.php
@@ -576,9 +576,11 @@ class Plugin extends VerySimpleModel {
     function __onload() {
         $this->info = PluginManager::getInfoForPath(INCLUDE_DIR.$this->ht['install_path'],
             $this->isPhar());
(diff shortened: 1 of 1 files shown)


ChrisTG742:
Full downgrade support on application level can be a hassle in my experience. Just think about datatypes that may have been dropped/changed or combined fields? How would you deal with that error-proven?
What about just saving something like a "snapshot" of the former settings/db tables when upgrading a plugin? By doing so you could provide a generic "roll back" feature for all plugins without the need to provide code for this in every plugin separately?
You could even let the user decide if he wants to restore the data/settings, or start with defaults again.
Just my two cents.

PR #6269, opened 2022-09-26
Show Topics including parents in dropdown list

empty88:
Child topics with the same name can not be identified without parent topic info


empty88:
Well, I just saw that the parent names are already included if an agent has "ability to see all departments" right. But I do not understand the connection, because even if an agent does not have this right, topics assigned to departments the agent has no access to are listed though, but ALL parents are invisible including parents the agent HAS access to.
A bit too confusing for me...

PR #6274, opened 2022-09-27
Fix to prevent dispatcher from including an already included file.

msigley:
This prevents fatal php errors. The error conditions to cause this can occur when you have lots of plugins installed as plugin developers kind of do what they what with no official plugin guidelines available.

In my case, the Ticket API broke after installing a plugin.


JediKev:
@msigley 

Please see the below that was merged into 1.17.x:
- https://github.com/osTicket/osTicket/pull/6245

Cheers.

msigley:
Thanks @JediKev. I tried to find an existing issue before I made this, but obviously your search spells are more powerful than mine. 😜

Ill close this.

PR #6277, opened 2022-09-27
Issue: File Storage Plugins Blacklist

aydreeihn:
This commit will temporarily add the file storage plugins to the multi instance blacklist until we update them to offer multi instance support.


protich:
Reviewed-By: Peter Rotich <peter@osticket.com>

PR #6282, opened 2022-09-28
issue: imagesx Boolean

JediKev:
This addresses an issue where if `imagecreatefromstring()` fails it returns false (boolean) and we then pass that to `imagesx()`. This method only accepts Gdimage (since PHP 8) so this adds a check for `$image` and if false we will default to `1` otherwise the image will be passed to `imagesx()` as normal.

### DIFF
--- a/include/class.file.php
+++ b/include/class.file.php
@@ -160,8 +160,8 @@ class AttachmentFile extends VerySimpleModel {
     function display($scale=false, $ttl=86400) {
         $this->makeCacheable($ttl);
 
-        if ($scale && extension_loaded('gd')) {
-            $image = imagecreatefromstring($this->getData());
+        if ($scale && extension_loaded('gd')
+                && ($image = imagecreatefromstring($this->getData()))) {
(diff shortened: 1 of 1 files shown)


include/class.file.php
```diff
@@ -162,7 +162,7 @@ function display($scale=false, $ttl=86400) {
 
         if ($scale && extension_loaded('gd')) {
             $image = imagecreatefromstring($this->getData());
-            $width = imagesx($image);
+            $width = $image ? imagesx($image) : 1;
```
protich: I think the fix here is incomplete - `imgesy` (line 167) would still expect an image. I think we should skip scalling  if  `imagecreatefromstring` fails.

```php

if ($scale  && extension_loaded('gd') 
          && ($image = imagecreatefromstring($this->getData()))) {

protich (changes requested):

JediKev:
update done

protich (approved):

PR #6285, opened 2022-09-28
file: Possible Key Collision

protich:
This  is a temporary fix for possible Fatal Crash when file key (hash lookup) matches more than one object. See TODO note for more information on future fixes.


protich:
It might make sense to cherry-pick the commit for v1.16.x or rebase the PR to it.

Issue #6291, opened 2022-10-03
Import from CSV via cli crashes


nebulade:
This seems to have started from version 1.17 on.

Our Cloudron app package tests have a case where user records are imported via the cli from a csv file. The agent import as such works, however the cli run crashes since a filedescriptor for the csv file gets `fclose`d twice.

First in the actual importer: https://github.com/osTicket/osTicket/blob/develop/include/class.import.php#L55
Then in the cli handling the action: https://github.com/osTicket/osTicket/blob/develop/include/cli/modules/agent.php#L154

The second gets an already closed stream handle. If there is a chance to get a PR upstreamed, I am happy to try a fix for this, but I am no osTicket internals expert nor php.

JediKev:
@nebulade 

Thank you for the report. This is a very simple fix actually. So, as you mentioned, the __destructor for the CsvImporter already closes the stream so when we get to fclose in the cli method the stream is no longer valid and causes the error. So all we need is a check to see if the stream is a resource and if so then we can fclose it otherwise we just continue (as it's already closed).

Here is the pull to address the issue:
- https://github.com/osTicket/osTicket/pull/6292

Cheers.

PR #6292, opened 2022-10-03
cli: User Import

JediKev:
This addresses issue #6291 where importing users from CSV via command line causes a fatal error. This is due to attempting `fclose()` on a stream that has already been closed. The CsvImporter `__descontruct()` method already closes the stream so when we get to the end of the import and attempt to call `fclose()` in the CLI method, it causes a fatal error as the stream (resource) is no longer valid (as it's already been closed). This adds a check to see if the stream is a valid resource and if so then we close it, otherwise we skip and continue (as it's already closed).

### DIFF
--- a/include/cli/modules/user.php
+++ b/include/cli/modules/user.php
@@ -168,7 +168,8 @@ class UserManager extends Module {
         default:
             $this->stderr->write('Unknown action!');
         }
-        @fclose($this->stream);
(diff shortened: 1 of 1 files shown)


nebulade:
Thanks for the quick reaction, without knowing the `is_resource()` function, if it does what it indicates by name, then looks good.

PR #6296, opened 2022-10-05
mailer: Make Headers Valid Again

protich:
**Message-Id**: osTicket uses encoded message Id to thread emails on reply. Laminas Mail, the new backend we're mangles up the Message-Id when set as a generic header. This commits corrects the issue by using MessageId object - making sure it is ASCII encoded.

**Date**: Date is auto-added upstream - so we're removing it from Mailer to avoid double header issues - but if need be, you can overwrite it by setting your own fake date!

**Return-Path**: Adds Return-Path Header interface to osTicket\Mail\Message

**References**: Use References Header interface to add references instead of using Generic Headers which might break due encoding.

**In-Reply-To**:  Add In-Reply-To Header interface

**Sender**: Add Sender Header interface

### DIFF
--- a/include/class.mail.php
+++ b/include/class.mail.php
@@ -187,6 +187,20 @@ namespace osTicket\Mail {
             }
         }
 
+        // Set Message Sender is useful for SendMail transport, its basically -f
+        // parameter in the mail() interface
+        public function setSender($email, $name=null) {
+            try {
+                // Exception is thrown on invalid email address
+                $header = new Header\Sender();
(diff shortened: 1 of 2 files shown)


protich:
I believe that's all the Headers we needed to handle at the moment... need extra reeding? Go read [RFC2822](https://www.ietf.org/rfc/rfc2822.txt) to learn more about headers you never knew existed.

**PS**: We should consider adding `Reply-To` header - perhaps an outgoing email with a spoofed header should have the original email as Reply-To field?

Issue #6295, opened 2022-10-04
Duplicate Date header in SMTP


coolacid:
### Prerequisites

* [X] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [ ] Do you have any errors in the PHP error log, or javascript console?
* [ ] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [X] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/osTicket/osTicket/wiki/Github-Issues-Guidelines)

### Description

When sending a SMTP diagnostic email to a gmail address, the ticket bounces with a duplicate header error. Inspecting the headers in the resulting email shows two instances of the `Date` header. 

### Steps to Reproduce

1. Send SMTP mail
2. View email headers
3. See duplicate `Date` header

**Expected behavior:** 

Email to be sent

**Actual behavior:** 

Email bounces when sent to strict SMTP servers (gmail). Fine when servers are not RFC strict. 

### Versions

OSTicket | v1.17 (1d8b790)
Web Server Software | Apache
MySQL Version | 5.7.39
PHP Version | 8.1.10

** Quick Fix: ** 

It appears as though the $$$$ Library adds the Date to the headers, when using SMTP. It would only be required when using sendmail transport.

JediKev:
@coolacid 

Thanks for the report. This is already something we are actively looking into.

Cheers.

coolacid:
Oh wow, thanks for the quick response. I tired to search issues/pr's but didn't see it.

JediKev:
@coolacid 

No problem (I live here and on the forum lol). This was something not reported here, rather the forum. But always good to have an "official" issue report here to track.

Cheers.

toxpal:
Before the issue gets resolved in new version, what file is responsible for email headers? I'll try to fix it myself as a temporary issue because all our emails sent from OSTicket to Gmail and other major email providers are rejected as spam because of duplicated headers.

JediKev:
@toxpal 

Please do a recursive search. The patches have already been made:
- https://github.com/osTicket/osTicket/pull/6296

You can either apply them manually or wait for the next release.

Cheers.

protich:
Alternatively, you can upgrade to the [tip of v1.17.x branch](https://github.com/osTicket/osTicket/tree/1.17.x) - which has latest fixes, including mailer issues.

[...]

PR #6297, opened 2022-10-05
POP3 Configration can not be saved

JensEB:
First error: abstract functions can not be private. It have to be protected or public...

Second error: the pop3 login function doesn't respond anything on successful login. This will be return false in MailBoxProtocolTrait::basicAuth()
So the settings can't be saved although the data is correct.

### DIFF
--- a/include/class.mail.php
+++ b/include/class.mail.php
@@ -331,7 +331,7 @@ namespace osTicket\Mail {
                 parent::login($user, $password, $tryApop);
                 return true;
              } catch (\Throwable $e) {
-                 throw new Exception(__('cannot login, user or password wrong'));
+                 throw new Exception(__('login failed').': '.$e->getMessage());
              }
          }
     }
--- a/include/class.mail.php
+++ b/include/class.mail.php
@@ -235,7 +235,7 @@ namespace osTicket\Mail {
         }
 
         abstract public function __construct($accountSetting);
-        abstract private function oauth2Auth(AccessToken $token);
+        abstract protected function oauth2Auth(AccessToken $token);
     }
 
     class ImapMailboxProtocol extends ImapProtocol {
@@ -326,6 +326,14 @@ namespace osTicket\Mail {
              return trim($result);
          }
 
+         public function login($user, $password, $tryApop = true) {
+             try {
+                parent::login($user, $password, $tryApop);
+                return true;
+             } catch (\Throwable $e) {
+                 throw new Exception(__('cannot login, user or password wrong'));
+             }
+         }
     }
 
     // MailBoxStorageTrait


ChrisTG742:
Could it be that you're trying to run osTicket >= 1.16 on PHP7.x ?

JensEB:
No, I use PHP 8.1
If I insert the wrong credentials, I get an "last request failed" error and if I insert the right credentials, I get an error "cannot login, user or password wrong".
The login function of the Pop3 class of laminas-mail don't have an return. So MailBoxProtocolTrait::basicAuth() returns nothing and the init function will interprete this as false...

include/class.mail.php
```diff
@@ -326,6 +326,14 @@ public function readLine() {
              return trim($result);
          }
 
+         public function login($user, $password, $tryApop = true) {
+             try {
+                parent::login($user, $password, $tryApop);
+                return true;
+             } catch (\Throwable $e) {
+                 throw new Exception(__('cannot login, user or password wrong'));
```
protich: Thanks for the find & fix 

I think we should  sarface whatever the error the mail server / backend kicked back. 

```php 
throw new Exception($e->getMessage());
```

Also curious why you're using POP instead IMAP.  Not judging ;)

protich (changes requested):

include/class.mail.php
JensEB: Yes you are right. I changed the exception message including the responded error...
I don't like POP, but a client of me need it. He has an Office365 mail without IMAP enabled. He says, it's a cheaper license.
But there is also the possibility to use POP, then that should work too.

protich (approved):

include/class.mail.php
```diff
@@ -326,6 +326,14 @@ public function readLine() {
              return trim($result);
          }
 
+         public function login($user, $password, $tryApop = true) {
+             try {
+                parent::login($user, $password, $tryApop);
+                return true;
+             } catch (\Throwable $e) {
+                 throw new Exception(__('login failed').': '.$e->getMessage());
```
JediKev, line 334: update this later to just `$e->getMessage()`

McCoyCOREUSA:
Same problem. This has been the most difficult upgrade for osTickets ever.

JediKev:
@McCoyCOREUSA 

If you get the "cannot login, user or password wrong" error then it sounds like Basic Authentication is disabled for your tenant. If this is the case you’ll need to upgrade [to 1.17.x](https://github.com/osTicket/osTicket/issues/6295#issuecomment-1272563459), [install the OAuth2 plugin](https://docs.osticket.com/en/latest/Guides/OAuth2%20Guide.html#setting-up-the-plugin), and [configure Modern Authentication](https://docs.osticket.com/en/latest/Guides/OAuth2%20Guide.html#authorization).

Cheers.

include/class.mail.php
JensEB: I wouldn't update this. $e-getMessage() returns only "last request failed". 
A normal User don't know, what this meens. So he know, It happens on login...

McCoyCOREUSA:
Upgrading to 1.17 and the other plugins is what we are trying to do, and why we are getting this error.

JensEB:
@McCoyCOREUSA  Yes, you have right. Say thanks to Microsoft and Google for oAuth2 requirement. ;-)
Did you make the changes of this pull request? Then you should be able to log in. Otherwise you have to waiting for the next release. It's already merged...

JensEB:
Can you describe in more detail what you want to do?
This pull request solves only connecting to an mail server by using basic auth and POP3...

McCoyCOREUSA:
I really don't know. We've had problem after problem with this upgrade. I have a guy doing it for me (I'm not a server admin). At first we were missing PHP 8+. Then that didn't work. Then we needed another plugin upgrade with osTickets, so we selected all plugins other than the Amazon specific ones. We just selected all to make sure we had what we needed. Now we are stuck again. 

Upgrading osTicket with my guy has always been a 45 minute task. So far, it's three hours over a week and I've been without a ticket system for my business for that entire time.

McCoyCOREUSA:
Now we get this...

Fatal error: Declaration of FsStoragePluginConfig::pre_save($config, &$errors) must be compatible with PluginConfig::pre_save(&$config, &$errors) in phar:///home/mccoysupportcom/domains/mccoysupport.com/public_html/ticketsystem/include/plugins/storage-fs.phar/storage.php on line 107

I think it's time to find a better solution than osTicket!

JediKev:
@McCoyCOREUSA 

That error occurs when you don't update your Filesystem Plugin. It’s not osTicket's fault nor problem; this falls on the responsibility of the person running the upgrade to make sure they do everything correctly.

Cheers.

McCoyCOREUSA:
We did. That's what we don't understand?

JediKev:
@McCoyCOREUSA 

Well maybe our downloads packager is serving an older version of the FS plugin with the specific combination you chose (which I doubt but not going to rule out here). I would recommend downloading solely the Filesystem plugin for 1.17.x version, replace your current plugin, and retest. You may also need to restart PHP-FPM if you’re running it.

Cheers.

McCoyCOREUSA:
okay will try that

McCoyCOREUSA:
This worked, to a point. Still had tons of errors which my server admin had to manually fix. Lots of problems with this version!

PR #6301, opened 2022-10-08
mailer: Improvements & Enhancements

protich:
**Spoofing**: When an SMTP Account **DOES NOT** allow spoofing then set the Sender (Envelope FROM) to the account's email address and leave the Header FROM as set.

**Originator**: When an SMTP account send fails - we now reset the Originator / Sender to the original From Address, incase spoofing setting changed the Sender.

### DIFF
--- a/include/class.mailer.php
+++ b/include/class.mailer.php
@@ -604,7 +604,7 @@ class Mailer {
                 // Attempt to send the Message.
                 if (($smtp=$smtpAccount->getSmtpConnection())
                         && $smtp->sendMessage($message))
-                     return $messageId;
+                     return $message->getId();
             } catch (\Exception $ex) {
                 // Log the SMTP error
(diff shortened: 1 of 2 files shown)


include/class.mailer.php
```diff
@@ -585,46 +585,60 @@ function($match) use ($domain, $message, $self) {
...
-                        $account->getHostInfo(),
+                        $smtpAccount->getEmail()->getEmail(),
+                        $smtpAccount->getHostInfo(),
                         $ex->getMessage()
                     ));
             }
-            // Reset FROM address
-            $message->setFrom($this->getFromEmail(), $this->getFromName());
+            // Attempt  Failed:  Reset FROM to original email and clear Sender
+            $message->setOrignator($this->getFromEmail(), $this->getFromName());
```
JediKev: Typo: `setOrignator` -> `setOriginator`

JediKev:
Tested and works (minus the typo)

protich:
Typo fixed.

Issue #3366, opened 2016-09-17
SMTP mail does not set domain name for HELO message


nevesnunes:
The SMTP implementation of the PEAR Mail interface can receive the following parameter:

```
    /**
     * Hostname or domain that will be sent to the remote SMTP server in the
     * HELO / EHLO message.
     *
     * @var string
     */
    var $localhost = 'localhost';
```

However, osTicket's wrappers (class.email.php and class.mailer.php) never pass this parameter to the factory. Thus, spam filters will be triggered (FSL_HELO_NON_FQDN_1 and HELO_LOCALHOST), since the default 'localhost' value is sent in the HELO message.

As long as the fqdn of the server is defined in /etc/hosts, passing the following parameter to the factory seems to work:

```
'localhost' => gethostbyaddr(gethostbyname(gethostname()))
```

rentasad:
I have the same problem and beeing unable to configure my smtp for outgoing email with the failure: 

> Invalid HELO string (localhost)

Jieiku:
This is still an issue.... In fact my fail2ban for postfix is banning the IP when osticket sends me an email because:

`Helo command rejected: Host not found`

and, it currently looks like this:

`proto=ESMTP helo=<osTicket.localdomain>`

instead of osTicket.localdomain it would be nice if I could configure it to be

`helo=<osTicket.example.com>`

or

`helo=<example.com>`

JediKev:
@xekon 

That sounds like an issue with your Postfix config. There are many guides online that explain how to set the HELO name (`smtpd_banner`, `smtp_helo_name`, etc.). You can also configure `smtpd_helo_restrictions` to reject unknown HELO names.

Cheers.

coolacid:
Quick work around for v1.17 -- 

Look for `osTicket\Mail\SmtpOptions` around line 1201 of include/class.email.php

Add directly after the `$smtpOptions` is returned:

```
$smtpOptions->setName("YOUR HOSTNAME HERE");
```

PR #6303, opened 2022-10-09
smtp: Use proper name when saying What's Up (HELO/EHLO)

protich:
This PR addresses issue #3366  - where outgoing email can get rejected when the target mail server is a strict jerk - requiring a valid host to even say hey! How rude! /s

We were setting the name to null which resulted in localhost being used as the name. Now we're asking the OS to give us the proper hostname of the server - hopefully the server admin has the correct hostname and reverse-lookup setup correctly for better email deliverability. 

**How to setup hostname for Linux users**
- `hostnamectl` will give you currently set hostname.
- To set one or change it - use ```hostnamectl set-hostname osticket.yourdomain.com``` with a valid DNS entry for reverse lookup. For local environment you can just add the entry to `/etc/hosts`.

### DIFF
--- a/include/class.mail.php
+++ b/include/class.mail.php
@@ -617,8 +617,9 @@ namespace osTicket\Mail {
                 case $auth instanceof NoAuthCredentials:
                     // No Authentication - simply return host and port
                     return [
-                        'host'      => $connect['host'],
-                        'port'      => $connect['port']
+                        'host' => $connect['host'],
+                        'port' => $connect['port'],
+                        'name' => $connect['name'],
                     ];
                     break;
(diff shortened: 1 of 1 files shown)


ChrisTG742:
This is a very useful improvement! I've seen quite a few providers getting more strict regarding host identities. And for the others not (yet) needing it, it still does not harm to provide it. ;)

The only thing one could question is, if the hostname of the machine is really always what you want (think about virtual hosts etc.), so querying $_SERVER['HOST_NAME'] could also be an alternative.

JediKev:
Tested and works 👍🏻

PR #6306, opened 2022-10-09
session: Add Noop Session Backend

protich:
osTicket support uploading your own logos as well as login backgrounds and calls logo.php (on both panels) to fetch the image in question. Previously we didn't care about session data so session handler was set to an inline noop function as a way to bypass starting a "real" session. The problem with the approach is; if osTicket is hosted behind a proxy with multiple web severs in the backend, the proxy needs to send all requests to the same server for specific user session. To do so session name (OSTSESSID) is used to route requests from the same session to the same backend server.

This PR adds a noop session backend allowing for ability to start a session while still ignoring data writes. Noop backend is used when DISABLE_SESSION is defined and set to false.

### DIFF
--- a/include/class.ostsession.php
+++ b/include/class.ostsession.php
@@ -18,6 +18,7 @@ class osTicketSession {
     static $backends = array(
         'db'        => 'DbSessionBackend',
         'memcache'  => 'MemcacheSessionBackend',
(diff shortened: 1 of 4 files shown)


JediKev:
See https://github.com/osTicket/osTicket/pull/6334

PR #6308, opened 2022-10-10
issue: File Upload Size

JediKev:
This addresses an issue where if the Issue Details field or any custom FileUpload field don't have `configuration` set any file upload to such fields will fail with `File too large` error. This is due to the `configuration` not being set so when we check the uploaded file size to the allowed file `size` from the field's config via `$config['size']` it fails. This adds a check for the `size` option in the `$config` array and if not set or set to NULL then we will default to 1MB. PHP's default `upload_max_filesize` is 2MB so this shouldn't be an issue (unless you set it lower - why tho??).

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -3965,7 +3965,11 @@ class FileUploadField extends FormField {
     }
 
     function getConfiguration() {
+        global $cfg;
+
         $config = parent::getConfiguration();
+        // If no size present default to system setting
+        $config['size'] ??= $cfg->getMaxFileSize();
         $_types = self::getFileTypes();
         $mimetypes = array();
         $extensions = array();


include/class.forms.php
```diff
@@ -3966,6 +3966,8 @@ function getFiles() {
 
     function getConfiguration() {
         $config = parent::getConfiguration();
+        // If no size present default to 1 MB
+        $config['size'] ??= 1048576;
```
JediKev: better way to do this? 🤔

ChrisTG742:
I have a proposal for an improvement:

```
function return_bytes($val)
{
  $units = array('', 'K', 'M', 'G', 'T', 'P', 'E', 'Z', 'Y');
  $val = trim($val);
  if ($val != "") {
    $unit = strtoupper($val[strlen($val) - 1]);
    if (in_array($unit, $units)) {
      $num = floatval(substr($val, 0, strlen($val) - 1));
      return ($num * pow(1024, array_search($unit, $units)));
    } else {
      return $val;
    }
  } else {
    return false;
  }
}

(...)

$config['size'] ??= return_bytes(ini_get('post_max_size'));
```

This is saver than just expecting a value to be present.
(Dunno if we still need the conversion to bytes as a separate function or if something like that is yet present in some class.)

Edit: Improved function a bit

PR #6312, opened 2022-10-12
update laminas-mail to release 2.18.0 and disable certificate validation on connections to mail servers

JensEB:
This pull request updates laminas-mail to 2.18.0 and its depencies. This request removes all dev-depencies, too.
We don't develop laminas-mail. So we need no development-depencies.

This I have done:
I downloaded the latest release of laminas-mail from github.
Then I update laminas-mail via composer.
I opened the folder ./include/laminas-mail/ in my console and run following command:
composer --no-dev update
This command updates all laminas-mail depencies from require-section and removes all dev-required depencies.
So we reduce the file count after updating to the latest version from 3500 files to less than 500 files.
This will reduce laminas-mail from 15 MB to 3,5 MB
For more details, see ./include/laminas-mail/composer.json

We don't need development parts from laminas-mail and this framework should not be greater than osTicket itself...

Last but not least, I disabled the certificate validation on mail server connection.
The new version of laminas-mail uses stream_socket_client() and not fsockopen() to connect to the mail server and has a setting to disable certificate validation. It solves many errors, too. So it solves encoding issues, that destroys fetched mails, if the source code is not probably encoded.
(See Release note to 2.14.2 https://github.com/laminas/laminas-mail/releases)


JediKev:
@JensEB 

I rewrote your pull for laminas-mail 2.23.0 as well as removed some custom code we had in-place until it was merged upstream (which it now is).
- https://github.com/osTicket/osTicket/pull/6543

Cheers.

PR #6319, opened 2022-10-15
Set timeout on smtp connections to avoid an RuntimeException

JensEB:
Some mail server close the connection before laminas-mail can quit the connection after osTicket sent an email. This will raise an RuntimeException.
To avoid this behavior, we can set an timeout to keep the connection alive. Laminas-mail will quit the connection by itself, if the job is done and send a QUIT command.

For more details see:
https://docs.laminas.dev/laminas-mail/transport/smtp-authentication/#smtp-transport-usage-for-servers-with-reuse-time-limit and
https://github.com/laminas/laminas-mail/issues/158

### DIFF
--- a/include/class.email.php
+++ b/include/class.email.php
@@ -1199,6 +1199,10 @@ class SmtpAccount extends EmailAccount {
             $setting = $this->getAccountSetting();
             $setting->setCredentials($this->cred);
             $smtpOptions = new osTicket\Mail\SmtpOptions($setting);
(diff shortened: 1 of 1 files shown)


JediKev:
@JensEB 

Thank you, however we have a patch coming for this already; we just haven’t made a pull yet.

Cheers.

JensEB:
Great, I will wait for it...

JediKev:
@JensEB 

As promised, here is said pull:
- https://github.com/osTicket/osTicket/pull/6350

Cheers.

PR #6328, opened 2022-10-20
Bug fix in 1.17

RatioJuris:
Fixed "abstract private function oauth2Auth(AccessToken $token);" @ 238 which was throwing error that it cannot be private so changed to "abstract public function oauth2Auth(AccessToken $token);"

### DIFF
--- a/include/class.mail.php
+++ b/include/class.mail.php
@@ -1,0 +1,880 @@
+<?php
+/**
+ * class.mail.php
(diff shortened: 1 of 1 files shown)


JediKev:
@isrgrajan 

Thank you however this has already been addressed:
- https://github.com/osTicket/osTicket/pull/6297

Cheers.

PR #6332, opened 2022-10-24
issue: Plugin Config Item Exists

JediKev:
This addresses an issue where the Audit Log plugin shows Ticket Views in the Audit Logs and shows the setting as enabled even when you disable and save via plugin config. This is due to the current `Config::exists()` method using the config `->get()` method and checking the return value to determine `true` or `false`. This is problematic, especially when values can be `NULL` or `0`. This updates the `Config::exists()` method to check if the key exists in the session or config array. If the key does exist then we return `true` otherwise we return `false`.

### DIFF
--- a/include/class.config.php
+++ b/include/class.config.php
@@ -77,7 +77,12 @@ class Config {
     }
 
     function exists($key) {
-        return $this->get($key, null) ? true : false;
+        if ((isset($this->session) && array_key_exists($key, $this->session))
+                || array_key_exists($key, $this->config)
+                || array_key_exists($key, $this->defaults))
+            return true;
+
+        return false;
     }
 
     function set($key, $value) {


include/class.config.php
```diff
@@ -77,7 +77,11 @@ function get($key, $default=null) {
     }
 
     function exists($key) {
-        return $this->get($key, null) ? true : false;
+        if ((isset($this->session) && array_key_exists($key, $this->session))
+                || array_key_exists($key, $this->config))
+            return true;
+
```
protich, line 84: While this somehow addresses the issue I think it's the wrong approach 1) Doesn't consider defaults & Duplicates the code etc. 

I think simply checking for the returned default is sufficient.

```php
return  ($this->get($key, null) !== null);
```

PR #6333, opened 2022-10-25
issue: urldecode Dispatcher

JediKev:
This addresses an issue where if using a webserver that encodes URLs by default (like NGINX) some of the AJAX URLs will not resolve. This ensures the URL is decoded with `urldecode()` before matching and dispatching.

### DIFF
--- a/include/class.dispatcher.php
+++ b/include/class.dispatcher.php
@@ -33,6 +33,8 @@ class Dispatcher {
             $_SERVER['REQUEST_METHOD'] = strtoupper($_GET['_method']);
             unset($_GET['_method']);
         }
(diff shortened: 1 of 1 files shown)


include/class.dispatcher.php
```diff
@@ -27,6 +27,7 @@ function __construct($file=false) {
     }
 
     function resolve($url, $args=null) {
+        $url = urldecode($url); // Decode URL for accurate matching
```
protich: I would move this above foreach - line 37.  Simply to group code blocks that are used together.

protich (approved):

PR #6338, opened 2022-10-31
fix database timezone issues

daynebenn:
switch from using `system_time_zone` which is automatically set when the database starts for the first time, to using `time_zone` which is the timezone that the database is actually using.

the `system_time_zone` variable is only used if the `time_zone` variable is set to `SYSTEM`

the `time_zone` variable is what the database is actually using.

if the `time_zone` variable is set to `SYSTEM`; use the `system_time_zone` variable

### DIFF
--- a/include/mysqli.php
+++ b/include/mysqli.php
@@ -116,7 +116,11 @@ function db_version() {
 }
 
 function db_timezone() {
-    return db_get_variable('time_zone', 'global');
+    $tz = db_get_variable('time_zone', 'global');
+    if ($tz == 'SYSTEM') {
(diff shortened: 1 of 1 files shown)


JediKev:
@daynebenn 

I've already made this exact same pull a while ago:
- https://github.com/osTicket/osTicket/pull/4186

Please read the original thread. Cheers.

daynebenn:
@JediKev
oh my goodness... lolol apologies for not seeing that... I just couldn't believe how long this has been an issue in osticket so wanted to suggest a fix for it. Looks like we came to the exact same conclusion as to how to resolve the issue :cloud:

PR #6342, opened 2022-11-02
issue: Miscellaneous Fixes

JediKev:
This includes some miscellaneous fixes we've accumulated since previous releases.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -45,6 +45,15 @@ class Bootstrap {
         }
         date_default_timezone_set('UTC');
 
+        if (!function_exists('exif_imagetype')) {
+            function exif_imagetype ($filename) {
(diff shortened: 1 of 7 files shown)


include/class.api.php
```diff
@@ -174,7 +174,7 @@ function requireApiKey() {
         # header
 
         if(!($key=$this->getApiKey()))
-            return $this->exerr(401, __('Valid API key required'));
+            return $this->exerr(401, __('Valid API key required: ').$_SERVER['REMOTE_ADDR']);
```
protich: I think this was _only_ staged when I was debugging proxies \ load balancers setup... I don't think it's worth changing translation string for it, plus I think returning generic error makes sense from security standpoint. 

That said, we can perhaps log a detailed debug message to the system logs with the IP in the future for debug purposes.

PR #6344, opened 2022-11-03
fetcher: Mail Fetch Order

protich:
osTicket currently fetches messages in LIFO order meaning the most recent messages get fetched first. This was intentional to avoid possible stalemate when fetched messages are left in the inbox (Do Nothing Option - which is not recommened). However, this can create out of order threading issues since subsequest replies can be fetched first.

This PR (last commit) addresses the issue by changing the logic so we can fetch older messages first (FIFO order) as long as the numbers of the emails in the folder being fetched are less than maximum messages to process at a time.

### DIFF
--- a/include/class.email.php
+++ b/include/class.email.php
@@ -981,6 +981,16 @@ class MailBoxAccount extends EmailAccount {
             ->filter(['type' => 'mailbox']);
     }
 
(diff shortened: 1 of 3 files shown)


protich:
Message sequence number range bug fixed with the latest commit.

PR #6345, opened 2022-11-03
fetcher: Delayed Mailbox Expunction

protich:
As part of post-processing options, when fetching emails, osTicket can archive (move) or delete (remove) a mail message from the server. When using mail storage backend like IMAP - both operations requires `expunge` command to truly commit the changes to the mail server.

That's all good and well, except Laminas Mail (our upstream darling) expunges the mailbox in real-time when a message is moved or deleted. This creates a problem for osTicket because **expunging causes message sequence numbers to reset at the server end during the session!!** Which in turn causes mismatch when we process the next message in the queued sequence. Additionally, expunging on every message operation cases unnecessary traffic to the mail server when every second counts.

To resolve the issue - until we switch to using `UIDs`  instead of message sequence numbers -  this commit simply bypasses Laminas Mail IMAP's removeMessage function so we can skip expunging on message move/remove operations. Don't worry, osTicket expunges the mailbox at the end of the fetch session (granted we don't segfault).

### DIFF
--- a/include/class.mail.php
+++ b/include/class.mail.php
@@ -520,28 +520,21 @@ namespace osTicket\Mail {
         /*
          * move an existing message to a folder
          *
+         * Caller should catch possible exception
          */
-        public function moveMessage($id, $folder) {
-            try {
-                parent::moveMessage($id, $folder);
-                return true;
-            } catch (\Throwable $t) {
-                // noop
-            }
-            return false;
+        public function moveMessage($i, $folder) {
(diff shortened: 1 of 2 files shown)


protich:
PR updated. Please review the last commit.

protich:
Rebased latest changes on dependent branch with existing PR

protich:
Rebased on top of `1.17.x`

PR #6349, opened 2022-11-07
Mail Parse Error Handling

protich:
On mail parse error, osTicket typically leaves the email alone and reports it as a fetch failure - which results in multiple fetch attempts until an someone manually moves the email out of the Fetch Folder. 

Granted that most parse errors are caused by poorly formatted mime message (emails that failes to decode) - this PR adds ability to create a ticket even when osTicket encounters mail parse error as long as we have valid email headers. The entire raw email is attached to the ticket as `message/rfc822` (.eml) attachment.


protich:
PR updated with a couple more commits.

JediKev:
Needs rebase for `issue/fetcher+expunge` branch

protich:
Rebased on top of `issue/fetcher+expunge` branch.

PR #6353, opened 2022-11-08
Email Duplicate Header bug fix for Windows

tliscomb:
@JediKev here is my fix for windows facing the duplicate header bug with gmail.

### DIFF
--- a/include/class.mailer.php
+++ b/include/class.mailer.php
@@ -619,6 +619,7 @@ class Mailer {
 
         try {
             // ostTicket/Mail/Sendmail transport
+            $message->getHeaders()->removeHeader('Subject');
             $sendmail =  new  Sendmail($args);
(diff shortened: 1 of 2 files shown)


JediKev:
@tliscomb 

I see now you are talking about Sendmail specifically; I was assuming SMTP. We will look into this as modifying lamias-mail directly is not a good idea. Also it’d be better for you to submit this directly to laminas-mail instead as that’s the main repo.

To note, we only use Sendmail if you don’t setup SMTP. Ideally you’d want to configure SMTP for outgoing mail.

Cheers.

tliscomb:
I found that this change has the same affect and does not mess with the 3rd party. 

https://github.com/osTicket/osTicket/pull/6353/commits/c68314c9e423570050c9f8218ee009b55f2ba610

PR #6354, opened 2022-11-08
1.17.x

tliscomb:
@JediKev This is cleaner and does not affect 3rd party libs. 

This is the windows bug fix for email duplicate headers bug.

### DIFF
--- a/include/class.mailer.php
+++ b/include/class.mailer.php
@@ -619,6 +619,7 @@ class Mailer {
 
         try {
             // ostTicket/Mail/Sendmail transport
+            $message->getHeaders()->removeHeader('Subject');
             $sendmail =  new  Sendmail($args);
             if ($sendmail->sendMessage($message))
                 return $messageId;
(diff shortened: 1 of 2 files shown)


protich:
@tliscomb - Thank you very much for the bug find, report and PR.  However, removing the headers in mailer can result in outgoing emails without subject getting sent out. Here is [a better fix](https://github.com/osTicket/osTicket/pull/6355) for the issue until we upgrade Laminas Mail library in the upcoming future major release.

Please give it a test and let us know.

protich:
Issue addressed with #6355

tliscomb:
> @tliscomb - Thank you very much for the bug find, report and PR. However, removing the headers in mailer can result in outgoing emails without subject getting sent out. Here is [a better fix](https://github.com/osTicket/osTicket/pull/6355) for the issue until we upgrade Laminas Mail library in the upcoming future major release.
> 
> Please give it a test and let us know.

Works great! Thank you.

PR #6366, opened 2022-11-13
mailer: Use Namespace

protich:
This commit instantiates the mailer using the correct namespace.

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -323,7 +323,7 @@ class Bootstrap {
 
     static function croak($message) {
         $msg = $message."\n\n".THISPAGE;
-        Mailer::sendmail(ADMIN_EMAIL, 'osTicket Fatal Error', $msg,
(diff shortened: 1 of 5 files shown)


JensEB:
There is another place where the namespace is missing...
In bootstrap.php in the static function croak($message), the mailer is used without namespace, too.

    static function croak($message) {
        $msg = $message."\n\n".THISPAGE;
        osTicket\Mail\Mailer::sendmail(ADMIN_EMAIL, 'osTicket Fatal Error', $msg,
            sprintf('"osTicket Alerts"<%s>', ADMIN_EMAIL));
        //Display generic error to the user
        Http::response(500, "<b>Fatal Error:</b> Contact system administrator.");
    }

JediKev:
@JensEB 

Thank you for the comment. We are aware; it's being added soon.

Cheers.

PR #6383, opened 2022-11-20
Use correct case for class references

gerundt:
I wanted to play with the static analysis tool [PHPstan](https://phpstan.org/) but have still problems with [discovering symbols](https://phpstan.org/user-guide/discovering-symbols). 🙈

But I fixed at least the incorrect case for some class references. I also found functions which need the `static` keyword.

### DIFF
--- a/ajax.php
+++ b/ajax.php
@@ -48,5 +48,5 @@ $dispatcher = patterns('',
     ))
 );
 Signal::send('ajax.client', $dispatcher);
-print $dispatcher->resolve(Osticket::get_path_info());
+print $dispatcher->resolve(osTicket::get_path_info());
 ?>
--- a/api/http.php
+++ b/api/http.php
@@ -26,5 +26,5 @@ $dispatcher = patterns('',
 // Send api signal so backend can register endpoints
 Signal::send('api', $dispatcher);
 # Call the respective function
-print $dispatcher->resolve(Osticket::get_path_info());
+print $dispatcher->resolve(osTicket::get_path_info());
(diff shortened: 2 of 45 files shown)


protich:
@gerundt - This is awesome! We were discussing internally how we need to replace our homwgrown lint tests with PHPstan.

Given that you've already started on it, do you mind taking a stab at it?

My thinking is when you run tests   `php setup/test/run-tests.php` - it can trigger hydration via composer (like we do in with Plugins) so we don't have to ship PHPstan with osTicket.

Please let me know if you need further clarification and thanks once again!

gerundt:
@protich Good to know, that you have interest in PHPstan. I will continue to investigate! At the moment it is only rudimental running. I have created the file `phpstan.neon` in the root folder and use the [PHPstan version in a docker container](https://phpstan.org/user-guide/docker).

# phpstan.neon

```
parameters:
	level: 0
	paths:
		- .
	excludePaths:
		analyse:
			- include/laminas-mail
			- include/mpdf
			- include/pear
			- setup/test
	bootstrapFiles:
		- bootstrap.php
```
Biggest problem is, that PHPstan use the autoloader from Composer. We can add a second *bootstrap* file only for PHPstan and include the important files, but currently it crash because of missing includes inside the osTicket classes. Its hard work to find this missing includes. 🙈

But I work on it! 👍

JediKev (commented):
So far this looks good. None of the changes (as of now) will affect functionality nor break anything 👍🏻

PR #6386, opened 2022-11-23
typo: Fix verify_email_addrs Help Tip

bestlong:
This fixes a typo with the `verify_email_addrs` help tip where `perfomed` should be `performed`.

### DIFF
--- a/include/i18n/en_US/help/tips/settings.email.yaml
+++ b/include/i18n/en_US/help/tips/settings.email.yaml
@@ -135,5 +135,5 @@ verify_email_addrs:
     content: >
         Enable this option to check if the email address has a mail
         exchanger (MX) in the domain's DNS. This is useful to detect
-        incorrectly typed email addresses. This is perfomed in addition to
+        incorrectly typed email addresses. This is performed in addition to
         checking the email address wellformedness.


JediKev:
@bestlong 

Thank you for the pull request. While the fix looks good I do have some suggestions:
- With osTicket we support two major versions at a time. For instance, currently we support `1.16.x` and `1.17.x`. Before making a pull request think about if it can address the issue in _only_ the latest version or _both_ supported versions. In this case, your typo can address the issue in `1.16.x` AND `1.17.x`. So what you want to do is make a branch off of the earlier version branch (in this case make a branch off of `1.16.x`) and make a pull request from there. That way we merge it into `1.16.x` and that gets merged into the latest version. **Note:** If the issue is only present in the latest version then you can make a branch off of `develop`.
- When making a branch it's good practice to come up with a short descriptive name for it. For example, in this case I would've put `typo/fix-verify-email-addrs-help-tip` as the branch name. That way when someone is looking at the commit history they see what the commit was related to a lot faster/easier.
- In the same realm, when making a commit it's always good practice to put a descriptive title and body. The title is typically a neater version of the branch name and the body is a good description of what the issue is and describes the changes that were made to fix it. We use the following format:
```git
typo: Fix verify_email_addrs Help Tip

This fixes a typo with the `verify_email_addrs` help tip where `perfomed` should be `performed`.
```

Hopefully this helps you out with future contributions in this project and others.

Cheers.

PR #6400, opened 2022-12-04
fetcher: Default Department

protich:
When fetching a mailbox, osTicket routes fetched emails to a department based on `To` or `Delivered-To` header entries. This allows for ability to fetch aliased emails from one primary account and route them based on configured system emails. On failure to match, the email (ticket) is routed to the department of the email account being fetched.

With the email backend overhaul in v1.17 - a bug was instroduced where unmatched emails ended up being routed to default department instead of the department of the email being fetched.

This PR addresses the issue...

### DIFF
--- a/include/api.tickets.php
+++ b/include/api.tickets.php
@@ -171,7 +171,7 @@ class TicketApiController extends ApiController {
         return $this->processEmail();
     }
 
-    function processEmail($data=false) {
+    function processEmail($data=false, array $defaults = []) {
 
         try {
             if (!$data)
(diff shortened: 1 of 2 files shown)


JediKev (approved):
Tested and working!

rcosg:
Hi, we have the same issue, the fix doesn't work for us

Are these the only 2 files that need to be changed, or are more changes needed?

Enviroment:
image

webmurkser:
Hi, the same problem here. In version 1.15.x assignment of a separate mailbox -> department without problems. After the update to 1.17.2 the assignment is a "gamble". Sometimes it works, sometimes not. 
SCR-20221230-lu2

PR #6419, opened 2022-12-28
Provide the ability to create tasks from code

empty88:
Provide the ability to create tasks from code, e.g. for custom filter actions


empty88:
sorry, messed it up

PR #6424, opened 2022-12-29
add translation for Event titles on reopened/closed

empty88:


include/class.task.php
```diff
@@ -587,7 +587,7 @@ function setStatus($status, $comments='', &$errors=array()) {
                 if ($t->ticket) {
                     $t->ticket->reopen();
                     $vars = array(
-                            'title' => sprintf('Task %s Reopened',
+                            'title' => sprintf(__('Task #%s reopened'),
```
JediKev, line 590: This should be broken into separate words using `sprintf()`

include/class.task.php
```diff
@@ -614,7 +614,7 @@ function setStatus($status, $comments='', &$errors=array()) {
 
                 if ($t->ticket) {
                     $vars = array(
-                            'title' => sprintf('Task %s Closed',
+                            'title' => sprintf(__('Task #%s closed'),
```
JediKev, line 617: Same suggestions as the above comment.

PR #6427, opened 2023-01-04
issue: Does Not Have A Value

JediKev:
This addresses an issue where using the `Does Not Have A Value` criteria on things like Short Answer Fields shows incorrect results. In fact, it shows all Tickets with and without values for the Field. This is due to using an overall negative filter (`column IS NULL` or `NOT column IS NOT NULL`) on a column from a LEFT JOIN causing the backend to return NULL for all values in that column. This adds a new variable called `$nullable` that is set to `false` if the `$method` is `nset`. This then passes `$nullable` to `getOrmPath()` so it can determine wether or not to force a normal JOIN instead of LEFT JOIN so that `xxx IS NULL`/`NOT xxx IS NOT NULL` will return the proper results. If the search method is not `nset` then it will use LEFT JOIN as before.

### DIFF
--- a/include/class.queue.php
+++ b/include/class.queue.php
@@ -976,9 +976,10 @@ class CustomQueue extends VerySimpleModel {
                 $qs = $ost->searcher->find($value, $qs, false);
             }
             else {
+                $nullable = ($method === 'nset') ? false : null;
                 // XXX: Move getOrmPath to be more of a utility
                 // Ensure the special join is created to support custom data joins
-                $name = @static::getOrmPath($name, $qs);
+                $name = @static::getOrmPath($name, $qs, $nullable);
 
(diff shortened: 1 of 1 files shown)


include/class.queue.php
```diff
@@ -1481,6 +1481,10 @@ static function getOrmPath($name, $query=null) {
             $root = $query->model;
             $meta = $root::getMeta()->getByPath($path[1]);
             $joins = $meta['joins'];
+            // If the method is 'nset' (ie. IS NULL) we want to force normal
+            // JOIN instead of LEFT JOIN to get proper results
+            if ($method === 'nset')
```
JediKev: Compare method in mangleQuerySet and pass bool to getOrmPath

PR #6430, opened 2023-01-05
Receive alert on assignment, even alerts for the department are disa…

empty88:
…bled

I think when a ticket is assigned to a team or staff, the team members/staff should receive an alert for the assignment even nobody wants to receive alerts from department.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -2062,7 +2062,7 @@ implements RestrictedAccess, Threadable, Searchable {
         }
         $dept = $this->getDept();
         // See if we need to send alerts
-        if (!$alert || !$cfg->alertONAssignment() || !$dept->getNumMembersForAlerts())
+        if (!$alert || !$cfg->alertONAssignment())
             return true; //No alerts!
 
         if (!$dept


JediKev:
@empty88 

Why would you force alerts for people who have them explicitly turned off? Doesn't make sense to allow disabling if you are just going to force it.

This doesn't make sense for the system overall so I am closing this.

Cheers.

empty88:
@JediKev 
My description was a bit missleading.
`|| $dept->getNumMembersForAlerts()` checks in other words if the dept of the ticket has active primary members or extended members with allerts turned on.

- Case 1: The dept of the ticket has only primary members (alerts auto on).
=> Every team member of the team, the ticket will be assigned to, will be notified if alerts for the team are turned on  - even team members who don't have access to the dept will be notified.
- Case 2: The dept of the ticket has only primary members, which are disabled agents for any reason.
=> **Not a single team member of the team, the ticket will be assigned to, will be notified** - even though alerts for the team are turned on
- Case 3: The dept of the ticket has no primary just extended members. All extended Members have their notification for the dept turned off
=> **Not a single team member of the team, the ticket will be assigned to, will be notified** - even though alerts for the team are turned on

I don't know if this behavior really makes sense.

Issue #5898, opened 2021-05-26
Incorrect count and pagination results for saved searches using dynamic form fields as criteria


kler:
### Prerequisites

* [X] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [ ] Do you have any errors in the PHP error log, or javascript console?
* [X] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [X] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

### Description

OsTicket have the ability to add custom form fields using Lists, giving the user the ability to for instance choose from a set of predefined options.
It's also possible to search tickets using dynamic form field choices as criteria's, to for instance look for tickets where the user have picked option **A** out of possible options **A**, **B** and **C**.

Because of issue #4972 counts (and therefore pagination) is not available in normal nor advanced searches; however counts *are* in fact available for saved searches, despite being possibly incorrect when using dynamic form fields as criteria as will be explored below.

Besides OsTicket showing the possibly incorrect count, this also has the side effect that the pagination is incorrect (since it's based on the count result).

![image](https://user-images.githubusercontent.com/966132/119651164-014fb380-be25-11eb-86f2-0f0b4c2ce236.png)

### Steps to Reproduce

- Create a List, example name: ShipList
  - Add three entries to ShipList:
    - Value: Ship 1
    - Value: Ship 2
    - Value: Ship 3
- Create a Form, example name: ShipList
  - Add entry:
    - Label: ShipList
    - Type: Custom List => ShipList
    - Visibility: Required
    - Variable: SHIP
- Create a Help topic, example name: ShipList
  - Under Forms, Add custom form: ShipList
  - Make sure label ShipList is enabled
- Create a Form, example name: FreeTextForm
  - Label: FreeTextForm
  - Type: Long Answer
  - Visibility: Required
  - Variable: FREE_TEXT_FORM
- Create a Help topic, example name: FreeTextForm
  - Under Forms, Add custom form: FreeTextForm
  - Make sure label FreeTextForm is enabled
- Create a new Ticket using Help topic called ShipList
  - Choose **Ship 1** from ShipList
- Verify content in table ost_form_entry_values
  - Entry is stored like this: `{"3":"Ship 1"}`
  - If the index integer is different from `3` in your case, take note of this and replace in following examples.
- Create a saved search with name: **Ship 1**
  - Add other field: **Ships / Ships**
  - Includes: **Ship 1**
[...]

PR #6435, opened 2023-01-11
issue: Queue Counts Incorrect

JediKev:
This addresses an issue where when a Queue is filtering by a specific field the counts are incorrect. This is due to using only the `constraints` instead of both `constraints` and `path_constraints`. This updates the `counts()` method to include the `path_constraints` if set so that the counts are properly calculated.

### DIFF
--- a/include/class.search.php
+++ b/include/class.search.php
@@ -984,7 +984,13 @@ class SavedQueue extends CustomQueue {
             }
 
             if ($Q->constraints && !$empty) {
-                $expr = SqlCase::N()->when(new SqlExpr(new Q($Q->constraints)), new SqlField('ticket_id'));
+                $constraints = $Q->constraints;
+                // Add path_constraints to get the correct counts
+                foreach ($Q->path_constraints as $pc) {
+                    if (!empty($pc[0]->constraints))
+                        $constraints[] = $pc[0];
+                }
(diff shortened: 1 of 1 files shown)


include/class.search.php
```diff
@@ -984,7 +984,13 @@ static function counts($agent, $cached=true, $criteria=array()) {
             }
 
             if ($Q->constraints && !$empty) {
-                $expr = SqlCase::N()->when(new SqlExpr(new Q($Q->constraints)), new SqlField('ticket_id'));
+                $constraints = $Q->constraints;
+                // Add path_constraints to get the correct counts
+                foreach ($Q->path_constraints as $pc) {
+                    if (!empty($pc[0]->constraints))
```
JediKev, line 990: May need to add another foreach for the $pc but still looking into that

JediKev: test annotation counts

JediKev: annotations are counting correctly here 👍🏻

JediKev: each path_constraint is separate so no need for another foreach (thank goodness)

jhlindnerberlin:
Is this resolving Issue #5898 ?

JediKev:
@jhlindnerberlin 

Not directly, but you can try it to see.

Cheers.

PR #6445, opened 2023-01-24
Auto-Close Answered Tickets

MarcBanyard:
This adds the ability to automatically close answered tickets after "X" amount of days of inactivity from the customer.

New Email Template created for "Closed Tickets"
The ability has also been added to the system to send Closed Ticket emails to customers.

Both of the above can be set per department (Admin Panel > Agents > Departments > Selected Department).

### DIFF
--- a/include/class.cron.php
+++ b/include/class.cron.php
@@ -28,6 +28,16 @@ class Cron {
     static function TicketMonitor() {
         require_once(INCLUDE_DIR.'class.ticket.php');
         Ticket::checkOverdue(); //Make stale tickets overdue
+        global $cfg;
+        $depts=Dept::getDepartments();
+        $a = 0;
+        foreach($depts as $d) {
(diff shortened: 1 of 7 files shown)


Steltek:
Minor issue with this PR: It relies on the "updated" timestamp, which includes miscellaneous events, like osTicket flagging a ticket as overdue. This leads to situations like this:

- Tickets should be closed after X days.
- Tickets are set to be marked as overdue in Y days where Y is < X.

Let X be 7 and Y be 3:

Day 0: Agent replies to ticket.
Day 3: The system flags the ticket as "overdue".
Day 7: The ticket is expected to be closed, but the 'updated' date refers to day 3.
Day 10: Only now is the ticket closed.


The "last reply" is tracked in the "lastupdate" field so to counter the above, the code change should work from:

'updated__lt' => SqlFunction::NOW()->minus(SqlInterval::DAY($days)))

to

'lastupdate__lt' => SqlFunction::NOW()->minus(SqlInterval::DAY($days)))

PR #6448, opened 2023-01-24
Ability to remove staff login link from the client login page

MarcBanyard:
Added the ability to remove the Staff Login Link from the client login page from the "Settings > System" page.


ChrisTG742:
Very useful enhancement! :)

TimzOwen (commented):
works perfect , what about hiding it instead of removing?

PR #5564, opened 2020-06-25
Feature: help topic organization and primary contact restrictions

KarlPerkins:
Modified from [ Pull Request #3494](https://github.com/osTicket/osTicket/pull/3494) to work with osTicket 1.41.x

This adds support for ability to restrict help topics to particular organizations and primary contacts.

Full details: [Issue #3156](https://github.com/osTicket/osTicket/issues/3156)

PR #3494, opened 2016-11-10
Feature/help topic organization and primary contact restrictions

cyclops26:
This adds support for ability to restrict help topics to particular organizations and primary contacts.

Full details: [https://github.com/osTicket/osTicket/issues/3156]

PR #6454, opened 2023-01-30
Extra controls for Help Topics to allow restricting them to Organizations and/or Primary Contacts

nekhbet:
This PR is based on https://github.com/osTicket/osTicket/pull/5564 by Karl Perkins (which, in turn, is based on https://github.com/osTicket/osTicket/pull/3494 by Eli Keimig).

**Changes** in regard to the last PR: 

- We made it work with the latest osTicket version.
- Replaced “each” function calls as they got deprecated in PHP 7.2 and removed in PHP 8.0.
- If a Help Topic is set to be shown only to allowed organizations or primary contacts it won’t be shown for anonymous users (validation is both frontend and backend).

The PR adds the functionality needed to enable Help Topics to be restricted to Organizations and/or Primary Contacts without affecting the existing Help Topics. 

All the changes provide additional features without **backward compatibility** issues:

- On an upgrade/new install, the default behavior of an existing Help Topic is to show it to everyone.
- The default behavior of any newly added Help Topic with no specified "Allowed Organizations" or “Primary Contacts Only” set to “no”,  is to show it to everyone.
- Because of this, you have the freedom to just ignore the new functionality if it doesn’t fit your business logic.
- The agents, when creating a new ticket for another person, see all Help Topics so they can manually assign if needed, a restricted Help Topic to a specific user’s ticket.

**How to use it?**

To use this functionality you need to add your organizations: 

msf1

Then add/update Help Topics from the Admin Panel:

Screenshot from 2023-01-27 08-27-24

With the desired permissions: 

- “**Allowed Organizations**” - Select which organizations can see and submit the current Help Topic. The default is to allow all organizations to see the topic.
- “**Primary Contacts Only**” - This determines whether or not a Help Topic is available to all users or just the primary contacts of the organization.

Then, on the frontend, the specific Help Topic will either: 

- Appear: 

  - To unauthenticated users, if there are no permissions set to the Help Topic.
  - To registered users, if they are a part of an allowed organization (if “Primary Contacts Only” is set to “Disabled”).
  - To registered users, if they are a part of an allowed organization and are a primary contact for an allowed organization (if “Primary Contacts Only” was set to “Enforced”).

- Not appear:
  - To unauthenticated users, if there are any permissions set to the Help Topic.
  - To registered users, if they are not part of an allowed organization.
  - To registered users, if they are not part of an allowed organization and they are not a primary contact for any of the allowed organizations (if “Primary Contacts Only” was set to “Enforced”).

Screenshot from 2023-01-26 11-08-33

### DIFF
--- a/bootstrap.php
+++ b/bootstrap.php
@@ -131,6 +131,7 @@ class Bootstrap {
 
         define('TOPIC_TABLE',$prefix.'help_topic');
         define('TOPIC_FORM_TABLE',$prefix.'help_topic_form');
+        define('TOPIC_ORGANIZATION_TABLE',$prefix.'help_topic_organization');
         define('SLA_TABLE', $prefix.'sla');
 
         define('EMAIL_TABLE',$prefix.'email');
--- a/include/class.client.php
+++ b/include/class.client.php
@@ -85,6 +85,8 @@ implements EmailContact, ITicketUser, TemplateVariable {
 
     function getId() { return ($this->user) ? $this->user->getId() : null; }
     function getEmail() { return ($this->user) ? $this->user->getEmail() : null; }
+    function isPrimaryContact() { return ($this->user) ? $this->user->isPrimaryContact() : null; }
+    function getOrganization() { return ($this->user) ? $this->user->getOrganization() : null; }
     function getName() {
         return ($this->user) ? $this->user->getName() : null;
     }
--- a/include/class.organization.php
+++ b/include/class.organization.php
@@ -160,6 +160,55 @@ class OrganizationModel extends VerySimpleModel {
     static function getPermissions() {
         return self::$perms;
     }
+
+    static function getOrganizations($localize=true) {
+        global $cfg;
+        static $organizations, $names = array();
+
+        // If localization is specifically requested, then rebuild the list.
+        if (!$names || $localize) {
+            $objects = self::objects()->values_flat(
+                'id', 'name'
+            )
+            ->order_by('name');
+
+            // Fetch information for all topics, in declared sort order
+            $organizations = array();
+            foreach ($objects as $O) {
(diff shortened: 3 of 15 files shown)


TechCare-Aus:
This is a great idea. Would love to see it part of the core.

jonsqp:
This is exactly what we need too, please add this to the osTicket core, we would be extremely grateful!

waczny:
This is what I need pretty cool.
1 quick question? 
Can I apply this on a current running system? 
I was just wondering because of the last file setup/inc/streams/core/install-mysql.sql won't be available after an installation or I can not re-run the installation with the same version.

waczny:
I found a bug. 
If you do the modification, you can not see the help topics in the queue helptopic filter if you put any or all organizations into the help topic. 
image
Can you fix this?

Elektroyeti:
> I found a bug. If you do the modification, you can not see the help topics in the queue helptopic filter if you put any or all organizations into the help topic. image Can you fix this?

This can be fixed by changing the following line in /include/class.search.php
@@ -1200,7 +1200,7 @@ class HelpTopicChoiceField extends AdvancedSearchSelectionField {
function getChoices($verbose=false, $options=array()) {
global $thisstaff;
if (!isset($this->_topics)) {
- $this->_topics = $thisstaff ? $thisstaff->getTopicNames(false, Topic::DISPLAY_DISABLED) :
+ $this->_topics = $thisstaff ? $thisstaff->getTopicNames(false, Topic::DISPLAY_DISABLED, null, true) :
Topic::getHelpTopics(false, Topic::DISPLAY_DISABLED);;
}

artyPD:
Are there any plans to merge this PR, @JediKev? This feature would be a huge help!

PR #6466, opened 2023-02-13
oauth2: Strict Matching

JediKev:
This is a rewrite of osTicket/osTicket-plugins#263 where it was attempting to implement Strict Matching. This fully implements Strict Matching and adds functionality to check for this when retrieving a token and when saving SMTP with Same As Remote Mailbox. This also adds a new disabled field to the General Settings to show what email will be used as well as a help tip that further explains Strict Matching.

### DIFF
--- a/include/class.email.php
+++ b/include/class.email.php
@@ -1276,10 +1276,17 @@ class SmtpAccount extends EmailAccount {
         return (strcasecmp($this->getAuthBk(), 'mailbox') === 0);
     }
 
+    /*
+     * Check if using mailbox auth and MailboxAccount exists if so
+     * return the MailboxAccount config, otherwise return it's own
+     * config
+     */
     protected function getConfig() {
-        return $this->isMailboxAuth()
-                ? $this->getEmail()->getMailboxAccount()->getConfig()
-                : parent::getConfig();
+        if ($this->isMailboxAuth()
+                && ($email=$this->getEmail())
+                && ($account=$email->getMailboxAccount()))
+            return $account->getConfig();
+        return parent::getConfig();
     }
 
     public function allowSpoofing() {
@@ -1335,7 +1342,11 @@ class SmtpAccount extends EmailAccount {
                 && !($creds=$this->getFreshCredentials($vars['smtp_auth_bk'])))
             $_errors['smtp_auth_bk'] = __('Configure Authentication');
 
-        if (($vars['smtp_auth_bk'] === 'mailbox') && !$this->checkStrictMatching())
+        // Check if set to active and using mailbox auth, if so check strict
+        // matching.
+        if ($vars['smtp_active'] == 1
+                && ($vars['smtp_auth_bk'] === 'mailbox')
+                && !$this->checkStrictMatching())
             $_errors['smtp_auth_bk'] = sprintf('%s and %s', __('Resource Owner'), __('Email Mismatch'));
 
         if (!$_errors) {
--- a/include/class.email.php
+++ b/include/class.email.php
@@ -1394,10 +1394,16 @@ class SmtpAccount extends EmailAccount {
  *
  */
 class EmailAccountConfig extends Config {
+    /*
+     * Get strict matching (default: true)
+     */
     public function getStrictMatching() {
-        return $this->get('strict_matching', false);
+        return $this->get('strict_matching', true);
     }
 
+    /*
+     * Set strict matching
+     */
     public function setStrictMatching($mode) {
         return $this->set('strict_matching', !!$mode);
     }
--- a/include/class.email.php
+++ b/include/class.email.php
@@ -536,7 +536,8 @@ class EmailAccount extends VerySimpleModel {
             $id = sprintf('%s:%d',
                 $this->getAuthBk(), $this->getId());
             if ($this->isOAuthAuth())
-                $id .= sprintf(':%d', $this->getAuthId());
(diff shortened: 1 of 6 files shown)


include/class.email.php
```diff
@@ -487,6 +487,19 @@ public function isAuthBackendEnabled() {
             : true;
     }
 
+    public function isStrict() {
+        $config = ($this instanceof SmtpAccount
```
protich: Why not put this logic in `getConfig()` function?  

PS: We hadn't had an issue yet because getCredentials considered auth backend (i.e delegated to mailbox), but it makes sense to have it as part of getConfig.

include/class.email.php
```diff
@@ -487,6 +487,19 @@ public function isAuthBackendEnabled() {
...
+                && $this->getAuthBk() === 'mailbox')
+            ? $this->getEmail()->getMailboxAccount()->getConfig()
+            : $this->getConfig();
+        return $config->getStrictMatching();
+    }
+
+    public function checkStrictMatching($token=null) {
+        $token ??= $this->getAccessToken();
+        return $token->isMatch($this->getEmail()->getEmail(), $this->isStrict());
```
protich: Token can be null. I would recommend.

```php
return ($token && $token->isMatch($this->getEmail()->getEmail(), $this->isStrict()))
```

include/class.email.php
```diff
@@ -537,6 +553,11 @@ public function getEmail() {
         return $this->email;
     }
 
+    public function getAccessToken() {
+        $cred = $this->getFreshCredentials();
+        return $cred->getAccessToken($this->getConfigSignature());
```
protich: Credential can be null here!

```php
return ($cred && $cred->getAccessToken($this->getConfigSignature())); 
```

protich (changes requested):
Some comments on the initial review.

protich:
To avoid requiring OAuth2 Plugin version upgrade in tandem with the core release - I'm suggesting we tag the strict mode flag to the backend id as shown below, for now, until we release v1.18.x series.

```diff
diff --git a/include/class.email.php b/include/class.email.php
index bc4943c6..0a1adea4 100644
--- a/include/class.email.php
+++ b/include/class.email.php
@@ -522,7 +522,8 @@ class EmailAccount extends VerySimpleModel {
             $id = sprintf('%s:%d',
                 $this->getAuthBk(), $this->getId());
             if ($this->isOAuthAuth())
-                $id .= sprintf(':%d', $this->getAuthId());
+                $id .= sprintf(':%d:%b',
+                    $this->getAuthId(), $this->isStrict());
 
             $this->bkId = $id;
         }
```

Thoughts? I'm open for a discussions as to why it makes sense as temp fix.

include/class.email.php
MarcBanyard: Just tested this and it needs to as per the recommendation from Peter as the token can be null.

include/class.email.php
MarcBanyard: Just tested this and it needs to as per the recommendation from Peter as the token can also be null.

include/class.email.php
JediKev: Yes update coming soon

MarcBanyard:
> To avoid requiring OAuth2 Plugin version upgrade in tandem with the core release - I'm suggesting we tag the strict mode flag to the backend id as shown below, for now, until we release v1.18.x series.
> 
> ```diff
> diff --git a/include/class.email.php b/include/class.email.php
> index bc4943c6..0a1adea4 100644
> --- a/include/class.email.php
> +++ b/include/class.email.php
> @@ -522,7 +522,8 @@ class EmailAccount extends VerySimpleModel {
>              $id = sprintf('%s:%d',
>                  $this->getAuthBk(), $this->getId());
>              if ($this->isOAuthAuth())
> -                $id .= sprintf(':%d', $this->getAuthId());
> +                $id .= sprintf(':%d:%b',
> +                    $this->getAuthId(), $this->isStrict());
>  
>              $this->bkId = $id;
>          }
> ```
> 
> Thoughts? I'm open for a discussions as to why it makes sense as temp fix.

Why wait till v1.18.x series, as the fix for OAuth2 is required for users now.

I agree moving the **Strict Email Matching** to the core is a good idea, so the above pull (now with the recommended fixes for a null token) resolves the issue along with the minor change required in [osTicket/osTicket-plugins#265](https://github.com/osTicket/osTicket-plugins/pull/265) for the OAuth2 plugin fix the issue for everyone who uses Microsoft 365 with Shared Mailboxes.

The sooner this can be released to the public the better as it is very much needed since Microsoft forcefully disabled basic auth on all tenants.
Even if it comes out as v1.17.3 with a revised version of the plugin as per pull [osTicket/osTicket-plugins#265](https://github.com/osTicket/osTicket-plugins/pull/265) so users can update and get working using OAuth2 and ensure they do not access shared mailboxes with a username and password as this breaches Microsoft's licencing terms.

JediKev:
@MarcBanyard 

You are misinterpreting what he was saying. He's saying in order to avoid us having to make a database patch and force a major upgrade (1.18.x) we can simply add a temporary fix for now to release v1.17.4 and new builds of plugins (with the fixes included) and then worry about a major release later on.

Cheers.

MarcBanyard:
@JediKev & @protich any idea when you plan on releasing this as a temporary fix so the community can start using it?

protich (approved):

PR #6469, opened 2023-02-15
OAuth2 Resource Owner

protich:
This PR adds on changes made by @JediKev on PR #6466 by adding logic to Mail Protocol to support Shared / Aliased
mailboxes.

### DIFF
--- a/include/class.mail.php
+++ b/include/class.mail.php
@@ -325,7 +325,9 @@ namespace osTicket\Mail {
                         throw new Exception('cannot login, user or password wrong');
                     break;
                 case $auth instanceof OAuth2AuthCredentials:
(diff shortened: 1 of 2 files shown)


protich:
PS: OAuth2 Plugin nolonger needs to mess with Resource Owner.

MarcBanyard:
Just tested with the released version of the plugin and it works fine on sending and receiving emails on shared mailboxes.
The only bit that does not show (but I presume it is now no longer required) is the strict email matching.

Can this be pushed and published as the next version so people can start using it along with any possible plugin changes required for this?

protich:
@MarcBanyard - We don't have a specific timeline. but we're in the process of preping it.

protich:
PR rebased on top of `develop` branch.

PR #6476, opened 2023-02-27
ContentAjaxAPI::log conflicts AjaxController::log

valeriop:
When viewing "System logs"  entries details popup window is empty and there is a corresponding php error log :

_PHP Fatal error:  Declaration of ContentAjaxAPI::log($id) must be compatible with ApiController::log($title, $msg, $level = LOG_WARN) in /var/www/osticket/web.1.17.git/include/ajax.content.php on line 22_

class ContentAjaxAPI extends AjaxController {
    function **log(**$id) {

### DIFF
--- a/include/ajax.content.php
+++ b/include/ajax.content.php
@@ -19,7 +19,7 @@ require_once INCLUDE_DIR.'class.ajax.php';
 
 class ContentAjaxAPI extends AjaxController {
 
-    function log($id) {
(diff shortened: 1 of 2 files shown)


JediKev:
@valeriop 

This has already been addressed here:
- https://github.com/osTicket/osTicket/pull/6399

Always do a recursive search before creating a pull. 😉 

Cheers.

PR #6479, opened 2023-03-01
Update class.mailer.php

bholliger:
Inline attachment is added as inline attachment and also as attachment, which should not be the case. Unset did not remove (the correct) attachment


protich:
@bholliger- Thank you for the find and the suggested fix. After reviewing it, I decided to reimplement the fix to avoid looping through attachments. Please see PR #6482.

PR #6493, opened 2023-03-16
issue: Thread Entry Actions z-index

JediKev:
This addresses a small annoyance where the thread entry actions dropdown always appears below the subsequent thread entry. This can be very annoying and even hide some of the options. This adds `z-index` of `auto` to the `thread-entry` div so that the dropdown always appears on top.


dylanrhodes0:
Works beautifully. Thank you.

PR #6494, opened 2023-03-16
oauth2: Strict Matching Bug

JediKev:
This addresses an issue where we were checking strict matching when using basic authentication. This is due to a small oversight in the smtp save checks. This adds a line to check if auth_bk is set to oauth2 before checking strict matching.


makydolon:
Indeed it fixed the HTTP500 error issue! Thank you!

PR #6495, opened 2023-03-20
issue: File Upload Type

JediKev:
This addresses an issue where uploading a file throws an Invalid File error even though the file is valid and there are no file restrictions. This is due to doing `strcasecmp()` on a variable that is some times empty. This adds a check to see if that variable has a value before doing the comparison to avoid false positives. In addition, this adds a new paramter called `$strict` that if true we run the mime type check. By default `$strict` is false which skips the check. We will eventually add a configurable setting to enable strict mime type check if preferred.

### DIFF
--- a/include/class.forms.php
+++ b/include/class.forms.php
@@ -3910,10 +3910,11 @@ class FileUploadField extends FormField {
         return $F;
     }
 
-    static function isValidFile($file) {
-        // Make sure mime type is valid
-        if (strcasecmp(FileObject::mime_type($file['tmp_name']),
-                    $file['type']) !== 0)
+    static function isValidFile($file, $strict=false) {
+        // If strict check mime type
+        if ($strict
+                &&!empty($file['type'])
+                && strcasecmp(FileObject::mime_type($file['tmp_name']), $file['type']) !== 0)
             return false;
 
         // Check invalid image hacks


joeldeteves:
Thanks for this - I was also having this issue. Doing a temp `patch` for now until this gets merged; hopefully we will have it included in the next release

JensEB:
I think, it's not a good choice, to block uploads with this mime type check.
If you upload a bmp file for example, you can have three different mime types. 
image/bmp, image/x-bmp and image/x-ms-bmp.

I uploaded a bmp file and it fails, because $file['type'] was image/bmp and FileObject::mime_type($file['tmp_name']) returned image/x-ms-bmp.
So I changed this check for me, to check only the group of the mime type. I will append the code for it...
However, I think it's better to fix the mime type than just saying the file is invalid...
What do you think?

```
        // don't check mime type here exactly. Check only mime type group
        // for example, there are three mime types for bmp (image/bmp, image/x-bmp and image/x-ms-bmp)
        // so upload will fail
        if (   !empty($file['type'])
            && ($mGroup = substr($file['type'], 0, strpos($file['type'], '/')) )
            && strcasecmp(substr(FileObject::mime_type($file['tmp_name']),0,strlen($mGroup)), $mGroup) !== 0
           ) {
            return false;
        }
```

AVPRiga:
> I think, it's not a good choice, to block uploads with this mime type check. If you upload a bmp file for example, you can have three different mime types. image/bmp, image/x-bmp and image/x-ms-bmp.
> 
> I uploaded a bmp file and it fails, because $file['type'] was image/bmp and FileObject::mime_type($file['tmp_name']) returned image/x-ms-bmp. So I changed this check for me, to check only the group of the mime type. I will append the code for it... However, I think it's better to fix the mime type than just saying the file is invalid... What do you think?
> 
> ```
>         // don't check mime type here exactly. Check only mime type group
>         // for example, there are three mime types for bmp (image/bmp, image/x-bmp and image/x-ms-bmp)
>         // so upload will fail
>         if (   !empty($file['type'])
>             && ($mGroup = substr($file['type'], 0, strpos($file['type'], '/')) )
>             && strcasecmp(substr(FileObject::mime_type($file['tmp_name']),0,strlen($mGroup)), $mGroup) !== 0
>            ) {
>             return false;
>         }
> ```

Thank you! It works for me. I think it happens because we have a lot of different MIME types for one zip or other file type like I had: application/x-7z-compressed and application/zip.
The fix from JediKev, does not work for my case and I also do not think it is necessary to block files if their type are not known, unless there is no security risk.

joeldeteves:
I would also like to see this feature removed as @AVPRiga mentioned. Allowed MIME types should be handled by the web server e.g. NGINX, not by the application.

include/class.forms.php
```diff
@@ -3912,7 +3912,7 @@ function uploadAttachment(&$file) {
 
     static function isValidFile($file) {
```
protich: I think making it optional by default is ideal then we can add an option on Attachments Settings to enable strict mime type check. 

```php
 static function isValidFile($file, $strict = false) { 
        // Check Mime Type
      if ($strict 
           && !empty($file['type']) 
           && strcasecmp(FileObject::mime_type($file['tmp_name']), $file['type']) !== 0)

```

JediKev: Done - waiting for github to update 🙄

PR #6496, opened 2023-03-21
Fix PR #6283: DB column value of "local backend" is 'client' and not 'local'

nicorac:
PR #6283 make it impossible for users to reset their password.
They now receive the error: **"Unable to reset password. Contact your administrator"**

This is because, in table `ost_user` the `backend` field contains `client` and not `local` for local authentication backend...

NOTE: this issue was already pointed out by other users here:
https://github.com/osTicket/osTicket/commit/06e334fa89c755a133023f2afe0ce36ca92dfae3

### DIFF
--- a/pwreset.php
+++ b/pwreset.php
@@ -22,7 +22,7 @@ if($_POST) {
                     $banner = __('Password reset is not enabled for your account. Contact your administrator');
                 }
                 elseif (!$acct->hasPassword()
-                        || (($bk=$acct->backend) && ($bk !== 'local')))
+                        || (($bk=$acct->backend) && ($bk !== 'client')))
                     $banner = __('Unable to reset password. Contact your administrator');
                 elseif ($acct->sendResetEmail()) {
                     $inc = 'pwreset.sent.php';


JensEB:
pull request #6283 should be removed again. 
An agent or user has always a local account where they can change the password. Local accounts are not blocked, if users or agents access osticket via ldap for example. There are only two ways to login now. Users should every time able to change their password.

So I suggest to remove the changes on ./pwreset.php made by #6283 completely and change the changes on ./scp/pwreset.php to:
`
                if (($bk=$staff->getAuthBackend()) && !($bk instanceof `osTicketStaffAuthentication))
                    $msg = __('Unable to reset password. Contact your administrator');
                elseif (!$staff->sendResetEmail())
`

So local accounts for users are always possible and for agents, if their account is not restricted to one method.

TimzOwen:
This works much better to avoid confliction on the agents & Cred stufff

TimzOwen (commented):
Thanks for this. makes it easier for local and client management or Cred

PR #6502, opened 2023-03-29
Set focus to search field on canned response selection

JensEB:
Since jQuery is updated to 3.6.x, there is no focus set to the search field into the select2 dropdown field, where you can search your canned responses. So you have to click in it on every ticket resonse, using canned responses.

This pull request will set the focus again...

### DIFF
--- a/include/client/profile.inc.php
+++ b/include/client/profile.inc.php
@@ -98,3 +98,12 @@ $selected = ($info['lang'] == $l['code']) ? 'selected="selected"' : ''; ?>
         window.location.href='index.php';"/>
 </p>
 </form>
+<script type="text/javascript">
+    // set focus to select2:search_field
+    $('form select').on('select2:open', function (e) {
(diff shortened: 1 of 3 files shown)


JensEB:
I've changed my commit to add focus to other select2 fields like:
- timezone field in agent profile
- timezone field in client profile
- new ticket by an agent, user search field
- ticket number search on ticket merge popup

JediKev:
@JensEB 

Thank you for the pull request, but this is way overkill imo. The auto-focus issue was actually due to a bug in jQuery 3.6.0 and Select2. jQuery 3.7.0 has been released that includes a fix for the issue. You can retest using the following pull:
- https://github.com/osTicket/osTicket/pull/6540

Cheers.

JensEB:
@JediKev 
Yes, I can confirm that your pull request solve the problem.
Cheers

PR #6520, opened 2023-04-17
issue: Duplicate User Copy/Paste Import

JediKev:
This addresses an issue where the Copy/Paste Import feature for Users sometimes creates a duplicate account for an existing User. This occurs when using a string like `User Name, email@domain.tld`, however, does not occur when using a string like `User Name,email@domain.tld`. This is due to the extra space after the comma as we are not trimming extra spaces correctly. This adds `trim()` to the values before they are used so that they correctly match existing values and avoids duplicates.

### DIFF
--- a/include/class.import.php
+++ b/include/class.import.php
@@ -189,7 +189,7 @@ implements Iterator {
             foreach ($this->headers as $h => $label) {
                 $f = $this->fields[$h];
                 $f->_errors = array();
-                $T = $f->parse($csv[$i]);
+                $T = $f->parse(trim($csv[$i]));
                 if ($f->validateEntry($T) && $f->errors())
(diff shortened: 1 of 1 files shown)


include/class.import.php
```diff
@@ -189,7 +189,7 @@ function next() {
             foreach ($this->headers as $h => $label) {
                 $f = $this->fields[$h];
                 $f->_errors = array();
-                $T = $f->parse($csv[$i]);
+                $T = trim($f->parse($csv[$i]));
```
protich: Why not trim the csv content before passing it to the field to parse? 

```php
$T = $f->parse(trim($csv[$i]));
```

My thinking is field might fail parsing - should extra space fail field validation.

PR #6562, opened 2023-06-12
update: README.md PHP & Laminas

JediKev:
This updates the `README.md` file to include the newly supported PHP 8.2 as well as an update to add more verbiage to supported MySQL versions. This also adds credit to `laminas-mail` in the OS projects listings.

PR #6601, opened 2023-09-07
Fix: assigned team gets no alerts if also assigned to agent

empty88:
If a ticket is assigned to a team and to a specific agent at the same time then the team gets no alerts when a client posts a new message to the ticket.

### DIFF
--- a/include/class.ticket.php
+++ b/include/class.ticket.php
@@ -3228,7 +3228,7 @@ implements RestrictedAccess, Threadable, Searchable {
             if ($cfg->alertAssignedONNewMessage() && $ticket->isAssigned()) {
                 if ($staff = $ticket->getStaff())
                     $recipients[] = $staff;
-                elseif ($team = $ticket->getTeam())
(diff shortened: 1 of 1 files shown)


digidigital:
I think the change might not be in line with the documentation and intended use case(?). 
See https://docs.osticket.com/en/latest/Guides/Alerts%20Guide.html

> New Message Alert for Assigned Agent/Team is sent to the assigned agent ***or*** team of agents when a message is appended to a ticket from an end user. This is sent to the agents as long as there is no response from any Agent on the ticket. If there is a response from an agent, only the responding agent will receive the New Message Alert if that alert in this section is enabled.

At least it is not clear if the "or" is meant as "or" or "xor"...

Related: https://github.com/osTicket/osTicket/issues/6659

JediKev:
@digidigital 

Just to update this post as well, this fix was correct and was merged as part of #6562. This is the intended functionality. The documentation simply needs updating.

Cheers.

PR #6602, opened 2023-09-07
Increase maxlength and size for 'match' regex queue filters

thanegill:
30 characters for the maxlength is terribly small for a regular expression.


protich:
Consider also changing regex validator in `TextboxField` to match the change here.

PR #6605, opened 2023-09-19
update: Add mysqli commit wrapper

amiraiman:


protich:
Without any discription it's hard to guess the goal here... what's the use-case since we already set `autocommit` to true by default?

PR #6616, opened 2023-10-11
Update raphael-min.js to 2.3.0

fillorkill:
2.1.0 invokes unsafe-eval error and dashboard graph does not display if CSP turned on. This fixes the issue.


JediKev:
@fillorkill 

Thank you for the submission. It looks like you have 2 commits when there should only be one. Please fix this (by either squashing the two commits or removing one) and update this pull request (by force pushing).

Cheers.

fillorkill:
force pushed. Thank you!

PR #6618, opened 2023-10-16
mail: NoValidateCert

JediKev:
This adds code to disable certificate validation for IMAP/POP3 and SMTP.


Eswar-147:
already same code kept and try not working

PR #6641, opened 2023-11-27
Add try catch block inside mysqli.php real_connect attempt

Adam7288:
Add try catch block around mysqli->real_connect() so multiple db hosts will work in the for loop inside bootstrap.php. Currently, if the first one fails, the script will die with the following error:

PHP Fatal error:  Uncaught mysqli_sql_exception: Connection refused in /support/include/mysqli.php:69


JediKev:
@Adam7288 

Thank you for the pull request. After review we decided to rewrite your pull with #6698. We see what you were going for, however, the try/catch needed to be in `Bootstrap::connect()` where we actually loop through the hosts. With this in mind, I'm closing this pull in favor of the rewrite. 👍🏻 

Cheers.

PR #6676, opened 2024-03-13
Fix ajax path for addRemoteCollaborator

sergejostir:
The plugin instance is in format %s.%s as evident from the following line.

https://github.com/osTicket/osTicket/blob/4689926b2d3d25754f0ddcf8d4e181a2817f6d56/include/class.util.php#L367

As a result we have to allow dot in the URL pattern matching, otherwise there is no match and adding remote collaborators doesn't work.


JediKev:
@sergejostir 

Can you rebase this onto `1.17.x` please?

Cheers.

sergejostir:
@JediKev I've done it.

PR #6684, opened 2024-04-12
issue: fix square characters being printed when printing tickets that's using languages like Thai

amiraiman:
Tested on a fresh install of osTicket version 1.18.1

Before adding the mpdf init options, the Thai characters are printed as square brackets:
image

After:
image

### DIFF
--- a/include/class.pdf.php
+++ b/include/class.pdf.php
@@ -74,7 +74,7 @@ class Ticket2PDF extends mPDFWithLocalImages
         $this->includenotes = $notes;
         $this->includeevents = $events;
 
-	parent::__construct(['mode' => 'utf-8', 'format' => $psize, 'tempDir'=>sys_get_temp_dir()]);
+	      parent::__construct(['mode' => 'utf-8', 'format' => $psize, 'tempDir'=>sys_get_temp_dir(), 'autoLangToFont' => true, 'autoScriptToLang' => true]);
 
         $this->_print();
 	}
@@ -115,7 +115,7 @@ class Task2PDF extends mPDFWithLocalImages {
         $this->task = $task;
         $this->options = $options;
 
-        parent::__construct(['mode' => 'utf-8', 'format' => $this->options['psize'], 'tempDir'=>sys_get_temp_dir()]);
+        parent::__construct(['mode' => 'utf-8', 'format' => $this->options['psize'], 'tempDir'=>sys_get_temp_dir(), 'autoLangToFont' => true, 'autoScriptToLang' => true]);
         $this->_print();
     }
 
--- a/include/class.pdf.php
+++ b/include/class.pdf.php
@@ -74,7 +74,7 @@ class Ticket2PDF extends mPDFWithLocalImages
         $this->includenotes = $notes;
         $this->includeevents = $events;
 
-	      parent::__construct(['mode' => 'utf-8', 'format' => $psize, 'tempDir'=>sys_get_temp_dir(), 'autoLangToFont' => true, 'autoScriptToLang' => true]);
+        parent::__construct(['mode' => 'utf-8', 'format' => $psize, 'tempDir'=>sys_get_temp_dir(), 'autoLangToFont' => true, 'autoScriptToLang' => true]);
 
         $this->_print();
 	}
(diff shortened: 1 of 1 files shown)


JediKev:
@amiraiman 

Thank you for the pull request. I like this approach better as [traditionally we've been updating the core mPDF config directly](https://github.com/osTicket/osTicket/commit/69c5095890717acc4dc84ac9c12bf529b22296fd) which [obviously gets overwritten during mPDF upgrade and forgotten about](https://github.com/osTicket/osTicket/commit/59dc587d46d79061f30da720494c54eadf73461b#diff-eb2964e8950015c256a401a01eed9510bb6f384866fc377ef16986faec603305). The way you are doing it will persist across mPDF upgrades.

With this being said, in this pull you are removing the `$this->autoScriptToLang;` line (further below your main changes) which is not necessary. Also, you are setting `allow_charset_conversion` to `false` whereas when it's `true` it [allows the system to change the character set with `charset=xxx` in the HTML headers](https://github.com/osTicket/osTicket/blob/develop/include/mpdf/vendor/mpdf/mpdf/src/Config/ConfigVariables.php#L86-L87). Additionally, you are forgetting about the poor Tasks! Lastly, since this issue is present in both `1.17.x` and `1.18.x` you'd make the pull against the `1.17.x` branch. That way it gets merged into `1.17.x` and then into `1.18.x`; addressing the issue in both series.

**TLDR;**
In order for this pull to be accepted, you will need to do the following:
- Rebase the entire pull on top of `1.17.x` instead of `develop`. If you don't know how, please close this pull and make a new one.
- Rebase your initial commit to add the `$this->autoScriptToLang;` line back. We don't like extra commits so consolidate your commits as much as possible instead of adding another needless/useless commit.
- Rebase your initial commit to remove the `, 'allow_charset_conversion' => false` part.
- Rebase your initial commit to add `, 'autoLangToFont' => true, 'autoScriptToLang' => true` to the `parent::__construct()` call in the Task2PDF class so that Tasks work as well.

Cheers.

amiraiman:
Hi @JediKev,

I've rebase the PR to target 1.17.x instead of the develop branch as suggested. I've also added the line `$this->autoScriptToLang;` back, remove `allow_charset_conversion` option as well as added the same mPDF init config to Task2PDF Class. I've tested this new changes for both ticket printing & task printing on 1.17 and 1.18, and both works as intended.

For some context, I added the `allow_charset_conversion` because the rendering of Thai characters still does not work when I tested them initially. However, I did another fresh install of osTicket v1.18 and only add `autoScriptToLang` and `autoLangToFont` and the PDF renders correctly.

With that, I'm wondering why you would want the `$this->autoScriptToLang;` line back since we already set it to true during the initialization? Does it do different things? I only found [this](https://mpdf.github.io/reference/mpdf-variables/autoscripttolang.html) on mPDF docs and can't find any usage info on it.

Lastly, would it be better to move mPDF construct to `mPDFWithLocalImages` class since both `Ticket2PDF` and `Task2PDF` extends from it?

JediKev:
@amiraiman 

Thank you for the updates. So after some research I found that that line was needed to replace a deprecated or, quote "redundant" [function SetAutoFont()](https://mpdf.github.io/installation-setup/installation-v6-x.html) (search for SetAutoFont) when [upgrading from 5.x to 6.x](https://github.com/osTicket/osTicket/commit/9581280e2e6a73e52aa8ae85162d709e9d5eec90). But, alas, there is a twist. The line really needs to be `$mpdf->autoScriptToLang = true;` in both places (as recommended in the link above). If you want to update those lines that'd be swell.

Also, you have weird extra spaces on the Ticket2PDF change. If you could rebase to remove those spaces that'd be great as well.

Oh, to answer your question about moving it to the parent class; it doesn’t currently have a __construct method and since we already define things in each class in the parent::__construct call you can just add it there like you already have it. This codebase is in maintenance mode so it’s fine; all of this is currently being rewritten with v2.0.

Cheers.

amiraiman:
Hi @JediKev

I've updated `$this->autoScriptToLang` to be set to true instead of merely calling the property as suggested.

Sorry for the extra two spaces, I disabled my IDE PHP auto formatter (Pint) because it would reformat the entire file, so I missed the extra spaces. Instead of removing all spaces, I removed two to ensure that the construct call is at the same indentation level as the rest of the method body.

PR #6694, opened 2024-05-13
mailer: sendmail() Failure

JediKev:
This addresses an issue where passing a full From string (eg. `"osTicket Alerts"<user@noemail.com>`) as the `$from` parameter for `Mailer::sendmail()` causes a fatal laminas-mail error. This is due to the full string not being a valid email address. This changes the full From string to just the email address in the needed places. In addition this adds an array of `$options` with the From Name as the `from_name` option which gets used to format the full From string later down the pipeline.


JediKev:
Refactor (make sure to add space in croak)

PR #6699, opened 2024-05-14
i18n: Schedules getDays()

JediKev:
This addresses an issue where using a non-english language breaks the Ticket view page for Tickets with no Due Date, SLA with grace period longer than 41 hours, and the SLA is attached to a Schedule with working days. This is due to mistakenly translating the day names in `getDays()` function at times we shouldn't be. PHP only likes English day names during actual date calculations. This adds a new parameter to `getDays()` called `$translate` that defaults to `false`. If `true` we will translate the day names, otherwise they will remain in English.

### DIFF
--- a/include/class.schedule.php
+++ b/include/class.schedule.php
@@ -1049,13 +1049,15 @@ class ScheduleEntry extends VerySimpleModel {
     }
 
     static function getDays() {
-        static $translated = false;
-        if (!$translated) {
-            foreach (static::$days as $k=>$v)
-                static::$days[$k] = __($v);
-        }
+        return self::$days;
+    }
 
-        return static::$days;
+    static function getTranslatedDays() {
+        static $translated = null;
+        if (!isset($translated))
+            $translated = array_map(function ($d) { return __($d); }, static::getDays());
+
+        return $translated;
     }
 
     static function getWeeks() {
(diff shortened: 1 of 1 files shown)


include/class.schedule.php
```diff
@@ -1048,12 +1048,10 @@ static function getFrequencies() {
         return static::$frequencies;
     }
 
-    static function getDays() {
-        static $translated = false;
-        if (!$translated) {
+    static function getDays($translate=false) {
```
protich: The logic here is porked a little... if you call `getDays(true)` first then subsequently call `getDays()` - you'll get translated days because we're overriding the propery (`static::$days`) with translated values. 

I'll recommend using array_map to make a copy and translate on the fly. 

```php
return $translate ? array_map(function ($d) { return __($d); }, static::$days) : static::$days;

```

include/class.schedule.php
```diff
@@ -1176,7 +1174,7 @@ function buildFields() {
                     'default' => "",
                     'layout' => new GridFluidCell(6),
                     'label' => __('Day of the Week'),
-                    'choices' => ScheduleEntry::getDays() + array(
+                    'choices' => ScheduleEntry::getDays(true) + array(
```
protich: Add `getTranslatedDays()` routine that calls `getDays` to make it absolutely clear what we're getting back. 

**PS**: You can even cache translated days in the routine - double win!

protich (changes requested):

include/class.schedule.php
```diff
@@ -1049,15 +1049,17 @@ static function getFrequencies() {
...
-                static::$days[$k] = __($v);
-        }
-
         return static::$days;
     }
 
+    static function getTranslatedDays() {
+        static $translated = null;
+        if (!isset($translated))
+            $translated = array_map(function ($d) { return __($d); }, static::getDays());
```
protich, line 1058: Looks good - but I think `static::getDays()` should be `self::getDays()`. It really doesn't matter in this context but it might down the road.

JediKev: done

protich (approved):

PR #6700, opened 2024-05-15
issue: Task Last Respondent

JediKev:
This addresses an issue where `Task::getLastRespondent()` returns the wrong Agent leading to the wrong people receiving alerts, etc. This updates the `Task::getLastRespondent()` function to make it limit the last respondent search to the Task's own Thread Entries rather than searching through all entries.


protich (approved):
Looks good!

PR #6702, opened 2024-05-20
issue: Referred Closed Tickets

JediKev:
This addresses an issue where Agents cannot see Referred Tickets in their Closed Queues. This is due to the `Staff::getTicketsVisibility()` function blindly using the `open` state instead of looking for `closed` as well. This updates the function to include the `closed` state in the status condition. This ensures the referred Tickets are shown in every Queue.

### DIFF
--- a/include/class.staff.php
+++ b/include/class.staff.php
@@ -713,7 +713,7 @@ implements AuthenticatedUser, EmailContact, TemplateVariable, Searchable {
                 'child_thread__referrals__team__team_id__in' => $teams)));
             $assigned->add($childRefTeam);
         }
-        $visibility = Q::any(new Q(array('status__state'=>'open', $assigned)));
+        $visibility = Q::any(new Q(array('status__state__in'=>['open', 'closed'], $assigned)));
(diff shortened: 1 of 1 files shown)


protich:
I don't think this was a bug... just like we don't have a concept of showing assigned tickets in every queue.

hazuremon:
Thanks @JediKev!

@protich The issue comes from a query I made on the support forum. Limited access agents are unable to see tickets they have closed in the closed queues, even when the "auto-referrer tickets on close" option is selected.

I'm not sure if this is the expected behavior, but it results in an issue where agents with limited access can only access the closed ticket through the ticket URL but cannot view closed tickets in the closed queue.

https://forum.osticket.com/d/103698-auto-refer-on-close-problem/5

c0nfus3d:
We have limited access agents, and they need to be able to go back and view their closed tickets. I understand they can view them if they have the ticket URL, but how are they supposed to know that off the top of their head? The ability for them to list and view their closed tickets is something we definitely need-- as they could have notes or other documentation they need to look back on.

PR #6708, opened 2024-05-25
Fix agent details not clearing on change

sergejostir:
If you use for example LDAP plugin, the agent details are automatically filled in if you select a username. However, the fields are not cleared in case the next selection is missing some of the values. Instead the details of the previous selection are left behind.

This PR ensures that the details of the previous person will be cleared in case the next person is missing some of the details.

### DIFF
--- a/scp/js/scp.js
+++ b/scp/js/scp.js
@@ -333,7 +333,7 @@ var scp_prep = function() {
         onselect: function (obj) {
             var fObj=$('.staff-username.typeahead').closest('form');
             $.each(['first','last','email','phone','mobile'], function(i,k) {
-                if (obj[k]) $('.auto.'+k, fObj).val(obj[k]);
+                $('.auto.'+k, fObj).val(obj[k] ? obj[k] : '');
             });
         },
         property: "username"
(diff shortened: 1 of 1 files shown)


JediKev:
@sergejostir 

Why not do something like `fObj.reset();` before filling in the form fields (ie. before the `$.each` call)? This way we can keep the form field whitelisting that the `if (obj[k])` check is doing.

Cheers.

sergejostir:
What do you mean by that?

JediKev:
@sergejostir 

The issue is the form not clearing old field values on typeahead selection, correct? So instead of removing the if statement, you could instead simply clear the form fields of their old values before filling in the new field values. If you look above the `$.each()` call the form is set to `var fObj`. So to clear the form you can simply call `fObj.reset();` above the `$.each()` call.

If you remove the if statement there could be potential for abuse like injection attempts, etc.

Cheers.

sergejostir:
The issue is not the whole form clearing, but the details that might not exist in the currently selected object not clearing.

For example, let's say we use LDAP plugin and we have two AD objects, the first one has phone number set and the second does not.
When you select the first object using the typeahead it will fill the details in the fields including the phone number. Then you change your mind and select the second object. The fields will be updated, but the phone number will remain from the first object because of the if clause.

I updated the code. Now it works the same as before minus the bug.

PR #6709, opened 2024-05-28
patches: Latest 05/28/2024

JediKev:
This is a pull containing recent small patches we've aggregated since last release. This includes the following:
- Adding an `isset()` check to ensure the `ticket` array key exists within the `$current` array and is set before using it.
- Adding a null coalescing operator after `$obj->getPermission()->getInfo()` in `Export::agents()` so that if the permission lookup fails or returns invalid data we will use an empty array to avoid fatal errors and let the export continue.

### DIFF
--- a/include/class.api.php
+++ b/include/class.api.php
@@ -387,8 +387,8 @@ class ApiXmlDataParser extends XmlDataParser {
     function fixup($current) {
         global $cfg;
 
-		if($current['ticket'])
-			$current = $current['ticket'];
+        if (isset($current['ticket']))
+            $current = $current['ticket'];
 
         if (!is_array($current))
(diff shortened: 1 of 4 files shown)


protich (commented):
Looks good.

JediKev:
change the `->getInfo() ?? []` to:
```diff
diff --git a/include/class.role.php b/include/class.role.php
index f2d07057..68187694 100644
--- a/include/class.role.php
+++ b/include/class.role.php
@@ -322,7 +322,7 @@ class RolePermission {
     }
 
     function getInfo() {
-        return $this->perms;
+        return $this->perms ?: [];
     }
 
     function merge($perms) {
```

PR #6724, opened 2024-06-18
issue: mimeDecode .eml Attachments

JediKev:
This reapplies a missing patch `af1c4a6` that added `message/rfc822` attachment support.


include/pear/Mail/mimeDecode.php
```diff
@@ -369,6 +369,7 @@ function _decode($headers, $body, $default_ctype = 'text/plain')
 						$return->body = ($this->_decode_bodies ? $this->_decodeBody($body, $encoding) : $body);
 					}
                     $obj = new Mail_mimeDecode($body);
+                    $return->body = $body;
```
protich: shoule we say if body is not set? This is because body can potentially be set in like `369`.
```php
if (!isset($return->body))
    $return->body = $body;
```

PR #6733, opened 2024-07-15
issue: Multiselect List Export

JediKev:
This addresses an issue where attempting to export a multiselection List Field with multiple selected items only shows the first selected item. This is due to the Item IDs containing spaces (due to the earlier `explode()` call) so the IDs do not match the available IDs. This adds a `trim()` around the Item IDs so that they match appropriately.

### DIFF
--- a/include/class.dynamic_forms.php
+++ b/include/class.dynamic_forms.php
@@ -1633,6 +1633,7 @@ class SelectionField extends FormField {
             $values = array();
             $choices = $this->getChoices();
             foreach (explode(',', $value) as $V) {
+                $V = trim($V);
                 if (isset($choices[$V]))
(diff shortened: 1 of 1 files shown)


include/class.dynamic_forms.php
```diff
@@ -1633,6 +1633,7 @@ function to_php($value, $id=false) {
             $values = array();
             $choices = $this->getChoices();
             foreach (explode(',', $value) as $V) {
+                $V = trim($V);
```
protich, line 1636: This works but we could do a one liner as shown below. Just fyi
```php
$values = array_intersect_key($this->getChoices(), array_flip(array_map('trim', explode(',', $value))));
```

Issue #6734, opened 2024-07-18
[Feature Request] Provide more Signals API endpoints for frontend development


changyy:
### Prerequisites

* [X] Can you reproduce the problem in a fresh installation of the "develop" branch?
* [X] Do you have any errors in the PHP error log, or javascript console?
* [X] Did you check the [osTicket forums](https://forum.osticket.com/)?
* [X] Did you [perform a cursory search](https://github.com/osTicket/osTicket/issues) to see if your bug or enhancement is already reported?

For more information on how to write a good [bug report](https://github.com/osTicket/osTicket/wiki/Github-Issues-Guidelines)

### Description

[Description of the bug or feature]

I tried to understand the osTicket plugin framework by studying those documents:

- https://docs.osticket.com/en/latest/Admin/Manage/Plugins.html
- https://github.com/osTicket/osTicket-plugins
- https://github.com/osTicket/osTicket/blob/develop/setup/doc/signals.md

notes:

- https://blog.changyy.org/2024/07/php-osticket-plugin-study-and-an-adding-extra-frontend-resources-plugin.html
  - https://blog-changyy-org.translate.goog/2024/07/php-osticket-plugin-study-and-an-adding-extra-frontend-resources-plugin.html?_x_tr_sl=zh-TW&_x_tr_tl=en&_x_tr_hl=zh-TW&_x_tr_pto=wapp

I'm trying to build a plugin to change the UI flow, like add a new button via js without changing server code. After I found the `$ost->addExtraHeader()` method, I cannot find a suitable event to hook more frontend resources.

Finally, I added a new one: `Signal::send('staff.header.extra', null);`

```
% git diff
diff --git a/include/staff/header.inc.php b/include/staff/header.inc.php
index 9ac217c5..9de3494c 100644
--- a/include/staff/header.inc.php
+++ b/include/staff/header.inc.php
@@ -55,6 +55,7 @@ if (osTicket::is_ie())
     <link rel="icon" type="image/png" href="<?php echo ROOT_PATH ?>images/oscar-favicon-16x16.png" sizes="16x16" />
 
     <?php
+    Signal::send('staff.header.extra', null);
     if($ost && ($headers=$ost->getExtraHeaders())) {
         echo "\n\t".implode("\n\t", $headers)."\n";
     }
```

Is there any Signal APIs I'm missing that would be enough to hook into more js/css resources?

### Steps to Reproduce

https://github.com/changyy/osTicket-plugins-Extra-Frontend-Resources

[...]

PR #6735, opened 2024-07-18
[Feature] Added signal API for frontend development on osticket plugin framework

changyy:
- Signal::send('staff.header.extra', null);
- Signal::send('client.header.extra', null);

ref: 

- #6734


Cm-8:
I find this addition useful for my purposes too.

Cm-8:
up!

PR #6738, opened 2024-08-06
Users directory: Add configuration option to show different user fields in users list view

ywarnier:
Allows for greater customization if the usage type does not match the typical OSTicket usage type.

Defining `define('USERS_COLUMNS', 'cdata__firstname|First Name:cdata__mobilephone|Mobile Number');` will automatically add the "First Name" and "Mobile Number" columns to the users repository list view (scp/users.php) and make it much easier to distinguish between homonymous users.

This PR does *not* provide search or sorting support for the added columns.

I tried to match the coding style of the existing code as much as possible.

There are possible optimisations in the way the user__cdata is loaded (currently one `$users->values($extra_column);` call for each additional field), but looking into it, I saw I would have had to either use the otherwise-apparently-unused-UserCdata class, or develop a new method for the User class which would provide a new way to query, and that seemed like a bad idea to go that far for a first contrib.

### DIFF
--- a/include/ost-sampleconfig.php
+++ b/include/ost-sampleconfig.php
@@ -53,6 +53,22 @@ define('DBPASS','%CONFIG-DBPASS');
 # Table prefix
 define('TABLE_PREFIX','%CONFIG-PREFIX');
 
+# Option: USERS_COLUMS
+#
+# Values: string of column-separated field|title from the ost_user__cdata
+#         table that we want to display additional to the default fields
+#         in the users directory list view.
+#         Check the columns in your ost_user__cdata table and use the
+#         field name 'cdata__'+columnname in the list below.
+#         If a language term is available (see
+#         include/i18n/[language_code]/LC_MESSAGES/messages.mo.php)
+#         you can use it after the field name with a | separator, like;
+#           'cdata__firstname|First Name:cdata__email2|Email'
+#         If no language term is available, just leave empty or put a
+#         term that should work for most of your agents.
+#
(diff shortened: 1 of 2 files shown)


JediKev:
@ywarnier 

Thank you for the first-time contribution. I would like to note that the current osTicket codebase (what we call legacy) is in maintenance mode while we focus on v2.0 development. This means the current codebase is only receiving maintenance and security releases. This also means we are not accepting any new features to the legacy codebase. With this being said, you can keep this pull open if you'd like. However, imo, it'd be better to fork osTicket and make your mod available in your own fork so you can keep track of issues, etc. separately as you cannot use this repo for such activities.

Cheers.

ywarnier:
Hi @JediKev 
Thanks for the heads up. I have a separate fork with that, so no problem. Although your answer brings new questions, if you've got a minute to educate the uneducated:
- What is the branch you use for the development of v2.0?
- Is it public?
- What is the planned schedule (roughly, like in a month, in half a year, in a year) for v2.0? (I'm setting up a new system now, so it might be useful, decision-wise, to have a rough idea)

I'm happy to ask this in a discussion topic if you like, but this is a follow-up to your answer.

JediKev:
@ywarnier 

1. Currently closed source so no public repo yet. Once it’s ready to be public the repo will be made public.
2. Same as above.
3. No set timeline at this time however you can follow the [public roadmap here](https://github.com/osTicket/roadmap). We admittedly don't update it often but we will eventually get around to updating it; we are more focused on the actual development.

Cheers.

ywarnier:
Thanks. Good luck!

PR #6766, opened 2024-11-18
patches: Latest 11/18/2024

JediKev:
This includes patches for multiple things including:
- Addressing an issue where Guest User sessions would quickly timeout even though the User session was configured for say 24 hours.
- Removing a random `print $emails` from the `emailtest.php` page.

### DIFF
--- a/include/class.ostsession.php
+++ b/include/class.ostsession.php
@@ -27,9 +27,8 @@ class osTicketSession {
         // session name/ssid
         if ($name && strcmp($this->name, $name))
             $this->name = $name;
-        // Session ttl cannot exceed php.ini maxlifetime setting
         $maxlife =  ini_get('session.gc_maxlifetime');
-        $this->ttl = min($ttl ?: ($maxlife ?: SESSION_TTL), $maxlife);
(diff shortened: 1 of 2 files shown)


protich:
@JediKev - Let's have a chat about this pull request. Especially about this comment above the change  `// Session ttl cannot exceed php.ini maxlifetime setting` and also we need to make sure we don't confuse or rather conflate session settings for logged-in user vs guest.

include/class.ostsession.php
```diff
@@ -29,7 +29,7 @@ function __construct($name, $ttl = SESSION_TTL, $checkdbversion = false) {
             $this->name = $name;
         // Session ttl cannot exceed php.ini maxlifetime setting
         $maxlife =  ini_get('session.gc_maxlifetime');
-        $this->ttl = min($ttl ?: ($maxlife ?: SESSION_TTL), $maxlife);
```
JediKev, line 32: remove comment

iceven00 (commented):
still getting login issues after reverting from this patch

PR #6773, opened 2024-12-09
Fix force_https auto-enabling on settings page

remram44:
If `force_https` is `off`, the checkbox on the settings page still shows up checked, and the settings will be enabled when you save the settings. Even if you haven't touched it.

This causes infinite redirect loops if using a reverse proxy to terminate TLS.

### DIFF
--- a/include/staff/settings-system.inc.php
+++ b/include/staff/settings-system.inc.php
@@ -65,7 +65,7 @@ $gmtime = Misc::gmtime();
             <td><?php echo __('Force HTTPS'); ?>:</td>
             <td>
                 <input type="checkbox" name="force_https" <?php
-                echo $config['force_https'] ? 'checked="checked"' : ''; ?>>
+                echo ($config['force_https'] == "on") ? 'checked="checked"' : ''; ?>>
                 <?php echo __('Force all requests through HTTPS.'); ?>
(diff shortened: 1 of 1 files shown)


JediKev:
@remram44 

Thank you for the pull request. While I can't replicate this issue I can see your point and I see no harm in this pull. With this being said, please rebase this pull on top of `1.17.x` as this issue would be the same in both `1.17.x` and `1.18.x`.

Cheers.

remram44:
Ok, rebased.

> I can't replicate this issue

Is your `force_https` setting set to something other than `off`?

JediKev:
@remram44 

Yes, if enabled it gets set to `'on'` and if disabled it gets set to an empty string `''`. You can see this from [the code here](https://github.com/osTicket/osTicket/blob/4689926b2d3d25754f0ddcf8d4e181a2817f6d56/include/class.config.php#L1326).

Cheers.

remram44:
I see! What probably happened is that the first time I enabled this, I couldn't reach osTicket at all, so I changed it from `on` to `off` in the database directly. Now because it's `off`, the checkbox is checked.

I think this patch should still go in to match the checkbox rendering to how the configuration is actually used, but I can see why no one else runs into this.

PR #6775, opened 2024-12-16
issue: hide delete link for advanced search queues

nicoletta-maia:
This hides the delete link for advanced search queues that have not yet been saved.
To reproduce the error, try performing an advanced search and afterwards try deleting the queue by clicking on the Delete item in the dropdown that appears to the right of the queue title


protich:
While I agree with the general concept here due to blank popup - why not simply clear (remove) the cached ad-hock search from the session on "delete".  I say so becuase the only way to remove such searches, at the moment, is to logout and log back in!

Issue #6772, opened 2024-12-08
Department visibility restrictions can be overridden in canned responses


ml247:
When viewing the list of canned responses at /scp/canned.php you will NOT see any canned responses that are in a department to which you have no access.

However you CAN still see such canned responses directly by entering the canned response ID number into the URL. E.g.:
/scp/canned.php?id=11

If I did not have departmental access to canned response 11, I will still be able to view it by entering the URL above.

This is in version 1.18.1

PR #6797, opened 2025-02-03
issue: Canned Response Access

JediKev:
This addresses issue #6772 where you can access a Canned Response with a Department set even if you don’t have the Premade Role Permission for the Canned Response's Department. This is due to not checking access before displaying the Canned Response. This adds a new function called `staffCanAccess()` to class Canned that checks to see if the Canned Response has a Department and if so ensures the Agent has a Role in that Department and that the Role contains the Premade Permission. This adds a new `if()` statement that calls `staffCanAccess()` if a Canned Response is found.

### DIFF
--- a/include/class.canned.php
+++ b/include/class.canned.php
@@ -311,6 +311,12 @@ extends VerySimpleModel {
 
         return true;
     }
+
+    function staffCanAccess($staff) {
+        if (!$staff instanceof Staff)
+            return false;
+        return (!$this->dept || (($role = $staff->getRole($this->dept)) && $role->hasPerm(Canned::PERM_MANAGE)));
+    }
 }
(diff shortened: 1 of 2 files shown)


include/class.canned.php
```diff
@@ -311,6 +311,12 @@ function update($vars,&$errors) {
 
         return true;
     }
+
+    function staffCanAccess($staff) {
+        if (!$staff instanceof Staff)
```
JediKev, line 316: I think I'll change this to:
```php
    function staffCanAccess(Staff $staff) {
        return (!$this->dept || (($role = $staff->getRole($this->dept)) && $role->hasPerm(Canned::PERM_MANAGE)));
    }

Issue #6817, opened 2025-04-02
Issue with Custom Regular Expressions Field


nandi96:
In the Custom Regular Expressions field, when I enter the | character, the system automatically removes it. The | character is necessary because it represents the "OR" operator in regex, allowing multiple conditions to be specified within a single expression. Where can we find this parameter in the source code?
osTicket v1.18.2 (53339df) + Apache + MySQL 10.4.34 + PHP 8.1.32

JediKev:
@nandi96 

We definitely allow `|` in regex. Can you provide an example with screenshots, etc.?

Cheers.

JediKev:
@nandi96 

Also, I just noticed that you are running PHP 8.1 but the requirements for v1.18.2 is PHP 8.2-8.4. So I would recommend that you upgrade PHP to 8.2-8.4 (8.4 is recommended) before retesting.

Cheers.

nandi96:
> [@nandi96](https://github.com/nandi96)
> 
> Also, I just noticed that you are running PHP 8.1 but the requirements for v1.18.2 is PHP 8.2-8.4. So I would recommend that you upgrade PHP to 8.2-8.4 (8.4 is recommended) before retesting.
> 
> Cheers.

Deal, now we installed a clean instance of ost 1.18.2 with php 8.4 and retested, same issue.
Screen rec: https://streamable.com/yeh1yr
@JediKev

JediKev:
@nandi96 

I'm sorry, but I don't visit 3rd party links. Please post screenshots/videos directly on this site.

Cheers.

JediKev:
@nandi96 

You were right, i was able to replicate this. Here is the fix:
- https://github.com/osTicket/osTicket/pull/6818

Thanks for the report! This will be included in next release.

Cheers.

PR #6818, opened 2025-04-02
issue: Regex OR Operator

JediKev:
This addresses issue #6817 where when using an OR operator in the Regular Expression validation textbox for a Short Answer field it gets removed without warning. This is due to a malformed Regex statement within the `Format::strip_emoticons()` function. This simply removes the brackets from the OR comparison as brackets are only used for range comparisons (eg. `/[\x{2310}-\x{231F}]/u`).


f3rrix:
Cool, thanks.  You're still checking for copyright twice tho :)

PR #6826, opened 2025-04-17
Remove regex bug that deletes pipe characters "|" from all thread entries

f3rrix:
In JediKev's b843fb15030c2566af415cda049be30a25ab4eb2 the regex was '/[\x{00B0}|\x{00A9}]/u' was added, with the intention of removing degrees -or -copyright.  But this bracketed section is a character class, not a sub-expression. So what it actually does is remove any of: degrees, |, or copyright.

Since an earlier line already removes copyright (A9) I just updated the new expression to be a character class of one. 

This could be even more concise as just: /\x{00B0}/u


JediKev:
@f3rrix 

I already have a pull for this 😅 #6818

Thanks for the effort though! Cheers.

f3rrix:
woops I commented in the wrong place. Deleted and pasted here:  Cool thanks. You're still checking for copyright twice tho.

PR #6840, opened 2025-05-22
Update class.mailparse.php

polarismail:
// We want to avoid joining the body of attached html e-mails. We check for attachments and don't recurse into rfc822 ones // This fixed the case where when someone forwards an e-mail with an html attachment, it includes the attachment's body in the ticket

### DIFF
--- a/include/class.mailparse.php
+++ b/include/class.mailparse.php
@@ -351,34 +351,56 @@ class Mail_Parse {
         return null;
     }
 
-    function getBody(){
-        global $cfg;
+// We want to avoid joining the body of attached html e-mails. We check for attachments and don't recurse into rfc822 ones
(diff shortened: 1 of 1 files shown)


polarismail:
// We want to avoid joining the body of attached html e-mails. We check for attachments and don't recurse into rfc822 ones
// This fixed the case where when someone forwards an e-mail with an html attachment, it includes the attachment's body in the ticket

CKeckNorthIO:
Good idea! But I'd suggest to make that configurable through a checkbox in "Admin Panel -> Email -> Settings". I'm sure there are users who like to have it the way it is now. Maybe you could update your PR with needed changes in emailsettings.php?
Just my 2 cents.

PR #6869, opened 2025-10-20
Upgrade mpdf to resolve PHP8.4 deprecations

jamieburchell:


JediKev:
@jamieburchell 

Thank you for the pull. Apparently `v8.2.6` is not the latest anymore so I made a new one for `v.8.2.7` here: #6897

This pull is also on top of `1.17.x` instead of `develop` as both 1.17.x and 1.18.x need these changes. Closing this in favor of the new pull.

Cheers.

PR #6884, opened 2025-12-09
Add docker containerezation 

Aeliot-Tm:
**What is done:**
- Added building & running with `docker compose up`.
- Flexible configuration of docker services both via environment variables and native override-file (`compose.override.yaml`).
- Added example of `compose.override.yaml` with mailhog for the catching of outgoing emails during local development.
- Added easy installation of [system plugins](https://github.com/osTicket/osTicket-plugins) by one environment variable `INSTALL_SYS_PLUGINS`.
- Added native support of environment variables by config.
- Added instructions for local development with docker.


Aeliot-Tm:
More complex solution then #5655

PR #6920, opened 2026-01-22
Fix session timeout unit mismatch causing auth loop.

ramalhoruben:
StaffSession and ClientSession populate maxidletime from config values expressed in minutes, but UserSession::isValidSession() compares it against (time() - expire) in seconds. 

```
        // is it expired??
        if ($maxidletime && ((time()-$expire) > $maxidletime))
            return false;
```

This unit mismatch leads to repeated re-authentication loops (notably in scp/login.php), until either user clear session data (cookies) from its browser or the condition that leads to this bug change, often through additional waiting time for the condition to lift.

Normalized maxidletime to seconds before comparison to ensure correct session expiration behavior.

### DIFF
--- a/include/class.usersession.php
+++ b/include/class.usersession.php
@@ -216,7 +216,7 @@ class ClientSession extends EndUser {
         // XXX: Change the key to user-id
         $this->session = new UserSession($user->getUserId());
         $this->setSessionToken();
-        $this->maxidletime = $cfg->getClientTimeout();
+        $this->maxidletime = $cfg->getClientTimeout()*60;
     }
 
     function getSessionUser() {
@@ -237,7 +237,7 @@ class StaffSession extends Staff {
             $staff->class = 'staff';
             $staff->session = new UserSession($staff->getId());
             $staff->setSessionToken();
-            $staff->maxidletime = $cfg->getStaffTimeout();
+            $staff->maxidletime = $cfg->getStaffTimeout()*60;
             $staff->checkip = $cfg->enableStaffIPBinding();
         }
         return $staff;


MicroframeDev1:
We tried to add this fix to our self-hosted codebase.  Claude believes the fix is incorrect for develop (as of 2026-04-23) and branch 1.18.3:

----------------------
● Thanks for digging into the session timeout issue — the symptom you're describing (repeated re-authentication loops)
    is real, but I believe the fix doubles an existing conversion rather than adding a missing one.
  
   The call chain in both develop and v1.18.3:
   // class.config.php
   function getStaffTimeout() {
       return $this->getStaffSessionTimeout();
   }
   function getStaffSessionTimeout() {
       return $this->get('staff_session_timeout') * 60;  // already returns seconds
   }
  
   getStaffTimeout() (and getClientTimeout()) already return seconds, not minutes. The * 60 conversion lives inside
   getStaffSessionTimeout() / getClientSessionTimeout(). Adding another * 60 in class.usersession.php would produce
   timeouts 60× too long — a configured 30-minute idle limit would become ~30 hours.
  
   The actual root cause of the re-authentication loop we encountered turned out to be in class.ostsession.php:
   DatabaseSessionRecord::lookupRecord() resets session_data to '' for any session where session_expire < NOW(), but
   writes don't refresh session_expire. This creates a cycle where the session's CSRF token (and user token) never
   persist across requests, causing every login attempt to appear invalid. The fix is to force
   session_regenerate_id(true) on the login GET when no session data exists, so the browser receives a fresh
   non-expired session before the CSRF token is written.
---------------------

I will be posting a PR against "develop" with the session and CSRF fixes.

MicroframeDev1:
Possibly related: https://github.com/osTicket/osTicket/pull/6949

Issue #6931, opened 2026-02-13
CSRFToken error fix in client ajax.php


liquid2026:
There is an issue, when i insert an image inside a new ticket on the client side. Because the upload does not set the header X-CSRFToken the ckeck will fail and redirect with 302 to index.php

Possible fix:
- Always insert the header inside redactor.min.js to prevent future conflicts
- I added `;this.p.headers["X-CSRFToken"]=document.querySelector('input[name="__CSRFToken__"]').value` right after `(this.p.url=this.p.data?this.p.url+"?"+this.p.data:this.p.url)` as a quick and dirty fix.

Sorry for not providing more informations at the moment.

JediKev:
@liquid2026 

Thank you for the report. Here is the proper fix:
- https://github.com/osTicket/osTicket/pull/6932

This issue only recently cropped up as we hardened the request handling so it now needs to be proper.

Cheers.

bbjay:
I just ran into this on v1.18.3
The editor tries to render an image with
`<img src="URL not supported">`

I can confirm #6932 is resolving this.
So why is this issue closed if the resolving PR isn't merged yet?

JediKev:
@bbjay 

Because the fix is linked here and it will be merged in the next release..

Cheers.

PR #6932, opened 2026-02-13
issue: Inline Image Upload CSRF

JediKev:
This addresses issue #6931 where uploading inline image on client portal fails. This is due to missing CSRF Token. During replication I noticed the key that set the CSRF Token in the upload data was `imageUploadData` but the real key is `imageData`. This simply updates the key name to address the issue. The issue only recently cropped up as we hardened the request handling.

### DIFF
--- a/js/redactor-osticket.js
+++ b/js/redactor-osticket.js
@@ -386,7 +386,7 @@ $(function() {
                 'imageCaption': false,
                 'imageManagerJson': 'ajax.php/draft/images/browse',
                 'imagePosition': true,
-                'imageUploadData': {
+                'imageData': {
                     '__CSRFToken__': $("meta[name=csrf_token]").attr("content")
                 },
                 'imageResizable': true,


TheWebMachine:
I attempted to apply this fix manually and it didn't work work me. Still can't upload images as Agent or Client to tickets or to Knowledgebase. Still getting CSRF token errors.

JediKev:
@TheWebMachine 

The change is to a javascript file. So, you have to clear the site's cache in the browser to clear the local file cache. You may also need to restart the webserver and such to ensure any file cache is cleared on the server. Then you can reload the site and it should be resolved. You can also try an incognito/private window for a quick test.

Cheers.

TheWebMachine:
I restarted web services, cleared browser caches, did the Incognito window. 

In the end, I had to fully reboot the web server to get it working. Not sure why cuz it's just basic apache2 on Debian with no Redis or memcache, but yeah. Looks fine now. Pardon the bother!

How did this fix not make it into the June release in time!? Is it really affecting that *few* users that nobody thought to merge it? I noticed it immediately upon upgrading and it took me forever to find this fix. 😔

JediKev:
@TheWebMachine 

I touched on it in the forum somewhere but basically, the last release was an accelerated security release and I didn't have time to pull everything in. It will in-fact be included in the next release though. Stay tuned!

Cheers.

PR #6950, opened 2026-04-23
Fix passwd1 maxlength lost on reset token flow

### LINKED ISSUE #6940: Password reset dialog have different maxlength for new and confirm inputs
When using a password reset link the dialog that prompts for new password have different maxlength for the "new" (30 characters) and "confirm" (128 characters) fields. This is likely due to the code overwriting the passwd1->configuration field instead of updating it: https://github.com/osTicket/osTicket/blob/2570d69892775fff5df40a7540e507131a0975ad/include/class.staff.php#L1611

https://github.com/osTicket/osTicket/blob/2570d69892775fff5df40a7540e507131a0975ad/include/class.staff.php#L1574-L1589

Noticed in v1.18.2 and confirmed in both the 1.18.x and develop branch

royont123:
Fixes #6940

When using a password reset link, the `passwd1` field's configuration 
was being overwritten with only `autofocus`, dropping the `length` 
setting and causing the "New Password" field to have a maxlength of 30 
instead of 128, inconsistent with the "Confirm Password" field.

Fix: Preserve the `length` setting when setting `autofocus` in the 
reset-token flow of `PasswordChangeForm`.

### DIFF
--- a/include/class.staff.php
+++ b/include/class.staff.php
@@ -1608,7 +1608,7 @@ extends AbstractForm {
         // required for agents.
         if (isset($_SESSION['_staff']['reset-token'])) {
             unset($fields['current']);
-            $fields['passwd1']->set('configuration', array('autofocus' => true));
+            $fields['passwd1']->set('configuration', array('autofocus' => true, 'length' => '128'));
         }
         else {
             $fields['passwd1']->set('layout',


royont123:
Hi maintainers — just following up on this PR when you have a chance. It fixes the password reset maxlength issue from #6940. Happy to make any changes if needed. Thanks!

PR #6969, opened 2026-07-17
Cursor/faster pipeline models

jazneb23:


jazneb23:
Opened against upstream by mistake. This is a private strangler-fig demo change for jazneb23/osTicket only — closing.

